PR #18, opened 2018-04-12
Add Widget Placement API

MortimerGoro:


app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -39,6 +41,8 @@ static const int GestureSwipeRight = 1;
 
 static const float kScrollFactor = 20.0f; // Just picked what fell right.
 
+static crow::BrowserWorldPtr sWorld;
```
bluemarvin: I really don't like having a static shared point especially since we have it now in two places (native-lib.cpp) and this one never gets cleared from what I can tell.

app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -500,18 +520,60 @@ void
...
+      int32_t width = 0, height = 0;
+      widget->GetSurfaceTextureSize(width, height);
+      m.env->CallVoidMethod(m.activity, m.dispatchCreateWidgetMethod, widget->GetType(),
+                            widget->GetHandle(), aSurface, width, height);
     }
   }
 }
 
+void
+BrowserWorld::AddWidget(const WidgetPlacement & placement) {
```
bluemarvin: nit: extra space between WidgetPlacement and & -> WidgetPlacement&

app/src/main/cpp/WidgetPlacement.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace crow {
+
+WidgetPlacementPtr
+WidgetPlacement::FromJava(JNIEnv* aEnv, jobject& aObject) {
+  if (!aObject) {
+    return nullptr;
+  }
+
+  jclass clazz = aEnv->GetObjectClass(aObject);
```
bluemarvin: Maybe think how to cache this since I'm pretty sure this and GetFieldID are expensive? Probably can be done as follow up.

app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -500,18 +520,60 @@ void
...
   if (m.env && m.activity && m.dispatchCreateWidgetMethod) {
-    for (WidgetPtr& widget: m.widgets) {
-      if (aName == widget->GetSurfaceTextureName()) {
-        int32_t width = 0, height = 0;
-        widget->GetSurfaceTextureSize(width, height);
-        m.env->CallVoidMethod(m.activity, m.dispatchCreateWidgetMethod, widget->GetType(),
-                              widget->GetHandle(), aSurface, width, height);
-        return;
-      }
+    WidgetPtr widget = m.FindWidget([=](const WidgetPtr & w) {
```
bluemarvin: Extra space here. Also, I would prefer if we avoid single character var names as it makes it a pain to do search and replace on them (I realize m is a single character but it always has a . next to it :) ).

MortimerGoro:
PR updated with most of the nits fixed. Also simplified the Widgets storage to continue using a flat container instead of mirroring the scenegraph.

We can do the GetFieldID cache & the better way to access BrowserWorld as follow ups.

PR #21, opened 2018-04-18
Update ElbowModel to handle both left and right hand with a single in…

bluemarvin:
…stance.


app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
```diff
@@ -438,7 +439,13 @@ DeviceDelegateWaveVR::StartFrame() {
     }
     vrb::Matrix controllerTransform = vrb::Matrix::FromColumnMajor(pose.poseMatrix.m);
     if (m.elbow) {
-      controllerTransform = m.elbow->GetTransform(hmd, controllerTransform);
+      ElbowModel::HandEnum hand = ElbowModel::HandEnum::Right;
+      if ((role == WVR_DeviceType_Controller_Left) && (m.devicePairs[id].type == WVR_DeviceType_Controller_Right)) {
```
MortimerGoro: This could be simplified to:
```
if (role != m.devicePairs[id].type) {
  hand = ElbowModel::HandEnum::Left;
}
```
I'd also find usefull to add a comment here, it wasn't quick to understand the comparisons between the role & the devicePairs types.

PR #22, opened 2018-04-18
Fallback to head tracking input in SVR. Improve ODG scrolling support. 

MortimerGoro:


app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -222,6 +223,7 @@ BrowserWorld::State::UpdateControllers() {
   for (ControllerRecord& record: controllers) {
     vrb::Matrix transform = device->GetControllerTransform(record.index);
     record.controller->SetTransform(transform);
+    root->ToggleChild(*record.controller, !device->IsControllerUsingHeadTracking(record.index));
```
bluemarvin: I think we need a better API for showing and hiding the controller. The WaveVR version will also need a way to hide the controller(s) in some situations. I would rather toggle the node when the state changes rather than poll it every frame. Maybe we need a ControllerModel that can be registered with the delegate. Something like:

class ControllerModel {
public:
  virtual void SetTransform(const Matrix& aTransform) = 0;
  virtual void SetVisible(const bool aVisible) = 0;
  virtual void SetEnabled(const bool aEnabled) = 0;
};

Perhaps the ControllerRecord could implement this API?

MortimerGoro: I like the API, it will help to remove some methods from the delegates. I'd also add the scrolling to it.

ControllerRecord could implement the API but to add support for immersive mode we may need to be able to register multiple ControllerModels or switch the active controller model.

MortimerGoro:
Closing this because we have moved this implementation to https://github.com/MozillaReality/FirefoxReality/pull/26

PR #27, opened 2018-04-27
Support building preview builds when master is updated

bluemarvin:


bluemarvin:
Testing updated taskcluster config on PR before merging to master.

PR #28, opened 2018-04-30
Add support for more controller buttons

MortimerGoro:


bluemarvin:
I think this okay but I think we want to distinguish between the trigger and the touch pad click, even though daydream doesn't have a trigger.

MortimerGoro:
Sounds good, I'll update the PR

MortimerGoro:
PR updated

PR #32, opened 2018-05-11
Keyboard implementation

MortimerGoro:


app/src/common/shared/org/mozilla/vrbrowser/BrowserWidget.java
```diff
@@ -195,7 +202,8 @@ public boolean onKeyMultiple(int aKeyCode, int repeatCount, KeyEvent aEvent) {
     @Override
     protected void onFocusChanged(boolean aGainFocus, int aDirection, Rect aPreviouslyFocusedRect) {
         super.onFocusChanged(aGainFocus, aDirection, aPreviouslyFocusedRect);
-        Log.e(LOGTAG, "BrowserWidget onFoucusChanged: " + (aGainFocus ? "TRUE" : "FALSE"));
+        GeckoSession session = SessionStore.get().getSession(mSessionId);
```
bluemarvin: Doesn't look like this line is used?

app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
```diff
@@ -325,7 +411,7 @@ public void run() {
 
     // WidgetManagerDelegate
     @Override
-    public void addWidget(final WidgetPlacement aPlacement, final WidgetAddCallback aCallback) {
+    public void addWidget(final WidgetPlacement aPlacement, final boolean aVisible, final WidgetAddCallback aCallback) {
```
bluemarvin: Why isn't aVisible part of WidgetPlacement? Okay, I see it is for widgets that don't have a placement struct. We should just create all widgets from the Java side. This is one of the things I wanted to talk about next week. Having it separate for now is probably fine. We can fix with the refactor.

app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -766,32 +794,34 @@ BrowserWorld::AddWidget(const WidgetPlacement& placement, int32_t aCallbackId) {
...
-                                    worldWidth);
-  widget->SetAddCallbackId(aCallbackId);
+  float worldWidth, worldHeight;
   widget->GetWorldSize(worldWidth, worldHeight);
 
-  vrb::Vector translation = vrb::Vector(placement.translation.x() * kWorldDPIRatio,
-                                        placement.translation.y() * kWorldDPIRatio,
-                                        placement.translation.z() * kWorldDPIRatio);
+  vrb::Matrix transform = vrb::Matrix::Identity();
+  if (aPlacement.rotation != 0.0) {
```
bluemarvin: It is possible the rotation axis is not along Y so there would still be a non-identity transform created even if rotation is 0.0f. Probably better to check if axis magnitude is near zero?

app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -766,32 +794,34 @@ BrowserWorld::AddWidget(const WidgetPlacement& placement, int32_t aCallbackId) {
...
+  translation += vrb::Vector(parentWorldWith * aPlacement.parentAnchor.x() - parentWorldWith * 0.5f,
+                             parentWorldHeight * aPlacement.parentAnchor.y(),
                              0.0f);
 
-  widget->SetTransform(parent->GetTransform().PostMultiply(vrb::Matrix::Translation(translation)));
-  m.root->AddNode(widget->GetRoot());
-  m.widgets.push_back(widget);
+  transform.TranslateInPlace(translation);
+  widget->SetTransform(parent->GetTransform().PostMultiply(transform));
+  // Fixme: Remove this once we have proper scaling of the pointer
```
bluemarvin: You can fix this in a follow up by setting the pointer geometry to an empty group or something.

rossman22590:
so awesome! GJ :)

Issue #35, opened 2018-05-17
Emulated touch events can leave buttons in wrong state


bluemarvin:
Buttons can be left in hover state if the button is pressed and then released off of the button.

bluemarvin:
Can now reproduce with noapi and bluetooth mouse.

PR #43, opened 2018-05-22
#35 Ensure MotionEvent consistency to maintain correct button state

bluemarvin:


bluemarvin:
Failure due to GeckoView API changes. Working to resolve it now.

MortimerGoro:
LGTM

Compared with a mouse connected in the flat flavor, and it works the same way. Good work!

Issue #38, opened 2018-05-18
Improve URL detection on the URLBar


MortimerGoro:
CXB-22 Currently you need to add the protocol to detect a URL, and it does a google search if you don't add it.

Check what other browsers do in order to detect when to automatically load a URL instead of a search.

bluemarvin:
There are two issues here I think. One we need to detect when the protocol is missing from the url. Awesome bar looks for '.' separated with no white space. GeckoView doing a search is actually a bug in GV. All search needs to be external to GV. I expect soon, GV will error out if the url can't be resolved.

PR #44, opened 2018-05-23
Improve URL detection on the URLBar

MortimerGoro:
Fixes issue #38


bluemarvin:
This looks good. Maybe as a follow up, we can add a SessionStore API where we detect a failure to load and downgrade to http:// on failure, but probably only on URLs that were entered without a protocol?

We should also find out what should be done regarding search to makes sure Mozilla gets credit for the search?

PR #46, opened 2018-05-25
Basic transparency for widgets

MortimerGoro:


bluemarvin:
I really don't like this. This only works if there is a single transparent UI element.

MortimerGoro:
It worked with the keyboard + permission + tab overflow layouts together. We can add sorting by far distance as a follow-up if we find a problem.

It's just a basic transparency solution that should work for now until we implement the right solution in VRB.

PR #57, opened 2018-06-12
Add telemetry core ping.

pocmo:


MortimerGoro:
We have the telemetry added via https://github.com/MozillaReality/FirefoxReality/pull/85

Sorry, not sure why we didn't merge this at the time it was pushed and when the new PR arrived we forgot about this one.

At least I added some changes using the right way you show in the PR and attributed that commit:
https://github.com/MozillaReality/FirefoxReality/commit/a512cdf5b47b0fc9491896d8f843d3b8db2e5c8e

Issue #59, opened 2018-06-13
Better keyboard placement and visibility detection


MortimerGoro:
- Improve the keyboard placement.
- Use the latest GV api for a better keyboard visibility detection

MortimerGoro:
https://github.com/MozillaReality/FirefoxReality/pull/69

PR #69, opened 2018-06-19
Improve keyboard visibility detection #59

MortimerGoro:


MortimerGoro:
@keianhzo I moved the code you added to initialize, I think that users want to remember the last input mode

```
        mKeyboardview.setKeyboard(mKeyboardQuerty);
        if (mOriginalLayoutParams != null)
            mKeyboardview.setLayoutParams(mOriginalLayoutParams);
        mKeyboardview.setVisibility(View.VISIBLE);
        mVoiceInput.setVisibility(View.GONE);
        mIsPopupVisible = false;
```

Issue #65, opened 2018-06-18
Implement focus mode


MortimerGoro:

MortimerGoro:
https://github.com/MozillaReality/FirefoxReality/pull/76

Issue #55, opened 2018-06-08
Ensure back button behaves correctly when in fullscreen mode.


bluemarvin:
Need to make sure and catch when gecko view is in fullscreen mode so that clicking the back button on the controller takes gecko view out of fullscreen mode instead of actually going back in history.

keianhzo:
This is working fine after te focus mode has been implemented, the back handler is always added when GeckoSession.ContentDelegate,onFullScreen is called.

PR #76, opened 2018-06-26
Implement Focus mode, full screen mode and widget resizing

### LINKED ISSUE #51: Full-screen mode
- Implement full-screen mode
- Enabled from chrome UI or from a JavaScript call
- Exit from full-screen mode

MortimerGoro:
- Implement Focus mode, full screen mode and widget resizing 
- Implement reusable Quad node which adds supports different scale modes (Quad is used by Widget nodes now)
- Implement world Fade Out/In API and blitter
- Implement WidgetResizer utility node

Fixes #51,  #55 and #65


bluemarvin:
On a large code change like this, if two people are requested for the review, I would prefer if both review it before it is landed. There are some things I have comments about. I'm going to go ahead and make the comments on the code and we can follow up in a later PR is needed.

bluemarvin:
I'm a little concerned about the fade blitter adding an entire frame buffer worth of fill. It seems to perform okay right now, but there isn't a lot of geometry in the scene currently. Is it possible to handle this with a lighting effect?

I found that if you are watching a video in full screen mode and go back. The UI doesn't go back to normal with URL bar and buttons. I didn't see any code added to check if in full screen mode onBack so that might be the issue.

I think there is also a bug in GeckoView where when going out of fullscreen mode, you can't scroll any more and need to reload the page to get scrolling to work. I see that that gets filed, I think it is affect Focus/Klar browser also.

MortimerGoro:
> On a large code change like this, if two people are requested for the review, I would prefer if both review it before it is landed. There are some things I have comments about. I'm going to go ahead and make the comments on the code and we can follow up in a later PR if needed.

Yep, he asked if it was possible to merge earlier because he was blocked by some APIs added on this PR but forgot to mention here that we'll handle your comments/reviews as follow-ups.

> I'm a little concerned about the fade blitter adding an entire frame buffer worth of fill. It seems to perform okay right now, but there isn't a lot of geometry in the scene currently. Is it possible to handle this with a lighting effect?

I thought today about adding something like a tintColor to the RenderState which allows to multiply the resulting v_color in the vertex shader with the fixed value. This would be easier to handle because we don't have to deal with the hierarchy of lights, save/restore lights, etc. We are also going to require it for fade transitions to private browsing session in the browser widget itself. What do you think?

> I found that if you are watching a video in full screen mode and go back. The UI doesn't go back to normal with URL bar and buttons. I didn't see any code added to check if in full screen mode onBack so that might be the issue.

I added this code to handle a similar scenario but forgot to add it to back handler too. Will test and add it:
```
if (mFocusDueToFullScreen) {
     mFocusDueToFullScreen = false;
      exitFocusMode();
}
```

> I think there is also a bug in GeckoView where when going out of fullscreen mode, you can't scroll any more and need to reload the page to get scrolling to work. I see that that gets filed, I think it is affect Focus/Klar browser also.

I didn't notice that with the youtube tests. Is that bug already filled to bugzilla?

bluemarvin:
> I thought today about adding something like a tintColor to the RenderState which allows to multiply the resulting v_color in the vertex shader with the fixed value. This would be easier to handle because we don't have to deal with the hierarchy of lights, save/restore lights, etc. We are also going to require it for fade transitions to private browsing session in the browser widget itself. What do you think?

That sounds good too. I don't think using lighting would be too bad, but this sounds easier. 

>> I think there is also a bug in GeckoView where when going out of fullscreen mode, you can't scroll >> any more and need to reload the page to get scrolling to work. I see that that gets filed, I think it is >> affect Focus/Klar browser also.
>
> I didn't notice that with the youtube tests. Is that bug already filled to bugzilla?

https://bugzilla.mozilla.org/show_bug.cgi?id=1471340

bluemarvin:
> Yep, he asked if it was possible to merge earlier because he was blocked by some APIs added on this PR but forgot to mention here that we'll handle your comments/reviews as follow-ups.

That's fine. One thing I will do if I need some code that hasn't landed is pull the branch and cherry pick. Once it lands git usually figures that out, if the patch has changed too much it easy to drop the cherry-picked branch on rebase.

keianhzo:
Totally my fault @bluemarvin I needed some commits to finish the private browsing task but I forgot to let you know. Will do that next time instead of merging, thanks!

> One thing I will do if I need some code that hasn't landed is pull the branch and cherry pick. Once it lands git usually figures that out, if the patch has changed too much it easy to drop the cherry-picked branch on rebase.

PR #82, opened 2018-06-28
[CXB-190] Mapping client side errors pages- text only

keianhzo:
Added support for cutom error pages.

@MortimerGoro As Gecko doesn't have proper support for error codes reporting this is a bit hacky at the moment, mainly parsing the loaded URL to guess when Gecko is reporting an error based on the URL parameters:

- about:neterror?e=dnsNotFound&u=https%3A//ghftyyghjyreasggh.com/&c=UTF-8&f=regular&d=ghftyyghjyreasggh.com%20could%20not%20be%20found.%20Please%20check%20the%20name%20and%20try%20again.
- about:certerror?e=nssBadCert&u=https%3A//self-signed.badssl.com/&c=UTF-8&f=regular&d=%20

For the moment I'm just using a template and injecting the params.


app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -445,8 +452,65 @@ public void onCanGoForward(GeckoSession aSession, boolean aCanGoForward) {
...
+                        session.loadUri("javascript:updateMessage('" + errorType + "', '" + errorURL + "', '" + errorDescription + "');");
+                    }
+                };
+                handler.postDelayed(r, 0);
+
+            } catch (UnsupportedEncodingException e) {
+                e.printStackTrace();
+            }
+
+        } else if (!aUri.startsWith("javascrtipt:")) {
```
MortimerGoro: "javascrtipt:" instead of "javascript:"

keianhzo:
Fixed

PR #85, opened 2018-07-02
[CXB-184] Add telemetry component support.

daoshengmu:
Refer #57 and [FirefoxTV](https://github.com/mozilla-mobile/firefox-tv/blob/master/app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt#L45). Adding [Android telemetry component support](https://github.com/mozilla-mobile/android-components), we measure the session spent time when using FxR.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -174,6 +174,7 @@ dependencies {
     svrImplementation fileTree(dir: "${project.rootDir}/third_party/svr/", include: ['*.jar'])
     implementation 'com.android.support:design:26.1.0'
     implementation 'com.google.vr:sdk-audio:1.140.0'
+    implementation "org.mozilla.components:telemetry:0.10"
 }
 
 if (findProject(':wavesdk')) {
--- a/app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
@@ -1,0 +1,101 @@
+package org.mozilla.vrbrowser.telemetry;
+
+import android.content.Context;
+import android.content.res.Resources;
+import android.os.StrictMode;
+import android.support.annotation.UiThread;
+import org.mozilla.telemetry.Telemetry;
+import org.mozilla.telemetry.TelemetryHolder;
+import org.mozilla.telemetry.config.TelemetryConfiguration;
+import org.mozilla.telemetry.event.TelemetryEvent;
+import org.mozilla.telemetry.net.HttpURLConnectionTelemetryClient;
+import org.mozilla.telemetry.ping.TelemetryCorePingBuilder;
+import org.mozilla.telemetry.ping.TelemetryMobileEventPingBuilder;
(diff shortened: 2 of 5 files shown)


daoshengmu:
@bluemarvin or @MortimerGoro Could you take a review? I was trying to assign a reviewer from the right sidebar but it doesn't show any option for me.

MortimerGoro:
@daoshengmu could you implement the TelemetryWrapper in Java instead of Kotlin?

If were are using Kotlin only for that I think that it's not worth to install the required additional greadle plugins which could add potential build errors

daoshengmu:
@MortimerGoro Sounds great! I have adjusted `TelemetryWrapper` from Kotlin to Java format.

app/build.gradle
```diff
@@ -174,6 +174,7 @@ dependencies {
     svrImplementation fileTree(dir: "${project.rootDir}/third_party/svr/", include: ['*.jar'])
     implementation 'com.android.support:design:26.1.0'
     implementation 'com.google.vr:sdk-audio:1.140.0'
+    implementation "org.mozilla.components:telemetry:$mozilla_components_version"
```
MortimerGoro: Set the version here instead of using $mozilla_components_version to be consistent with the other dependencies

build.gradle
```diff
@@ -1,6 +1,8 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
+    ext.mozilla_components_version = '0.10'
```
MortimerGoro: This can be removed when the version is set inline in the dependency declararion

app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
```diff
@@ -0,0 +1,116 @@
...
+    private class Method {
+        private static final String FOREGROUND = "foreground";
+        private static final String BACKGROUND = "background";
+    }
+
+    private class Object {
+        private static final String APP = "app";
+    }
+
+    public static boolean isTelemetryEnabled(Context context) {
```
MortimerGoro: We have a utility method to check that: SettingsStore.getInstance(ctx).isTelemetryEnabled()

MortimerGoro: Also we should call setTelemetryEnabled() when the setting is changed in FxR

daoshengmu: I remove set/getTelemetryEnabled() APIs from TelemetryWrapper. Users only can use set/getTelemetryEnabled() by `SettingsStore.getInstance(ctx)`

daoshengmu:
Some updates:
- Removing set/getTelemetryEnabled() APIs from TelemetryWrapper. Users only can use set/getTelemetryEnabled() by `SettingsStore.getInstance(ctx)`.

- Make the component version define in `app/build.gradle` instead of in other files.

PR #92, opened 2018-07-05
Better handling of the app state and opened gecko sessions.

### LINKED ISSUE #91: Sessions are not correctly restored in some situations
After leaving the app, in case you do not "continue" or "quit" and just do something else, the sessions (opened private sessions) are not correctly restored when app is reopened.

keianhzo:
Fixes #91 

There were some situations were private sessions were not being correctly restored. For exmaple after leaving the app, in case you do not "continue" or "quit" and just go somewhere else an open the app later the app sessions were still alive but we were not restoring the private/public sessions correctly.

Please test in other headsets as I've only tested on Oculus Go and thetre might be some Android lifecycle related differences on those.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
@@ -27,6 +27,7 @@ import org.mozilla.vrbrowser.ui.BrowserWidget;
 import org.mozilla.vrbrowser.ui.KeyboardWidget;
 import org.mozilla.vrbrowser.ui.NavigationBarWidget;
 import org.mozilla.vrbrowser.ui.OffscreenDisplay;
+import org.mozilla.vrbrowser.ui.SettingsStore;
 import org.mozilla.vrbrowser.ui.SettingsWidget;
 import org.mozilla.vrbrowser.ui.TopBarWidget;
 
@@ -128,6 +129,21 @@ public class VRBrowserActivity extends PlatformActivity implements WidgetManager
             }
         });
         initializeWorld();
+
+        int sessionId = SettingsStore.getInstance(this).getCurrentSessionId();
+        if (sessionId != SessionStore.NO_SESSION_ID) {
+            GeckoSession session = SessionStore.get().getSession(sessionId);
+            if (session != null) {
+                SessionStore.get().setCurrentSession(sessionId);
+                mPreviousSessionId = SettingsStore.getInstance(this).getPreviousSessionId();
+
+                if (session.getSettings().getBoolean(GeckoSessionSettings.USE_PRIVATE_MODE)) {
+                    mNavigationBar.setPrivateBrowsingEnabled(true);
+                    mTopBar.setPrivateBrowsingEnabled(true);
+                    mBrowserWidget.setPrivateBrowsingEnabled(true);
+                }
+            }
+        }
     }
 
     protected void initializeWorld() {
@@ -191,6 +207,10 @@ public class VRBrowserActivity extends PlatformActivity implements WidgetManager
         if (mPermissionDelegate != null) {
             mPermissionDelegate.release();
         }
+
+        SettingsStore.getInstance(this).setCurrentSeesionId(SessionStore.get().getCurrentSessionId());
+        SettingsStore.getInstance(this).setPreviousSeesionId(mPreviousSessionId);
+
         super.onDestroy();
     }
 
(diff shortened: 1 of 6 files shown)


app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
```diff
@@ -128,6 +129,21 @@ public void run() {
...
         initializeWorld();
+
+        int sessionId = SettingsStore.getInstance(this).getCurrentSessionId();
+        if (sessionId != SessionStore.NO_SESSION_ID) {
+            GeckoSession session = SessionStore.get().getSession(sessionId);
+            if (session != null) {
+                SessionStore.get().setCurrentSession(sessionId);
+                mPreviousSessionId = SettingsStore.getInstance(this).getPreviousSessionId();
+
+                if (session.getSettings().getBoolean(GeckoSessionSettings.USE_PRIVATE_MODE)) {
```
MortimerGoro: The private browsing mode state setters are starting to get duplicated in different parts of the code. Could you use the SessionChangeListener.onCurrentSessionChange listener to handle the state and get rid of the setters in VRBrowserActivity?

app/src/common/shared/org/mozilla/vrbrowser/ui/SettingsStore.java
```diff
@@ -73,4 +73,25 @@ public void setTelemetryEnabled(boolean isEnabled) {
         TelemetryHolder.get().getConfiguration().setCollectionEnabled(isEnabled);
     }
 
+
+    public void setCurrentSeesionId(int sessionId) {
```
MortimerGoro: typo: seesion

app/src/common/shared/org/mozilla/vrbrowser/ui/SettingsStore.java
```diff
@@ -73,4 +73,25 @@ public void setTelemetryEnabled(boolean isEnabled) {
...
+        return mPrefs.getInt(mContext.getString(R.string.settings_key_session_id), SessionStore.NO_SESSION_ID);
+    }
+
+    public void setPreviousSeesionId(int sessionId) {
+        SharedPreferences.Editor editor = mPrefs.edit();
+        editor.putInt(mContext.getString(R.string.settings_key_previous_session_id), sessionId);
+        editor.commit();
+    }
+
+    public int getPreviousSessionId() {
```
MortimerGoro: Would it make sense to name the methods: getCurrentSessionId() and getCurrentPrivateSessionId()
I think it's clearer to call them that way if I have understand the usage correctly.

app/src/common/shared/org/mozilla/vrbrowser/ui/SettingsStore.java
```diff
@@ -73,4 +73,25 @@ public void setTelemetryEnabled(boolean isEnabled) {
...
+        SharedPreferences.Editor editor = mPrefs.edit();
+        editor.putInt(mContext.getString(R.string.settings_key_session_id), sessionId);
+        editor.commit();
+    }
+
+    public int getCurrentSessionId() {
+        return mPrefs.getInt(mContext.getString(R.string.settings_key_session_id), SessionStore.NO_SESSION_ID);
+    }
+
+    public void setPreviousSeesionId(int sessionId) {
```
MortimerGoro: The setter can also be automatically done using the SessionChangeListener.onCurrentSessionChange listener insteaf of using external calls

MortimerGoro:
I think it would be a good idea to use the listeners to handle the state changes instead of the explicit setters. We are going to add more ways to change sessions in the future (carrousel, voice input) so it would be better to handle the state in a single place

app/src/common/shared/org/mozilla/vrbrowser/ui/SettingsStore.java
MortimerGoro: mmm you'd also need a extra variable to know if the current is public/private. So it's ok

keianhzo:
Yes, using SessionChangeListener is the right way, changed all but the current/previous session id setters as when the activityis detroyed we don't get any session callback but we still need to save the session for restoring when the activity is recreated (as sessions are still alive in Gecko).

app/src/common/shared/org/mozilla/vrbrowser/ui/TopBarWidget.java
```diff
@@ -56,6 +59,8 @@ public void onClick(View view) {
         });
 
         mAudio = AudioEngine.fromContext(aContext);
+
+        SessionStore.get().addSessionChangeListener(this);
```
MortimerGoro: unregister the listener on
```
 @Override
public void releaseWidget() {
}
```

keianhzo: Fixed

PR #98, opened 2018-07-06
Added new controllers for Focus, Go and Daydream

keianhzo:


app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -301,7 +301,7 @@ DeviceDelegateOculusVR::GetControllerModelCount() const {
 const std::string
 DeviceDelegateOculusVR::GetControllerModelName(const int32_t aModelIndex) const {
   // FIXME: Need Oculus based controller
-  static const std::string name("vr_controller_daydream.obj");
+  static const std::string name("vr_controller_oculusgo.OBJ");
```
bluemarvin: Please keep model file names all lower case.

keianhzo: Fixed

keianhzo:
Fixed scaling issues based on #111 and file names

PR #99, opened 2018-07-06
Added mini tray v6 by request of design

keianhzo:


MortimerGoro:
Also check the correct scale with https://github.com/MozillaReality/FirefoxReality/pull/111

keianhzo:
The size is not terrible so I think we can leave it like that until design reviews.

PR #101, opened 2018-07-08
Support off render thread gl texture and buffer creation

bluemarvin:


MortimerGoro:
Great work, no hang when loading the tray.

The only problem I found is that it crashes in Daydream, when launched from the icon (it works ok launched from the VR launcher)

PR #112, opened 2018-07-09
Delay loader initialization to fix a crash in Daydream

MortimerGoro:
…plicated activity calls)


bluemarvin:
@MortimerGoro looks good. Feel free to merge when taskcluster finishes.

Issue #108, opened 2018-07-08
Resize behavior in focus mode is inconsistent.


bluemarvin:
STR:
1) Go into focus mode.
2) Go into resize mode.
3) Make window taller with drag handle.
4) Leave focus mode.

Expected result: Window returns to default size.
Actual result: window remains the same size while content is default size so it looks stretched in the window.

PR #114, opened 2018-07-09
Focus/Full screen should remember window size set by user

### LINKED ISSUE #105: Focus/Full screen should remember window size set by user
STR:
1) Go into focus mode
2) Set window size to 3x
3) Leave exit focus mode
4) Go back into focus mode

Expected result: Screen should still be 3x
Actual result: screen is default size.

keianhzo:
Fixes #105

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/BrowserWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/BrowserWidget.java
@@ -6,6 +6,7 @@
 package org.mozilla.vrbrowser.ui;
 
 import android.content.Context;
+import android.graphics.PointF;
 import android.graphics.Rect;
 import android.graphics.SurfaceTexture;
 import android.view.KeyEvent;
@@ -38,6 +39,7 @@ public class BrowserWidget extends View implements Widget, SessionStore.SessionC
     private int mHandle;
     private WidgetPlacement mWidgetPlacement;
     private WidgetManagerDelegate mWidgetManager;
+    private PointF mLastWorldSize;
 
     public BrowserWidget(Context aContext, int aSessionId) {
         super(aContext);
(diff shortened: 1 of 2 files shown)


MortimerGoro:
We need to handle custom sizes too, check this method

```
    // WidgetManagerDelegate.Listener
    @Override
    public void onWidgetUpdate(Widget aWidget) {
        if (aWidget != mBrowserWidget) {
            return;
        }

        // Browser window may have been resized, adjust the navigation bar
        float targetWidth = aWidget.getPlacement().worldWidth;
        float defaultWidth = WidgetPlacement.floatDimension(getContext(), R.dimen.browser_world_width);
        targetWidth = Math.max(defaultWidth, targetWidth);
        // targetWidth = Math.min((targetWidth, defaultWidth * 2.0f);

        float ratio = targetWidth / defaultWidth;
        mWidgetPlacement.worldWidth = targetWidth;
        mWidgetPlacement.width = (int) (WidgetPlacement.dpDimension(getContext(), R.dimen.navigation_bar_width) * ratio);
        mWidgetManager.updateWidget(this);
    }
```

keianhzo:
Reverted Fullscreen to always 2x and added support for custom focus mode presets.

bluemarvin:
I'm still seeing issues with resizing windows.

STR:
1) Go into manual resize mode.
2) Stretch window so it is taller.
3) Exit focus mode.
expected result: Window returns to default size.
actual result: Window stays tall while content resizes and gets stretched.

Also, when in resize mode after a manual resize, clicking the 1x button doesn't change the height of the window and it stay tall.

keianhzo:
Doesn't look like it's related to this fix. It's also happening on master and it seems to be related to the way we are handling the browser widget resizing in handleResizeEvent. It only happens in the case you described.

I'll file a bug and we can address this in a follow up task.

bluemarvin:
Sorry, got my issues mixed up. Was thinking of #108

PR #119, opened 2018-07-10
Close button fails to update position after leaving focus mode

### LINKED ISSUE #102: Close button fails to update position after leaving focus mode
STR:
1) Go into focus mode.
2) Make scree 3x
3) Leave focus mode

Expected results: Close button updates position to be above browser window.
Actual results: Close button retains position from large focus mode window.

keianhzo:
Fixes #102


MortimerGoro:
Is the close button using a Widget as width as the browser?

We are rendering to textures so in this scenario that's a too big texture just for a button. Can you make the Widget just use the size of the button?

keianhzo:
Fixed uiwidget size. There is a preoblem with the worldWidth that will be fixed in a follow up commit

PR #133, opened 2018-07-11
(#132) Fix gecko about: pages

bluemarvin:


bluemarvin:
Pages listed by about:about and if they work:
about:addons - no
about:buildconfig - yes
about:cache - yes
about:checkerboard -yes
about:config - there is a page but it doesn't allow you to see or edit config
about:credits - yes
about:license - yes
about:logo - no
about:memory - yes
about:mozilla - yes
about:networking - yes
about:performance - yes
about:plugins - yes
about:serviceworkers - yes
about:support - yes but reports application is fennec
about:telemetry - yes
about:url-classifier - yes
about:webrtc - yes

PR #135, opened 2018-07-11
[CXB-183] Enable crash report

daoshengmu:
Enable Java and GV crash report. I also refer MOZ_CRASH() from Gecko to give abort() when we need to crash it at the native code.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/SettingsStore.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/SettingsStore.java
@@ -28,6 +28,7 @@ public class SettingsStore {
 
     private Context mContext;
     private SharedPreferences mPrefs;
+    private final static boolean enableCrashReportByDefault = true;
     // Enable telemetry by default (opt-out).
     private final static boolean enableTelemetryByDefault = true;
 
@@ -37,7 +38,7 @@ public class SettingsStore {
     }
 
     public boolean isCrashReportingEnabled() {
-        return mPrefs.getBoolean(mContext.getString(R.string.settings_key_crash), false);
+        return mPrefs.getBoolean(mContext.getString(R.string.settings_key_crash), enableCrashReportByDefault);
     }
 
     public void setCrashReportingEnabled(boolean isEnabled) {


app/src/main/cpp/Assertions.h
```diff
@@ -0,0 +1,51 @@
+/* -*- Mode: C++; tab-width: 20; indent-tabs-mode: nil; c-basic-offset: 2 -*-
```
bluemarvin: Why is this here? Just call abort(), there is no need for an assertion macro.

daoshengmu: Yep. calling abort() is enough for the crash reports, I was wondering if we wanna show some logs for developers before crashes.

If you still think we don't need this, I am happy to remove it.

bluemarvin: As of right now we don't have a condition where we crash on purpose. I would rather add something when there is a need.

bluemarvin:
Has there been a decision from legal as to whether crash reporting (and telemetry) are on or off by default?

daoshengmu:
We are considering to follow the decisions what Desktop and Fennec did. Telemetry is opt-in from non-release channel, the crash reporting is on but we should let users to choose to submit the report.

MortimerGoro:
`the crash reporting is on but we should let users to choose to submit the report.`

How does that dialog work. Should we implement it in VR? IIRC we only have created the UI to enable/disable crash reporting, but not to let users choose to submit the report

bluemarvin:
Who is we? There is going to be very few users initially. Having crash reporting and telemetry on by default would be preferable but needs approval. Also, how is a user going to choose to submit a report? The app crashed, there won't be any way to ask them.

daoshengmu:
If we only have one process, when the process is died, we have no chance to ask if users want to submit their reports.

I am wondering if we should do some changes to make GV run at the content process, the user sensitive data is at that process, we can ask users to submit the crashes from there. The crashes from the main process (like our Java, Native code) should be submitted directly.

bluemarvin:
This isn't a desktop environment, there is only one render thread. If the process with the render thread crashes, there will be not prompt. If e10s is enabled and the child process crashes, the report is submitted if crash reporting is enabled. There is no prompting mechanism.

MortimerGoro:
Another option: show the crash submit dialog  at the start of the next launch after the crash.

daoshengmu:
@bluemarvin We sometimes still need the prompt for the content. For example. when users are viewing sensitive pages like bank websites, they would not want to submit their crash reports.

bluemarvin:
GeckoView doesn't send URLs in crash data.

daoshengmu:
Sounds great. I would like to choose more easier way. Think about when we need to ask prompting or just follow the preference setting. If we do need to ask prompting, @MortimerGoro's `submit dialog at the start of the next launch after the crash` is a good way to go.

bluemarvin:
The easy way is to not prompt. GeckoView doesn't have a prompt API for crash data. The GeckoView team would rather not add one.

daoshengmu:
@bluemarvin I have updated my patch for removing the `Assertions.h`, please help review again.

Thanks.

larsbergstrom:
If we are doing full crash dumps (which for GV do not include URLs or IPs but do include a memory dump) then it must be opt-in.

If we are doing the minimal crash reporting via Sentry of the other Firefox apps, which does not contain info at all, it may be opt-out.

So, it appears that we will need to be opt-in for crash reporting and may need another switch and possibly prompts, depending on how hard we want to push for the user to send this information.

bluemarvin:
I'm closing this since we have the control panel now and won't be able to enable by default until we can prompt the user.

PR #137, opened 2018-07-12
Refactor controller handling code to it's own classes

MortimerGoro:


app/src/main/cpp/ControllerManager.cpp
```diff
@@ -0,0 +1,479 @@
...
+    if (controller.transform) {
+      controller.transform->RemoveFromParents();
+    }
+    controller.Reset();
+  }
+  m.gestures = nullptr;
+}
+
+void
+ControllerManager::UpdateControllers(const float aFarClip, const std::vector<WidgetPtr>& aWidgets, const TrayPtr& aTray, bool& aRelayoutWidgets) {
```
bluemarvin: I'm wondering if this function should remain part of BrowserWorld? It seem like a lot of state to pass in only to use the controller's transformMatrix and buttons?. It seems like controller manager shouldn't need to know about widgets, or trays, or relayout. BrowserWorld feels like the better place for all these relatively independent pieces to meet.

MortimerGoro: I'm ok with leaving it in BrowserWorld to reduce the dependencies. I'll update the PR.

PR #138, opened 2018-07-12
Avoid deadlock when pausing the compositor in immersive mode

MortimerGoro:
Firefox Reality pauses the compositor when gfxVRExternal presentation begins.

There is a edge race condition which causes a deadlock. It happens when pausing the compositor while gfxVRExternal is waiting for a SubmitFrame result.


MortimerGoro:
GV side: https://bugzilla.mozilla.org/show_bug.cgi?id=1475270

bluemarvin:
Sorry, I didn't see @kearwood was also tagged for review when I pushed it. He can still review and we can follow up if need be.

Issue #141, opened 2018-07-12
Hide exit button on top of the browser widget in home


MortimerGoro:
Exit button doesn't do anything in home so must be hidden and only shown when actions available

PR #143, opened 2018-07-12
Hide close button when not is Private Mode.

keianhzo:
#141


bluemarvin:
I've played around a little more with this patch. And I can get into a state where the close button does not appear when in private browsing mode. I don't see anything obvious in this patch. I'll try and come up with and STR.

bluemarvin:
I have not been able to reproduce this bug yet. I guess we can land this and open a new issue once the STR has been found.

PR #144, opened 2018-07-12
Moved settings panel closer to the user

keianhzo:
Let me know how do you feel about that, just move a bit closer to the user but maybe bigger font sie in necessary.


bluemarvin:
It's more legible but I think making the fonts larger would also help.

keianhzo:
Made font bigger. Let's merge this, see how people react and make any required updatesin a follow up.

PR #153, opened 2018-07-13
Fix immersive mode quaternion and view matrices

MortimerGoro:
WebVR expects inverse rotation


MortimerGoro:
Need to merge this to avoid some conflict in the immersive controllers PR. Create some issue follow up if you want to mention something.

Issue #148, opened 2018-07-12
Using resize multiplier button can fail to properly resize window in focus mode


bluemarvin:
STR:
1) Go into focus mode
2) Resize window to 2x
3) Manually make window wider
4) Press resize 2x button again
expect results: Window is restored to 2x size
actual results: window stays same size but content gets resized.

keianhzo:
I've been trying for a while and can't reproduce it, Could this be a duplicate of #108 ?

PR #159, opened 2018-07-13
Fix browser resizing issues

keianhzo:
#108 
#148 

Fix resizing cases where worldWidth doesn't change but widget width does.

### DIFF
--- a/app/src/main/cpp/BrowserWorld.cpp
+++ b/app/src/main/cpp/BrowserWorld.cpp
@@ -583,6 +583,13 @@ BrowserWorld::UpdateWidget(int32_t aHandle, const WidgetPlacementPtr& aPlacement
       return;
   }
 
+  int32_t oldWidth = 0;
+  int32_t oldHeight = 0;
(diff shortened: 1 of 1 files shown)


app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -606,7 +611,7 @@ BrowserWorld::UpdateWidget(int32_t aHandle, const WidgetPlacementPtr& aPlacement
     newWorldWidth = aPlacement->width * kWorldDPIRatio;
   }
 
-  if (newWorldWidth != worldWidth) {
+  if (newWorldWidth != worldWidth || oldWidth != aPlacement->width) {
```
MortimerGoro: Don't we need to check height? What happens if both worldWidth & aPlacement->width are equal, but height changes?

keianhzo: That's what I thought but for some reason when scaling in portrait mode what changes is the width not the height. Anyway I've added both checks just in case although the one that makes it work is the width one.

Issue #142, opened 2018-07-12
Stop GeckoSession when session is not active


keianhzo:
STR:
- Open YouTube, play a video
- Open a Private session

Result: The regular session video is still playing
Expected result: Video should stop playing when the session is left.

PR #160, opened 2018-07-13
Make previous session inactive when switching sessions

keianhzo:
#142 

I'll add a task to refactor all this private mode code as it doesn't really belong here and it's growing.


keianhzo:
I opened issue 161 for refactoring the private session code so I'll address this in that task if that's ok.

Issue #93, opened 2018-07-05
GV preferences not applied on first launch


MortimerGoro:
e.g: Enabling WebVR preference and loading the WebVR sample gets the "page says WebVR found but no displays available." on the first launch.
On next launches it runs ok.

bluemarvin:
This is due to the fact that PrefsHelper.setPref set the prefs after gfxPref are read. After the first run, the prefs are written to the user profile so that they are read correctly the next time FxR is run. The best solution until an new Prefs API is added to gecko is to write out a user.js with these prefs to the profile directory so they may be read before gfxPrefs are read.

PR #174, opened 2018-07-16
(#93) Ensure Gecko VR prefs take effect on first run

bluemarvin:


app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -419,6 +423,24 @@ public void setPermissionDelegate(GeckoSession.PermissionDelegate aDelegate) {
         }
     }
 
+    private void vrPrefsWorkAround(Context aContext) {
+        File path = GeckoProfile.initFromArgs(aContext, null).getDir();
+        String prefFileName = path.getAbsolutePath() + File.separator + "user.js";
+        Log.i(LOGTAG, "Creating file: " + prefFileName);
+        try {
+            FileOutputStream out = new FileOutputStream(prefFileName);
```
MortimerGoro: File is not closed if there is an exception. You can use the new Java syntax which calls close automatically, even if there are exceptions or returns.
```
try (FileOutputStream out = new FileOutputStream(prefFileName)) {
    out.method();
} 
```

bluemarvin:
PR updated and rebased.

PR #176, opened 2018-07-17
(#175) Ensure listeners are cleared from SessionStore on destroyActivity

bluemarvin:


keianhzo:
Now the activity seems to be correctly destroyed. I'm good to merge after the conflicts are solved.

bluemarvin:
Rebased

PR #177, opened 2018-07-17
Meadow env v2 with day/night cubemaps instant transition

keianhzo:
[CXB-181]

Actually the day cubemap with the fade effect is almost the same as the night cubemap so we could just just remove the night one and let the fade be the transition.


app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -782,80 +789,81 @@ vrb::TransformPtr
...
-    std::array<GLushort, 24> cubeIndices{
-        0, 1, 2, 3,
-        3, 2, 6, 7,
-        7, 6, 5, 4,
-        4, 5, 1, 0,
-        0, 3, 7, 4,
-        1, 5, 6, 2
-    };
+void
+BrowserWorld::LoadSkybox(const vrb::TransformPtr transform, const std::string& basePath) {
```
bluemarvin: Looks like indentation got messed up here? The lambda function looks like it is using four space instead of two?

MortimerGoro:
- There is a very noticeable change after the day/night fades. I'd remove the night cubemap and just use the fade, nobody is going to notice it.

- The illumination of the browser window doesn't feel right with this environment, it's too dark, white looks like grey.

- It takes a long time to load

`Having looked at this now in app. I really don't think we should ship this`

I agree

`Why does it even have a skybox? We could get the same effect just setting the clear color to sky blue`

The skybox has some gradient effect. We could replicate it procedurally and get rid of 6 1024x1024 textures. But not sure if it's worth the effort to do that now, even more if we are not sure of landing this...

keianhzo:
Agree with all the above. Fixed indent and removed night skybox. Let's land this so everybody can see it and make decisions based on how it actually looks like.

keianhzo:
Created branches for different approaches:

- **cubemap + 4K**: meadow_env
6x1K textures for sky + far objects cubemap
1x4K texture for object

- **palette**: meadow_env_palette
6x1K textures for sky cubemap
64x64 texture for object

- **2x4K**: meadow_env_2maps
6x1K textures for sky cubemap
2x4K texture for objects

bluemarvin:
I don't see the same tearing on the Go.

bluemarvin:
Far clipping plane on VIVE Focus is too close.

keianhzo:
This is not the final version just something that design want s to test on today's testing session. We still need to do more tweaks on this env.

PR #188, opened 2018-07-18
Filter resource and jar URLs

### LINKED ISSUE #130: Hide some intermittent URL changes
Sometimes I saw something like  "jar://" in the URL while loading a web. We need to hide those intermittent URL changes. Not sure if this only happens with local URLs in assets

keianhzo:
Fixes #130 

We are nor hiding all jar: and resources: URLs. Now when at home the URL bar is empty, same as Firefox does

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/NavigationURLBar.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/NavigationURLBar.java
@@ -92,7 +92,12 @@ public class NavigationURLBar extends FrameLayout {
     public void setURL(String aURL) {
         int index = -1;
         if (aURL != null) {
-            index = aURL.indexOf("://");
+            if (aURL.startsWith("jar:"))
+                return;
+            else if (aURL.startsWith("resource:"))
+                aURL = "";
+            else
+                index = aURL.indexOf("://");
         }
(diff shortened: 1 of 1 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/NavigationURLBar.java
```diff
@@ -92,7 +92,10 @@ public boolean onEditorAction(TextView aTextView, int actionId, KeyEvent event)
     public void setURL(String aURL) {
         int index = -1;
         if (aURL != null) {
-            index = aURL.indexOf("://");
+            if (aURL.startsWith("jar:") || aURL.startsWith("resource:"))
```
MortimerGoro: Could this cause intermittent blank URLS while loading a web. I mean you have a correct URL, and while a jar:// URL is received it will show blank but a shirt time before showing the final URL. 

I think a single URL bar transition would be better from a design perspective (from previous  URL to next final URL, without the intermediate blank)

keianhzo: I couldn't reproduce the intermediate jar: URL issue but I've separated it in two cases as we always want to set the URL blank for resource: URL.

PR #192, opened 2018-07-18
Fade effect gets out of sync

### LINKED ISSUE #146: Fade effect gets out of sync
@keianhzo 
STR:
1) Go into private browsing mode
2) Go into focus mode
3) Go out of focus mode
expected results: Fade effects is still present.
actual results: Fade effect is off even though still in private browsing mode.

keianhzo:
Fixes #146 
Fixed #146 

This will get refactored when I address #161


MortimerGoro:
Conflict needs to be resolved

keianhzo:
Should be ok now

bluemarvin:
This still doesn't work for me. STR:
1) Go into Private Browsing
2) Go into Focus Mode
3) Exit Private Browsing

Expected result: fade blitter is still in effect.
Actual result: fade blitter has been disabled.

bluemarvin:
So this does fix #146 but doesn't fix the case I out lined in my comment above. We can close this and I'll open a new issue.

PR #203, opened 2018-07-19
Fix tray collision check

### LINKED ISSUE #96: Restored commented code on Tray Intersection check
This check got commented while adding Tray support as it wasn't always returning. Probably related as how the collider normal is calculated. Research, fix and uncomment.

https://github.com/MozillaReality/FirefoxReality/blob/ac6f01547e519e05bb7b7a6158feaf89e1cfc6dc/app/src/main/cpp/Tray.cpp#L103-L106

keianhzo:
Fixes #96 

When dot product betweeen direction and collider face normal is bigger than the threshold we are pointing to the collider, so return otherwise.

### DIFF
--- a/app/src/main/cpp/Tray.cpp
+++ b/app/src/main/cpp/Tray.cpp
@@ -100,10 +100,10 @@ namespace {
 
     bool Intersect(const vrb::Vector& aStartPoint, const vrb::Vector& aDirection, vrb::Vector& aResult, bool& aInside) {
       const float dotNormals = aDirection.Dot(colliderNormal);
-//      if (dotNormals > -kEpsilon) {
-//        // Not pointed at the plane
-//        return false;
-//      }
+      if (dotNormals < -kEpsilon) {
+        // Not pointed at the plane
+        return false;
+      }
       const float dotV = (colliderMin - aStartPoint).Dot(colliderNormal);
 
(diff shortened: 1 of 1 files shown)


app/src/main/cpp/Tray.cpp
```diff
@@ -100,10 +100,10 @@ namespace {
 
     bool Intersect(const vrb::Vector& aStartPoint, const vrb::Vector& aDirection, vrb::Vector& aResult, bool& aInside) {
       const float dotNormals = aDirection.Dot(colliderNormal);
-//      if (dotNormals > -kEpsilon) {
-//        // Not pointed at the plane
-//        return false;
-//      }
+      if (dotNormals < -kEpsilon) {
```
bluemarvin: I don't think this is correct. Maybe the normals for the tray are wrong? For the intersection ray to even been in the same plane the angle between them must be obtuse. Since the collision ray should be pointed into the screen and the normal for the surface should be pointing out of the screen. If the normal for the surface points into the screen too, then the poly shouldn't even be visible.

MortimerGoro:
We are going to use Android UI for the tray, so this is not important now.

MortimerGoro:
The decision is made, we commented it in the design review meeting

PR #219, opened 2018-07-20
Support loading an environment from a known location. Fixes #207

### LINKED ISSUE #207: Support loading an environment from a known location
Suggestion: in order to make it easier for the art team to iterate on-device more quickly, it would be nice if we could probe a known location (e.g. `/sdcard/reality/FILE`) before loading the version of the environment in the app package. Then, they could iterate locally, `adb push` their changes, and restart the app to see it rather than having to drop it to the engineering team.

cc @MortimerGoro @keianhzo @thenadj

MortimerGoro:
Fixes #207

Depends on https://github.com/MozillaReality/vrb/pull/39/files

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
@@ -454,6 +454,11 @@ public class VRBrowserActivity extends PlatformActivity implements WidgetManager
         }, 20);
     }
 
+    @Keep
+    String getStorageAbsolutePath() {
+        return getExternalFilesDir(null).getAbsolutePath();
+    }
+
     void createOffscreenDisplay() {
         int[] ids = new int[1];
         GLES20.glGenTextures(1, ids, 0);
--- a/app/src/main/cpp/BrowserWorld.cpp
+++ b/app/src/main/cpp/BrowserWorld.cpp
@@ -46,12 +46,17 @@
 
 #include <array>
 #include <functional>
+#include <fstream>
 
 #define ASSERT_ON_RENDER_THREAD(X)                                          \
   if (m.context && !m.context->IsOnRenderThread()) {                        \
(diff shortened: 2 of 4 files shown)


MortimerGoro:
Example of injected environment in the data file of the app (via adb or Android File Transfer):

captura de pantalla 2018-07-20 a las 16 40 54

bluemarvin:
Issue #231 must be resolved before we ship.

MortimerGoro:
yes, it needs to be disabled for release. Some options:

- Enable only on debug builds
- Include a FxR dev mode (e.g. clicking on the logo in about) to show a menu with some toggle options.
- Manually disable it before ship

keianhzo:
I think enabling it only in debug builds is the most straightforward way without new UI and involving design.

bluemarvin:
Enabling on debug is probably okay. But I also want to insure we fix the ModelLoaderAndroid API and classes so that it is explicit where the model is being loaded from in code. I don't want to use the path as the switch for where the model gets loaded from.

larsbergstrom:
The goal here is to enable the designers - including those not on our team - to grab a build from the store and use it to iterate on environments while we work on building out more complete in-browser configuration UI.

I'm fine with a config flag as long as it's easy to turn on.

Sideloading custom builds is unfortunately proving to be extremely error-prone for non-developers, so I'd prefer to avoid grabbing builds from TC or a developer and asking designers to sideload them.

Is the concern here security or users getting attached to using it as a configuration point we have to then support forever? I'm not super concerned about the security issues, given that it's coming from local storage, but happy to do a formal review.

bluemarvin:
I'm not thrilled about shipping a backdoor that users don't explicitly turn on.

bluemarvin:
But there is also the code API, from reading the code right now there is no way to tell where the model is loaded from with out going into the guts of the implementation and then also know the path of what is being loaded. I would prefer it to be more explicit.

PR #243, opened 2018-07-24
Need to show FxR Nightly Version number in builds

### LINKED ISSUE #210: Need to show FxR Nightly Version number in builds
Hardware: all
Steps to Reproduce: N/A
Actual Behavior: Nightly version number not shown anywhere
Expected Behavior: Nightly version number is visible to end users in "About Firefox Reality" button, and is automatically submitted as part of bug report when user selects "Report an issue" from within FxR
Recommendation: Nightly version number is visible to end users in "About Firefox Reality" button, and is automatically submitted as part of bug report when user selects "Report an issue" from within FxR

keianhzo:
Fixes #210

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/SettingsWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/SettingsWidget.java
@@ -15,16 +15,16 @@ import android.widget.CompoundButton;
 import android.widget.ImageButton;
 import android.widget.Switch;
 import android.widget.TextView;
-
 import org.mozilla.geckoview.GeckoSession;
-import org.mozilla.vrbrowser.R;
-import org.mozilla.vrbrowser.SessionStore;
-import org.mozilla.vrbrowser.SettingsStore;
-import org.mozilla.vrbrowser.WidgetPlacement;
+import org.mozilla.vrbrowser.*;
 import org.mozilla.vrbrowser.audio.AudioEngine;
 
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+import java.util.GregorianCalendar;
+import java.util.TimeZone;
 
 public class SettingsWidget extends UIWidget {
     private static final String LOGTAG = "VRB";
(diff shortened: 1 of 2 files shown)


bluemarvin:
Why not just use BuildConfig.VERSION_CODE ? If it is 1 then it is a developer build.

bluemarvin:
I'm not getting the correct date. You can test by adding:

taskclusterBuild=true

To user.properties

bluemarvin:
Every thing is shifted by one. The leading number is for architecture. It's 1 for 32 bit arm right now.

```java
cal.set(Calendar.YEAR, Integer.parseInt(versionCode.substring(1, 2)) + 2016);
cal.set(Calendar.DAY_OF_YEAR, Integer.parseInt(versionCode.substring(2, 5)));
cal.set(Calendar.HOUR, Integer.parseInt(versionCode.substring(5, 7)));
cal.set(Calendar.MINUTE, Integer.parseInt(versionCode.substring(7, 9)));
formatted = format.format(cal.getTime());
```

Even with the everything shifted, I still get the day off by one. The day is 205 which should be July 24 but it is reporting July 25.

keianhzo:
Actually it wasn't taking into consideration leap years. Should be ok now.

bluemarvin:
Now the date is showing the day before. I wonder if timezone has something to do with it?

bluemarvin:
This seems to work for me:

```java
    private String versionCodeToDate(int aVersionCode) {
        if (aVersionCode == 1) {
            return getContext().getString(R.string.settings_version_developer);
        }

        String versionCode = Integer.toString(aVersionCode);
        GregorianCalendar cal = (GregorianCalendar)GregorianCalendar.getInstance();
        SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd");

        String formatted;
        try {
            int year = Integer.parseInt(versionCode.substring(1, 2)) + 2016;
            int dayOfYear = Integer.parseInt(versionCode.substring(2, 5));
            cal.set(Calendar.YEAR, year);
            cal.set(Calendar.DAY_OF_YEAR, dayOfYear);
            formatted = format.format(cal.getTime());

        } catch (StringIndexOutOfBoundsException e) {
            formatted = getContext().getString(R.string.settings_version_developer);
        }

        return formatted;
    }
```

keianhzo:
Yes, for some reason removing the time makes it work fine. We still need the leap year calculation though. I've updated with that fixed. Should work now.

bluemarvin:
The leap year change doesn't make sense to me. For example, if Jan 1 is 1, it will still be 1 on a leap year. Since the GregorianCalendar is being used, it should know about leap years and take that into account when turning the number of days into a month and year. Is that not the case?

keianhzo:
Yep, totally right, sorry for that, I was being stupid when testing.

PR #246, opened 2018-07-25
Fix #110: Security risk with Focus Mode

### LINKED ISSUE #110: Security risk  with Focus Mode
Current URL is hidden while in Focus mode. If the user clicks a link it doesn't see the real URL where he is navigating to.

We need to show the URL change someway

MortimerGoro:
Fix #110: Security risk with Focus Mode


MortimerGoro:
I didn't implement the show/hide on hover because programmatic hover listeners on Android were not very reliable. Maybe it's a problem with the MotionEventGenerator

PR #255, opened 2018-07-25
Updates styles to match design provided in UIS-5

caseyyee:
* Changes the layouts to be slightly  more generic in styles to easily fit more error page scenarios.  
* Alignment is default for easy localization.
* Uses system fonts.


keianhzo:
@cvan you are probably better suited for reviewing this, feel free to merge after you review.

PR #276, opened 2018-07-28
Check app button when in immersive mode. Fixes #275

### LINKED ISSUE #275: App Button fails to exit immersive mode
Hardware: GVR and Wave
Steps to Reproduce:
1) Enter Immersive mode
2) Press App Button
Actual Behavior: Nothing happens
Expected Behavior: Immersive mode is exited.
Recommendation: The application is not checking the app button for GVR and Wave when in immersive mode.

bluemarvin:


app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -178,11 +178,26 @@ struct BrowserWorld::State {
     fadeBlitter = FadeBlitter::Create(create);
   }
 
+  void CheckBackButton();
   void UpdateControllers(bool& aRelayoutWidgets);
   WidgetPtr GetWidget(int32_t aHandle) const;
   WidgetPtr FindWidget(const std::function<bool(const WidgetPtr&)>& aCondition) const;
 };
 
+void
```
MortimerGoro: Could we reuse CheckBackButton() function also from UpdateControllers()?

MortimerGoro: controller.lastButtonState complicated to do that

PR #282, opened 2018-07-30
Enable JS console output in debug builds

MortimerGoro:


bluemarvin:
Can you open a new issue and tag it design. We should be able to turn this on from a Release build.

PR #302, opened 2018-08-02
Add BuildConfig.GIT_HASH so that the git hash may be displayed in the settings menu

### LINKED ISSUE #301: Embed git hash in each build so we know what revision was used to build product.
Hardware: All
Recommendation: Make the git hash part of the BuildConfig object.

bluemarvin:
Fixes #301


bluemarvin:
We should make this available in the Settings panel so we know what commit was used to build a particular version.

Issue #291, opened 2018-08-01
Delay when entering WebVR in some demos


MortimerGoro:
Some demos (e.g. PlayCanvas Lab) have a noticeable delay when entering WebVR.

Did some research on GV code and found that VRDisplayHost::SubmitFrame() call exits a lot of times because  aFrameId and mDisplayInfo.mFrameId don't match

```
  // Ensure that we only accept the first SubmitFrame call per RAF cycle.
  if (!mFrameStarted || aFrameId != mDisplayInfo.mFrameId) {
    return;
  }
```
It happens for about 100 frames when the presentation starts, and then is ok. If we exit & enter WebVR again is fast. But if we reload and enter WebVR again same delay problem.

cvan:
Good catch! Is there an example site or reduced test case you can point to for me to reproduce this?

MortimerGoro:
I think that we can use this fix for now: https://github.com/MozillaReality/FirefoxReality/pull/307

The problem I see is that when a WebVR frame (the first frame in the STR example)  takes a longer time than the default threshold (50ms),  Gecko watchdog starts to generate new VR frames by calling NotifyVRVsync, which increases the frame number in VRDisplayHost::StartFrame. Because of that, later a ton of WebGL frames are rendered but not submitted (more than 100 frames in the STR example) until the WebGL frame catches up the VRDisplayHost::StartFrame() counter, which was way ahead. This was usefull in desktop to prevent kick offs from Oculus Rigt/Vie but not sure if it is helping in mobile . Also there is a chance that something else is still wrong.

The Gecko WebVR implementation is changing to move the RAF callbacks for WebVR out of the gecko IPC queues. So I don't think it's worth it to fix this in the GV with the current implementation. So I don't think that its worth it to try to fix this in Gecko until the new VRProcess architecture lands.

STR
* Without the patch
  * Open https://developer.playcanvas.com/en/tutorials/webvr-lab/
  * EnterVR is very delayed
* With the patch:
  * Open https://developer.playcanvas.com/en/tutorials/webvr-lab/
  * EnterVR is much faster

MortimerGoro:
Created this GV bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1480808

#307 improved the delay issue on the tested demo. But it causes a delay on the first frame on demos that were quick to enter (e.g. WebVR sample). It seems that something in the code is using the maxRAF value to trigger the first VR frame.

[...]

PR #307, opened 2018-08-02
Increase max WebVR raf. Fixes delay when entering WebVR in some demos.

MortimerGoro:
More info here:  https://github.com/MozillaReality/FirefoxReality/issues/291


bluemarvin:
Let's land this but open an bug on gecko WebVR. The watch dog should not cause frames renders to back up.

PR #313, opened 2018-08-03
Update the ExternalVR Wait

bluemarvin:
Updated the API but ran out of time to test it. I would recommend just trying it. Should return false on timeout.


MortimerGoro:
Tested and it seems to work ok

Issue #285, opened 2018-07-31
Refactor FxR use of VRExternalShmem for Immersive Mode


bluemarvin:
From @kearwood 

- Do not assume that there are any layers.  Always check that VRLayerState.type is equal to LayerType_Stereo_Immersive before accessing VRLayerState.layer_2d_content
- I’d suggest always doing the same thing when pushing and pulling state from the browser, and avoiding any logic in the PullBrowserState function that does anything other than update the struct.  (eg, the code that un-pauses the compositor is in the PullBrowserState function)
- Don’t assume that only one action will need to be taken in response to an updated browser state.  If you are blocking until a new frame is submitted, you will likely miss other signals, such as the user stopping VR presentation or navigating to another page.  This can lead to deadlocks.
- Using condition variables is a good idea, but perhaps we should be broadcasting (pthread_cond_broadcast rather than pthread_cond_signal).
- I’d suggest always using a condition variable when pushing and pulling state between Gecko and FxR, not just in some calls.  It should be up to the code that pulls the state to check the updated state for all changes that need to be acted on, and not just the particular condition it was waiting for.
- The VRBrowserState and VRSystemState structs should not be memcpy’ed from the shmem unless the mutex is actually locked.

MortimerGoro:
Closed via #328

Issue #263, opened 2018-07-26
Immersive Mode Appears to Crash Upon Entry when viewing WebVR Content


avrignaud:
Hardware: Oculus Go
Steps to Reproduce:

Here are the WebVR experiences that I tried and that didn’t work. It’s Task Definition Z8XVVBuASjevbCz3V77siQ and was pulled down and onto the Go at around 10:30 am this morning. All of these apps loaded their browser version and then when immersive mode is selected the device freezes and is unresponsive until it’s restarted.Song Exploder Presents: Inside Music (Audio works but nothing else)
Virtual Art Sessions in VR
Sketchfab Virtual Reality
WebVR LabI checked out “Check out My Moves on aframe.io/a-saturday-night” and it entered 360º and worked (tracking my head) but there was no interactivity and it seems as thought the text is mis-aligned in the game engine. 

Actual Behavior: All of these apps loaded their browser version and then when immersive mode is selected the device freezes and is unresponsive until it’s restarted.
Expected Behavior: Does not crash.
Recommendation: investigate and Fix. :)

MortimerGoro:
I'm going to investigate this. I also got a freeze with PlayCanvas based demos (After fixing a WebGL crash) https://github.com/MozillaReality/FirefoxReality/issues/260

Probably is the same issue in both.

fernandojsg:
A-Saturday-night is using a quite old A-Frame version (0.6.1) so it has many issues as positioning you incorrectly on the roof.
Sadly the demo is also not adapted for 3dof, so you could just watch dances from other people.
I believe there's no plan to upgrade the demo to a newer version of A-Frame.

MortimerGoro:
More info about the each:

### Virtual Art Sessions

That VR experience is only compatible with room scale capable devices. I found that they use device names to check that capability so in Oculus go browser and FxR (in Oculus) thinks that the device is the Oculus Rift and allows to enter vr. In the Mireage it doesn't let the user to EnterVR because they name check fails.

Because of this the website never submits a frame when entering WebVR. On the Oculus Go you can see that it shows a infinite loading spinner but the user can exit when clicking back. On FxR we have the same problem but we still need to implement https://github.com/MozillaReality/FirefoxReality/issues/306 to allow exit from a WebVR experience that never sends frames.

### Sketchfab Virtual Reality

[...]

Issue #259, opened 2018-07-26
Freeze when JavaScript doesn't call WebVR SubmitFrame


MortimerGoro:
FxR deadlocks if JavaScript code doesn't call SubmitFrame in WebVR raf.

https://github.com/MozillaReality/FirefoxReality/issues/258 cal help with this.

@kearwood also mentioned that we need to check if NotifyVSync is called when the compositor is paused while in WebVR. The GV VRManager class should deal with this situation.

MortimerGoro:
Closed via #328

Issue #257, opened 2018-07-26
WebVR link traversal animation


MortimerGoro:
We are going to need an animation for WebVR link traversal. We can start with something simple like a loading spin.

nancyhang:
@thenadj can we get a design for this please?

thenadj:
Slacked PNG to Imanol.

MortimerGoro:
Closed via #328

PR #328, opened 2018-08-07
Refactor immersive mode implementation for hang prevention and loading animations

### LINKED ISSUE #306: WebVR enter/exit loading animation flows
Some WebVR experiences can take some delay to start (specially the first frame), there can be some errors in the process or pages that don't call SubmitFrame. We need a good WebVR loading animation flow to handle all this situations while having a good user experience.

This is some proposal of the flow:

* Interval between we Enter VR and receive the first frame to render from the WebGLContent
   * Quick mutex timeout to check back button click and allow users to exit VR
   * Quick mutex timeout to render the loading animation spinner
* After showing the first frame from WebVR
  * Quick mutex timeout to check back button click and exit VR
  * Longer mutex timeout to detect a failed frame:
     * Skip the submit (hopefully the last submitted surface could be used with timewarp applied)
     * If more than x frames fail in a row we can show the loading animation again or a specific screen with a warning for this situation
* We could also add some transitions for Enter/Exit WebVR in the future

MortimerGoro:
* Fixes #306, #285, #263, #259, #257, #122 
* Depends on https://bugzilla.mozilla.org/show_bug.cgi?id=1481393 for a correct exit from WebVR in all the situations
* Link traversal doesn't work because of this: https://bugzilla.mozilla.org/show_bug.cgi?id=1481475
* The spinner is temporal, we'll add a rotation 3D model when ready

Some pages that can be used for testing:

- Presentation with no Submit Frames:
  - https://webvr.info/samples/test-slow-render.html?noFrames=1
  - https://virtualart.chromeexperiments.com/vr/

- Slow EnterVR 
  - https://developer.playcanvas.com/en/tutorials/webvr-lab/

- Presentation that stops submitting frames after some time (good to test the frame discard code on each platform)
  
  -[03-vr-presentation.html.zip](https://github.com/MozillaReality/FirefoxReality/files/2267103/03-vr-presentation.html.zip)


app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -793,37 +807,54 @@ BrowserWorld::DrawWorld() {
   m.drawListTransparent->Draw(*m.rightCamera);
   VRB_GL_CHECK(glDepthMask(GL_TRUE));
 #endif // !defined(VRBROWSER_NO_VR_API)
-  m.device->EndFrame();
+
+   m.device->EndFrame(false);
```
bluemarvin: Extra space on indent?

app/src/main/cpp/ExternalVR.cpp
```diff
@@ -128,17 +128,23 @@ struct ExternalVR::State {
...
     data.version = mozilla::gfx::kVRExternalVersion;
     data.size = sizeof(mozilla::gfx::VRExternalShmem);
-    pthread_mutex_init(&(data.systemMutex), nullptr);
-    pthread_mutex_init(&(data.browserMutex), nullptr);
-    pthread_cond_init(&(data.systemCond), nullptr);
-    pthread_cond_init(&(data.browserCond), nullptr);
+    pthread_mutex_init(&data.systemMutex, nullptr);
+    pthread_mutexattr_t attr;
+    pthread_mutexattr_init(&attr);
+    pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
```
bluemarvin: I would prefer we avoid recursive locks (http://www.fieryrobot.com/blog/2008/10/14/recursive-locks-will-kill-you/)

Why is this needed?

bluemarvin:
I really don't like the spinning wait animation. Maybe something that pulses would be better? A spinning  image in the dark is really uncomfortable in VR (for me). @thenadj

MortimerGoro:
> I think we need to re-think the locking strategy. I think we are overloading the mutexes. I think we might need four mutexes and two condition variable. One mutex to protect the browser data, one to protect the system state, one plus condition variable to block browser side, and one pause condition variable to block on system side. Using the same mutex to both protect data and block seems to be causing a lot of unnecessary issues. I realize this is more difficult because we need to change the Gecko side too.

mmm I think that adding more mutexes could add some race conditions. For example what would happen if Gecko pushes some change to the browser data in the interval between we Pull the browser data (exit condition not satisfied) and before locking using the mutex for the condition variable? I think that we'd miss the awake. That doesn't happen now because the mutex is released atomically when the condition variable starts the lock.

Other alternative would be to  create a private function (e.g.) called PullBrowserStateInternal() that does the logic that we don't want to duplicate in the code. We are only want to reuse some code with checks after pulling the browser data while not presenting and  when we are using the condition variable in a WebVR frame. This would be a variation of the lock argument we had before 

`I really don't like the spinning wait animation. Maybe something that pulses would be better? A spinning image in the dark is really uncomfortable in VR (for me). @thenadj`

I agree. I like the one in the Oculus browser better.

BTW, I had to add some transparent area on right & bottom of the PNG to make the center of the image the same as the center of the spiral. But it's still confusing.

bluemarvin:
The race condition is a good concern. I wonder if we should hide VRExternalShmem. Maybe instead of looking at VRExternalShmem directly, we spawn a thread that does a condition var wait on the VRBrowserState. Every time VRBrowserState is changed, the thread wakes up and figures out what has changed and passes that to the VR render thread as an event? Then the render thread does a timed wait on events. If there is no event, we can show the waiting animation?

app/src/main/cpp/LoadingAnimation.cpp
```diff
@@ -0,0 +1,93 @@
...
+      m.root->AddNode(m.spinner->GetRoot());
+    }
+  };
+
+  aLoader->RunLoadTask(nullptr, task, callback);
+}
+
+
+void
+LoadingAnimation::Update() {
```
bluemarvin: This reminds me, I want to update vrb::Updatable::UpdateResource() to get passed the delta time from last update so it can be used for animation nodes as well. https://github.com/MozillaReality/vrb/issues/34

bluemarvin:
The remainder of the code looks good. But I would really like to see if we can figure out a way to simply  how we interface with VRExternalShmem so the locking isn't so complicated.

MortimerGoro:
My concern with the event thread is that we may end up in the same situation to check/pull the events: mutex + wait locks, needing the same lock pattern.

I updated the PR removing the PTHREAD_MUTEX_RECURSIVE attribute by using the internal function I mentioned in the prev comment. I think it's a good pattern, the blogpost you shared recommends it too (http://www.fieryrobot.com/blog/2008/10/14/recursive-locks-will-kill-you). Also added some comments to make it easier to understand.

We can merge this for now and continue the locking conversation and the loading animation delta times as a follow-ups.

bluemarvin:
After sleeping on it, a separate thread might be overkill. The problem is that we can already miss things since nothing is queued. By having a dedicated thread we can create an event pump back to the render thread so we decrease the chance that we miss a state change in the VRExternalShmem.

avrignaud:
Note: once spinner model is animated (3D model), we should also remove the small spinner in the URL bar.

bluemarvin:
@andrenatal I think the spinner 3D model currently will only show when going into immersive mode. The spinner in the URL bar is probably a different issue?

Issue #320, opened 2018-08-04
The application icon on Wave in Vive Port is smaller than other application icons.


bluemarvin:
Hardware: Wave VR
Steps to Reproduce: View the FxR application in Vive Port
Actual Behavior: Application icon is smaller than other application icons.
Expected Behavior: Our icon should be the same shape and size as the other applications.
Recommendation: Need to investigate if there is a manifest flag or icon size that should be provided so the FxR application icon looks the same size and shape as other applications on the platform.

![waveicon](https://user-images.githubusercontent.com/3066/43670782-f631cace-9744-11e8-92be-db3f7c1204ec.png)

keianhzo:
Don't know why this wasn't closed when referenced from #338 @bluemarvin are you ok with closing this one?

PR #338, opened 2018-08-10
Update icon background size and color

### LINKED ISSUE #311: Update FxR icon back ground to match other assets?
@thenadj currently the background for the various FxR icons is white. Should it be change to the same purple being used for the app store assets?

keianhzo:
Fixes #311 #320 Updated background color and updated icon size.

screenshot 8

43922949-20e50c92-9bd5-11e8-9c22-b3726da34eb7


bluemarvin:
https://github.com/MozillaReality/FirefoxReality/blob/master/app/src/main/res/drawable/ffgvr_background.png

and

https://github.com/MozillaReality/FirefoxReality/blob/master/app/src/main/res/drawable/ffgvr_foreground.png

need to be updated. The background should be purple and the Fx logo should probably be smaller.

Here is a photoshop template:
https://developers.google.com/vr/downloads/stencil-examples.zip

Taken from:
https://developers.google.com/vr/distribute/daydream/publishing-requirements

Additionally, here is a foreground and background taken from a GVR example:
vr_icon
vr_icon_background

bluemarvin:
It looks like Wave doesn't use standard android icon correctly so we will need a custom one for the Focus. From looking at the example Wave applications, the icon back ground should not have any transparency and go edge to edge. It seems the home environment adds rounded corners to the full icon. So the Fx logo should probably also be a little bigger in this custom icon.

thenadj:
@bluemarvin Can this be done programmatically or do you need a new file from me?

bluemarvin:
@thenadj I think @keianhzo has everything he needs, unless you want something other than a solid color behind the logo.

keianhzo:
Fixed GVR and Wave Icons. I couldn't test the wave fix in the actual headset so please let me know if it's fixed.

bluemarvin:
gvr-icon

I think our Fx logo is too large on Daydream? All the other apps have much smaller top layer icons.

bluemarvin:
I've verified Wave is fixed.

thenadj:
Yes, it's too large. Please match size for consistency.

keianhzo:
Previous version was inside the Daydream template bounds for the icon but it seems that they leave plenty of padding. Updated Daydream size to better match the other platform icons.

Issue #267, opened 2018-07-27
Entering Private Mode does not display Private Mode details page


avrignaud:
Hardware: All
Steps to Reproduce: Enter private mode
Actual Behavior: Does not show private mode detail page
Expected Behavior: Should see new page with private mode details, same as in Firefox Desktop and in design doc
Recommendation: implement private mode page per design

bluemarvin:
Make sure the page is designed to work with issue #269 so that it may be localized.

keianhzo:
@thenadj I guess we could start with the design process for the Private Mode page while Gecko onLoadError API lands.

avrignaud:
@keianhzo Let me add @fernandojsg to this. Fernando, both this bug and https://github.com/MozillaReality/FirefoxReality/issues/269 need some html code work in order to be implemented. Is this something @caseyyee or @cvan might be able to help with? Think @MortimerGoro and @keianhzo will need to do some work after these pages are done to use strings.xml to create localized versions as well.

caseyyee:
@keianhzo Just a html page like the neterror.html? Let me know, can help throw that together.

keianhzo:
@caseyyee Yes, something like the Private Mode landing page for the Firefox Private mode, not sure what @thenadj has in mind for that. 

As soon as Gecko onLoadError lands this page will be localized. The initial localization approach was to load all the  strings from the android strings.xml file but I wonder if it's a more HTML natural way of dealing with localization instead of sending all the page strings as parameters when loading the URL. @caseyyee @cvan @bluemarvin 

Or do we prefer to centralize all the localization in the strings.xml?

avrignaud:
Adding @fernandojsg . I believe he's asked Casey to help here. Goal is to get strings into strings.xml for easier localization.

thenadj:
For now we are using the exact same page design/content/colors as Firefox has for Private Browsing.

bluemarvin:
This has nothing to do with the linked GV bug.

nancyhang:
I deleted it.

cvan:
Request: could we add a purple border to the window and the (X) close button to help better indicate the user is in Private Browsing?

keianhzo:
@avrignaud @ferjm @caseyyee What's the status on the HTML/CSS for the PB homepage?

caseyyee:
Alright, will make a PR into the same spot as the error pages, @keianhzo.  Will flag @thenadj for design review.

[...]

PR #170, opened 2018-07-16
[CXB-182] New error templates and code refactor

keianhzo:
CXB-182

Added the new error templates and made a code refactor of the error handling code.

PR #340, opened 2018-08-10
Privacy mode pages

caseyyee:
Partially addresses https://github.com/MozillaReality/FirefoxReality/issues/267

Will still need help from @keianhzo to hook up.

The styles are borrowed from https://github.com/MozillaReality/FirefoxReality/pull/170 with some slight changes.  Seems that DPI may have changed?  I needed to change default font sizing to something a little larger.


keianhzo:
@caseyyee Looks good to me, just not sure about the "bookmarks and downloads" part as we don't currently have any. Maybe we should remove that part for the moment.
Other than that I'll wait for @thenadj review to merge and hook up about:privatebrowsing

app/src/main/assets/html/private.html
```diff
@@ -0,0 +1,31 @@
...
+            <p>When you browse in a Private Window. Firefox Reality <strong>does not save</strong>:</p>
+
+            <ul class="two-column">
+                <li>visited pages</li>
+                <li>cookies</li>
+                <li>searches</li>
+                <li>temporary files</li>
+            </ul>
+
+            <p>Firefox Reality <strong>will save</strong> your:</p>
```
keianhzo: Don't currently have support for downloads and bookmarks. We should remove that part until we get that.

caseyyee: OK cool.  Will do that.

caseyyee:
@keianhzo 
Removed bookmarks and downloads bit.  Also bumped up the text size a little.

thenadj:
Looks good to me based on the screen shot. Thx @caseyyee

thenadj:
@caseyyee I checked the PB page in the latest build and there is a tiny little thing I'd like you to change. The two sets of bullets don't horizontally align correctly. I like the vertical set on the right which is a few pixels higher than the vertical set on the left. Match the left set so that they horizontally align and then it's ready. Thx!

Issue #121, opened 2018-07-11
Creating a new swap chain for immersive mode on Daydream fails


bluemarvin:
When creating a new swap chain on Daydream devices, rendering to the swap chain will intermittently fail. When the original swap chain is recreated it seems to work fine again.

MortimerGoro:
I was able to reproduce a similar issue in Oculus Go this way:

* Enter the WebVR sample demo
* Remove the headset and let it sleep
* Activate the headset again and resume the launch (hide the dev window)

I tried to replicate it more times but it worked ok.

bluemarvin:
I only see this issue on the Pixel. When I try the same code on the Lenovo Mirage immersive mode works as expected.

bluemarvin:
I was able to fix the Pixel issue as follows:

```c++
void
ExternalBlitter::EndFrame() {
  /*
  if (m.surface) {
    // We need to detach the SurfaceTexture to prevent the Gecko WebGL compositor from getting blocked.
    m.surface->DetachFromGLContext();
  }
   */
  if (m.surface) {
    m.surface->ReleaseTexImage();
    m.surface = nullptr;
  }
}

void
ExternalBlitter::StopPresenting() {

  if (m.surface) {
    m.surface->ReleaseTexImage();
    m.surface = nullptr;
  }
  for (std::map<const int32_t, GeckoSurfaceTexturePtr>::iterator iter = m.surfaceMap.begin(); iter != m.surfaceMap.end(); iter++) {
    if (iter->second) {
      iter->second->DetachFromGLContext();
    }
  }
  m.surfaceMap.clear();
}
```

After talking with @snorp, it seems that from his experience, calling DetachFromGLContext() after each frame is not reliable.

bluemarvin:
I will test this on other platforms tomorrow.

bluemarvin:
This should be fixed by #346

PR #346, opened 2018-08-11
Make WebVR work with multiprocess enabled

MortimerGoro:
The big patch is here: https://bugzilla.mozilla.org/show_bug.cgi?id=1482613

Do not merge this until that patch lands.

The patch also has some benefits for non-multiprocess:

- Custom SurfaceTexture sets are used for WebVR, no more conflicts with the compositor SurfaceTextures. We can even enable the compositor and WebVR works. This will be useful for link traversal (we may need to enable the compositor in the transition)
- We no longed need to dettach SurfaceTextures at the end of the VR frame. 
- It might help with #121 (didn't try yet)

I tested the implementation with multiprocess and non-multiprocess on Oculus Go, Vive and Pixel 1.

### DIFF
--- a/app/src/main/cpp/ExternalBlitter.cpp
+++ b/app/src/main/cpp/ExternalBlitter.cpp
@@ -81,10 +81,6 @@ ExternalBlitter::Create(vrb::CreationContextPtr& aContext) {
 void
 ExternalBlitter::StartFrame(const int32_t aSurfaceHandle, const device::EyeRect& aLeftEye,
                             const device::EyeRect& aRightEye) {
-  if (m.surface) {
-    m.surface->ReleaseTexImage();
-    m.surface = nullptr;
-  }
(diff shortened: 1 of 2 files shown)


MortimerGoro:
cc @fernandojsg It might be interesting to add multiprocess and non multiprocess modes to the benchmarks

fernandojsg:
@MortimerGoro would it be possible to launch it with some parameter to enable/disable it?

bluemarvin:
@fernandojsg We could add an intent parameter to disable multiprocess. I would expect e10s to be the only option at some point in the future (possibly after Fenix ships next year).

MortimerGoro:
@rbarker they are detached in the [GeckoSurfaceTexture destructor](https://github.com/MozillaReality/FirefoxReality/blob/master/app/src/main/cpp/GeckoSurfaceTexture.cpp#L212) when we exit WebVR

MortimerGoro:
I'm merging this to land the fix to not detach SurfaceTextures every frame. We can enable multiprocess in a different PR once the Video sharing issue is fixed

PR #368, opened 2018-08-16
Fix black flicker when widgets are shown. Fixes #367

### LINKED ISSUE #367: Black flicker happens when showing a widget
## Hardware
Oculus Go

## Steps to Reproduce
1. Open the settings panel from tray

## Current Behavior
During a split second a black flicker can be seen covering the whole widget surface texture. After that the widget is rendered fine.

## Expected Behavior
A black flickering shouldn't happen. Opening the settings panel is a common task an the flickering is quite noticeable for the end user.

## Context

ezgif com-video-to-gif

bluemarvin:
Prevent widgets from being rendered in the scene until the SurfaceTexture has content


bluemarvin:
@MortimerGoro gave the okay to merge since he will not be able to review until next week.

PR #392, opened 2018-08-21
Disable nsProfileLock sig handler to prevent conflict with Oculus controller. Fixes #322

### LINKED ISSUE #322: Pressing "o" button on Oculus Go controller on first run causes application exit.
Hardware: Go
Steps to Reproduce:
1) Uninstall FxR
2) Install FxR
3) Run FxR
4) Long press "o" button on controller
Actual Behavior: Application exits back to home environment.
Expected Behavior: Controller should be re-calibrated.
Recommendation: This is similar to the previous bug that was fixed by SDK 1.15 expect it is only the "o" button that is affected. Will probably need Oculus support since it is only reproducible on first run.

bluemarvin:


MortimerGoro:
Good catch!

PR #402, opened 2018-08-23
replace Home URL to hosted New-Tab/Feed page (fixes #324)

### LINKED ISSUE #324: Replace `DEFAULT_URL` with a remotely hosted, updatable page
FYI: **https://mixedreality.mozilla.org/fxr/feed** is currently just a placeholder.

First question, with GeckoView: if the protocol of the URL is `about:` or `resource://`, do we get more "chrome" privileges?

I'd like to have the start page that automatically updates. Easy hack that would work today:

```html
<iframe width="100%" height="100%" src="https://mixedreality.mozilla.org/fxr/feed/?flavor=%FLAVOR%>
```

I've been working on the frontend and backend architecture the past few days (super simple):

- https://github.com/MozillaReality/ContentFeed/issues/8#issuecomment-410030975
- https://github.com/MozillaReality/ContentFeed/issues/9
- https://github.com/MozillaReality/ContentFeed/milestone/2

And the week prior I worked on some design sketches, wireframes, and three visual options:
- https://github.com/MozillaReality/ContentFeed/issues/6#issuecomment-409829270

Anyway, ignore the placeholder design.

- https://mixedreality.mozilla.org/fxr/feed/?flavor=googlevr
- https://mixedreality.mozilla.org/fxr/feed/?flavor=noapi
- https://mixedreality.mozilla.org/fxr/feed/?flavor=oculusvr
- https://mixedreality.mozilla.org/fxr/feed/?flavor=svr
- https://mixedreality.mozilla.org/fxr/feed/?flavor=wavevr

FYI: I'll later want to change the URL to something else like https://mixedreality.mozilla.org/fxr/home/. Hope that's okay.

Wanted to kick off this conversation to determine the technical feasibility, best practices, and .

I worked on the Firefox Marketplace, and updating the "packaged app" via an `<iframe>` and local caching was something that wasn't anticipated when the project began. Anyway, the technology has improved a lot. And we can use Service Workers for offline capabilities, syncing clients with new content, and prefetching or even allowing download of content, images, videos, etc.

I have a lot of ideas here, but let's keep it simple for the MVP.

Opinions, concerns about replacing the `resource://android/assets/html/index.html`? I know currently the URL bar is blank now when you're on `index.html` page (thank you for that! :smile:). Anyway, should this landing page be called or be different from `about:home`? `about:xr`?

bluemarvin:


app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -30,7 +30,8 @@ public static SessionStore get() {
         }
         return mInstance;
     }
-    public static final String DEFAULT_URL = "resource://android/assets/html/index.html";
+    public static final String HOME_URL = "https://webxr.today/?region=";
+    public static String DEFAULT_URL = "https://webxr.today";
```
bluemarvin: I would prefer something like this:
```java
private static final String HOME_URL_COUNTRY_CODE_TEMPLATE = "https://webxr.today/?region=";
private static final String DEFAULT_HOME_URL = ""https://webxr.today";
private static String sHomeUrl = DEFAULT_HOME_URL;

public static void setHomeUrlContryCode(final String aCountryCode) {
    if (aCountryCode == null) {
        sHomeUrl = DEFAULT_URL;
        return;
    }
    sHomeUrl = HOME_URL_COUNTRY_CODE_TEMPLATE + aCountryCode;
}

public static String getHomeUrl() {
    return sHomeUrl;
}
```

And then replace `SesionStore.DEFAULT_URL` with `SesionStore.getHomeUrl()`

bluemarvin:
Did I submit this for a PR? I must have done that by accident when I was looking at the branch on my phone. 😓

bluemarvin:
@cvan you can close the PR if you want to wait until the patch is ready. Sorry, not sure how I managed to accidentally submit this as a PR from my phone.

cvan:
@bluemarvin no worries. in any event, thanks for opening a PR :+1:

it's still a work in progress, but could you take a look when you get a chance to see if this addresses things?

cvan:
> I think one problem with this is we probably try and load the home url before we know what the region code is so at start up, the home page will not have the region set and will be DEFAULT_URL. We could save the region in a pref and get that before loading the home url at start up? But that wouldn't help for first run. We could also load a tmp FxR splash page until we know what the country code is?

yeah, good call - that occurred to me while I was reviewing the code. would it be possible to `postMessage({region: state.mRegion}, SessionStore.HOME_NO_REGION_ORIGIN)` to the page, and I can listen for the message and accordingly update the region (i.e., the Portals/Top Sites and Experiences, etc.)?

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -30,7 +30,7 @@ public static SessionStore get() {
         }
         return mInstance;
     }
-    public static final String DEFAULT_URL = "resource://android/assets/html/index.html";
+    public static final String HOME_NO_REGION_ORIGIN = "https://webxr.today";
```
cvan: note: this is pointing to our staging server (which reflects [`master` of the `ContentFeed` repo](https://github.com/mozillareality/contentfeed/tree/master)). expect the domain to change very soon. it's just a placeholder for now.

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -307,23 +309,23 @@ public void removeSession(int aSessionId) {
     }
 
     private void pushSession(int aSessionId) {
-        boolean isPrivateMode  = mCurrentSession.getSettings().getBoolean(GeckoSessionSettings.USE_PRIVATE_MODE);
+        boolean isPrivateMode = mCurrentSession.getSettings().getBoolean(GeckoSessionSettings.USE_PRIVATE_MODE);
```
cvan: I can remove these changes if they clutter the PR. let me know.

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -347,7 +349,7 @@ public Integer getSessionId(GeckoSession aSession) {
         return null;
     }
 
-    public String gerUriFromSession(GeckoSession aSession) {
+    public String getUriFromSession(GeckoSession aSession) {
```
cvan: there was a typo here: `ger` -> `get`

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -402,6 +404,38 @@ public void setCurrentSession(int aId) {
             mCurrentSession.setActive(true);
     }
 
+    public String setRegion(String aRegion) {
+        Log.d(LOGTAG, "SessionStore setRegion: " + aRegion);
+        String result = "worldwide";
```
cvan: perhaps I should move the log statement to reflect what the region could be

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -53,6 +53,7 @@ public static SessionStore get() {
         boolean mIsInputActive;
         GeckoSession.ProgressDelegate.SecurityInformation mSecurityInformation;
         String mUri;
+        String mRegion;
```
cvan: should I be storing the state here or elsewhere?

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -66,6 +67,7 @@ public static SessionStore get() {
     private GeckoSession.PermissionDelegate mPermissionDelegate;
     private int mPreviousSessionId = SessionStore.NO_SESSION_ID;
     private String mLastUri;
+    private String mHomeUri = HOME_NO_REGION_ORIGIN;
```
cvan: is this fine to store as a member variable here?

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -402,6 +404,35 @@ public void setCurrentSession(int aId) {
             mCurrentSession.setActive(true);
     }
 
+    public String setRegion(String aRegion) {
+        Log.d(LOGTAG, "SessionStore setRegion: " + aRegion);
+        String result = "worldwide";
+        if (mCurrentSession == null || aRegion == null) {
+            return result;
+        }
+        State state = mSessions.get(mCurrentSession.hashCode());
```
cvan: is this necessary, or is there a better way?

app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
```diff
@@ -221,7 +221,7 @@ protected void onNewIntent(final Intent intent) {
     void loadFromIntent(final Intent intent) {
         final Uri uri = intent.getData();
         if (SessionStore.get().getCurrentSession() == null) {
-            String url = (uri != null ? uri.toString() : SessionStore.DEFAULT_URL);
+            String url = (uri != null ? uri.toString() : SessionStore.get().getHomeUri());
```
cvan: I thought about just changing `SessionStore#loadUri` to accept a null `uri` and keeping the conditional logic in that method. would make things a tad more DRY. thoughts?

app/src/common/shared/org/mozilla/vrbrowser/search/SearchEngine.java
```diff
@@ -83,6 +83,8 @@ private SearchEngine(@NonNull Context aContext) {
         String geolocationJson = SettingsStore.getInstance(mContext).getGeolocationData();
         GeolocationTask.GeolocationData data = GeolocationTask.GeolocationData.parse(geolocationJson);
         mEngine = Engine.getEngine(data);
+
+        SessionStore.get().setRegion(data.getCountryCode());
```
keianhzo: I think we should also set the region in onGeolocationRequestSuccess so we always get the most updated region after a geolocation request finishes.

app/src/googlevrFlat/java/org/mozilla/vrbrowser/BrowserActivity.java
```diff
@@ -173,7 +173,7 @@ private void loadFromIntent(final Intent intent) {
             final MyGeckoViewPermission permission = new MyGeckoViewPermission();
             permission.androidPermissionRequestCode = REQUEST_PERMISSIONS;
             mGeckoSession.setPermissionDelegate(permission);
-            uriValue = (uri != null ? uri.toString() : SessionStore.DEFAULT_URL);
+            uriValue = (uri != null ? uri.toString() : SessionStore.get().getHomeUri());
```
keianhzo: Yep, we will never get a region upon first run after install unless we use something like postMessage. 

You will get updated region info after the geolocation request finishes in subsequest home reloads though (after adding the setRegion call to onGeolocationRequestSuccess)

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
keianhzo: Isn't region a app wide setting not bound to a specific session? I guess we could handle it a SessionStore field?

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
keianhzo: Not sure why does setRegion need to return the region that's being set.

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
keianhzo: Commented below, region is a app wide setting so maybe there is no need of setting it per session and just do it as a SessionStore field?

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -402,6 +404,35 @@ public void setCurrentSession(int aId) {
...
+        if (state == null) {
+            return result;
+        }
+        result = state.mRegion = aRegion.toLowerCase();
+        return result;
+    }
+
+    public String getHomeUri() {
+        String result = SessionStore.HOME_WITHOUT_REGION_ORIGIN;
+        if (mCurrentSession == null) {
```
keianhzo: Region as a SessionStore field will remove all this session sanity checking logic as getHomeUri will always return the same region bound home uri regardless of the session state.

app/src/common/shared/org/mozilla/vrbrowser/search/SearchEngine.java
cvan: oh, good call

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
cvan: thanks, good call - yeah, it doesn't need to

cvan:
@keianhzo @bluemarvin thanks for the feedback. looks like I've got everything working.

the remaining item for the first-time region not being encoded in the URL (to use `postMessage`, etc.) - I've filed that as a follow-up issue: #409

thanks in advance for your help and patience. very much appreciated 👍

app/src/googlevrFlat/java/org/mozilla/vrbrowser/BrowserActivity.java
cvan: filed #409 - thank you!

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -414,29 +414,14 @@ public void setCurrentSession(int aId) {
...
-            return result;
-        }
-        State state = mSessions.get(mCurrentSession.hashCode());
-        if (state == null) {
-            return result;
-        }
-        result = state.mRegion = aRegion.toLowerCase();
-        return result;
+        aRegion = aRegion != null ? aRegion.toLowerCase() : "worldwide";
+        mRegion = aRegion;
```
keianhzo: nit: maybe just save an assignment and assign directly to the field?

app/src/common/shared/org/mozilla/vrbrowser/ui/NavigationBarWidget.java
```diff
@@ -494,9 +494,6 @@ public void onProgressChange(GeckoSession session, int progress) {
     public void onSecurityChange(GeckoSession geckoSession, SecurityInformation securityInformation) {
         if (mURLBar != null) {
             boolean isSecure = securityInformation.isSecure;
-            if (SessionStore.DEFAULT_URL.equalsIgnoreCase(SessionStore.get().gerUriFromSession(geckoSession))) {
```
cvan: this is because the Home URI is now served over HTTPS, so `isSecure` will already be `true`

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
cvan: yeah, probably should have - thanks

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
cvan: actually, I could remove this method completely - since I removed its only [use below](https://github.com/MozillaReality/FirefoxReality/pull/402/files#r212514020). for another time… 👍

PR #403, opened 2018-08-23
Compile fix for new onLoadError() GV API

bluemarvin:
The new GeckoView API has landed in nightly so update FxR so it will continue to compile on task cluster.


daoshengmu:
I am curious why I need to revert this patch when I am using `geckoViewLocal=/Projects/mozilla/geckoViewLocal/geckoview.aar` in user.properties.

bluemarvin:
It means your local build needs to be rebased with m-c tip.

PR #425, opened 2018-08-27
Added url bar clear support

### LINKED ISSUE #314: Missing clear button in URL bar
# Hardware: 
All
# Steps to Reproduce:
- Click on the URL bar
- Start typing
# Actual Behavior:
- You don't have any button to delete the whole URL bar text
# Expected Behavior:
- We should show a button so you can instantly delete all the text from the URL bar.
# Recommendation:
We should add a button by the voice input button to delete all text.

Fennec shows both he voice input button and the clear button at the same time at the end of the URL bar.

Other apps replace the voice input button for a clear button once there is any text in the text input. If all the text is deleted you get back the voice input button. 

The voice input button will always be visible in the keyboard anyway.

keianhzo:
Fixes #314 Added clear button and support for backspace long press clear. Also added back handler to the keyboard.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
@@ -656,6 +656,11 @@ public class VRBrowserActivity extends PlatformActivity implements WidgetManager
     }
 
     @Override
+    public void keyboardDismissed() {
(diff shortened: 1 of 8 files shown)


keianhzo:
Fixed

MortimerGoro:
It works with the keyboard exit button but now when the keyboard is dismissed clicking outside of it

keianhzo:
Fixed

Issue #377, opened 2018-08-21
Keyboard placement blocks URL bar


MortimerGoro:
## Hardware

Oculus Go

## Steps to Reproduce

1. Open the keyboard
2. Try to set the cursor or select characters in the URL bar

## Current Behavior

Touch conflicts with the keyboard and the URL bar

## Expected Behavior

No touch conflicts

## Possible Solution

Quick: Move the keyboard a bit
Best solution: Better placement solution for the keyboard

## Error Logs and Stack Traces

<details open>

```
```

</details>

avrignaud:
Keyboard is very close to URL bar, and aiming to select a character is very tricky. Explore moving keyboard away from URL bar. @thenadj

MortimerGoro:
Fixed via https://github.com/MozillaReality/FirefoxReality/pull/426

Issue #256, opened 2018-07-25
English Keyboard input improvement suggestions for FxR


avrignaud:
Hardware: All
Steps to Reproduce: N/A
Actual Behavior: N/A
Expected Behavior:
1. Expected to have "/" key available on main panel for easy URL entry
2. Expected to have ".com" button available to speed entry
3. Expected previously entered URL text to remain (so it is possible to add to it), and not have it erased
Recommendation:
1. / Key on main keyboard
2. .com button on main keyboard
3. URL text remains when selected unless backspace is selected (this presumes text is already selected)

thenadj:
Adding forward slash to keyboard improvements in Trello. 
Keeping URL text is a good idea. 
We are also adding a URL history UI to easily pick previously typed URL.

MortimerGoro:
Another user requested this: https://github.com/MozillaReality/FirefoxReality/issues/84

avrignaud:
All sounds good - and I also like the idea of a shift key on both sides of keyboard.  Apologies for missing the .com button!

cvan:
btw, filed issue #288 for Caps Lock/Shift issues

thenadj:
Just FYI - This is currently #6 on our design priority list.

MortimerGoro:
Have we considered using a split keyboard like Oculus/Chrome does (e.g numbers on a side of the keyboard, instead of a keyboard layout switch). 

Maybe you mentioned it a design review meeting but don't remember now.

thenadj:
@MortimerGoro Yes we have but we will not make changes for v1. It's on or list for v1.1 to explore split keyboard.

thenadj:
Updated keyboard design is complete. 
[UIS-6 Keyboard_011.pdf](https://github.com/MozillaReality/FirefoxReality/files/2308545/UIS-6.Keyboard_011.pdf)

MortimerGoro:
> URL text remains when selected unless backspace is selected (this presumes text is already selected

Can you give more details about this issue? I think that clearing the selected text after typing is a common behavior on browsers.

avrignaud:
[...]

PR #426, opened 2018-08-27
Fix keyboard placement blocks URL bar. Fixes ##377

MortimerGoro:


MortimerGoro:
Fixes #377 and point 6 from #256

PR #428, opened 2018-08-27
Add minimum scroll threshold

MortimerGoro:
Makes easier to start clicking on links using the touchpad

### DIFF
--- a/app/src/main/cpp/BrowserWorld.cpp
+++ b/app/src/main/cpp/BrowserWorld.cpp
@@ -72,6 +72,21 @@ static const int GestureSwipeLeft = 0;
 static const int GestureSwipeRight = 1;
 
 static const float kScrollFactor = 20.0f; // Just picked what fell right.
+// Deadzone used when clicking so that buttons are easier to activate.
+static const float kSrollDeadZone = 20.0f; //
+// Minimum scroll threshold to start scroll gestures.
+// The smaller, the more sensitive it is the scroll.
+// A very sensitive scroll can make it hard to click links, specially
+// on devices which don't have a trigger button (e.g. Daydream)
+#if defined(VRBROWSER_GOOGLEVR)
+static const float kMinScrollThreshold = 0.01f;
+static const int kMinScrollFrames = 1;
+#else
+static const float kMinScrollThreshold = 0.00f;
+static const int kMinScrollFrames = 0;
+#endif // defined(GOOGLEVR)
+
+
 static const float kWorldDPIRatio = 2.0f/720.0f;
 
 #if SPACE_THEME == 1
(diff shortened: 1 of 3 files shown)


bluemarvin:
Slow scrolling on the Go is also very jerky. I don't think we want to have a scroll threshold or at least it needs to only be at the start of the scroll event and should not be applied over the entire time the touch pad is touched. Even then, I'm not sure how helpful it will be since some people never take their thumb off of the touch pad.

MortimerGoro:
I'll try doing at the start only. It's specially for Daydream (we can set 0.0 on others) since I find it difficult to click links because it gets scrolled a bit before pressing

bluemarvin:
The dead zone was supposed to help since we stop updating the touch X/Y during the press so as long as the pointer is over the target after the slight wiggle the click should still work. The problem is mostly with really small targets.

app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -68,6 +68,21 @@ static const int GestureSwipeLeft = 0;
 static const int GestureSwipeRight = 1;
 
 static const float kScrollFactor = 20.0f; // Just picked what fell right.
+// Deadzone used when clicking so that buttons are easier to activate.
+static const float kSrollDeadZone = 20.0f; //
+// Minimum scroll threshold to start scroll gestures.
+// The smaller, the more sensitive it is the scroll.
+// A very sensitive scroll can make it hard to click links, specially
+// on devices which don't have a trigger button (e.g. Daydream)
+#if defined(GOOGLEVR)
```
bluemarvin: I think this should be `#if defined(VRBROWSER_GOOGLEVR)`

bluemarvin:
One thing that might make it feel less jumpy is to ignore scroll *after* a click for a short period?

MortimerGoro:
> One thing that might make it feel less jumpy is to ignore scroll after a click for a short period?

The problem for is before the click. e.g on a google search I want to click a link, I press the touchpad, but before the click it scrolls a bit and a different link is clicked. I usually have my thumb off of the touch pad, so it can be a different problem from people having it always on.

I reduced the values to only use one frame and a smaller threshold. This is only applied in Daydream.

I don't want it to affect scroll responsiveness. Let me know if you still feel that with the latest values. I'm testing on a Google Pixel 1, maybe is more sensitive on Mirage for the better overall performance.

bluemarvin:
oops, edited wrong issue. I'm still looking at this one.

cvan:
is this good to be merged? or should it be closed?

larsbergstrom:
@bluemarvin @MortimerGoro 

Are we going to proceed with this, or should it be punted to post-1.1 or closed?

bluemarvin:
@MortimerGoro Do we still want to do this? It would seem the trend is to move away from touch surfaces on controllers. Also, Daydream is looking like a dead platform.

MortimerGoro:
We can close this, it only affected old daydream trackpad touches, so it's not a priority

PR #436, opened 2018-08-28
Pause media when the app goes to background.

### LINKED ISSUE #429: stop audio/video when backgrounded.
## Steps to Reproduce

1. Watch video on CNN

## Current Behavior

CNN.com video just keep playing until FxR is killed.

## Expected Behavior

CNN.com should stop playing when backgrounded or page exited

MortimerGoro:
Fixes #429.

Gecko side PR: https://phabricator.services.mozilla.com/D4472
We need to wait for it to land this

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
@@ -176,11 +176,13 @@ public class VRBrowserActivity extends PlatformActivity implements WidgetManager
     @Override
     protected void onPause() {
         mAudioEngine.pauseEngine();
+        SessionStore.get().setActive(false);
         super.onPause();
     }
 
     @Override
     protected void onResume() {
+        SessionStore.get().setActive(true);
         mAudioEngine.resumeEngine();
         super.onResume();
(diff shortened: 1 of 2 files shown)


app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
```diff
@@ -176,11 +177,19 @@ protected void onStop() {
     @Override
     protected void onPause() {
         mAudioEngine.pauseEngine();
+        GeckoSession current = SessionStore.get().getCurrentSession();
```
bluemarvin: I think SessionStore.setActive() would be better since in the future we will probably have more than just one active GeckoSession. Also, this potentially won't work if there are multiple sessions on the store stack from a onCreateNewSession call.

bluemarvin:
We will need to remember to bump the GV version number once the GV dependent bug lands before we can land this PR.

bluemarvin:
@MortimerGoro I have upgraded the pinned GV version and tested that this patch compiles against it. Once you address the review comments it's ready to land 🙂

MortimerGoro:
PR updated

PR #441, opened 2018-08-29
Fix WebVR remote controller names

### LINKED ISSUE #434: Hand controller doesn't work in Three.js VR demos
## Steps to Reproduce
Try Three.js VR examples

## Current Behavior
The hand controller doesn't work (it works on Oculus Browser)

## Expected Behavior
Hand controller should work

MortimerGoro:
`Gear VR Controller` is used in Oculus and Vive Focus
`Daydream Controller` is used in Daydream

Fixes  #434


fernandojsg:
Cool! that should fix our compatibility issues with previous content :+1:

Issue #433, opened 2018-08-28
Research stale frames on WebVR


MortimerGoro:
Some demos such as  360 panorama have about ~20 stale frames per second. Staled frames should be 0 in such a simple demo.

A potential cause: I saw that intermittently some WebGL RAF frames took more time than needed. We need to know why that happens.

MortimerGoro:
@daoshengmu is researching this issue

daoshengmu:
I don't see there is any delay in the IPC channel. The avg. time only takes 0.7 ms. Beside,  it utilizes SharedTextureEGL to forward textures from WebGL to the VR module. I am wondering if we use incorrect APIs from the Oculus part.

```
08-29 17:35:03.820 27235 27258 I vrsubmit: SendSubmitFrame, type 7, frame id 3576, timestamp 11431896.116117
08-29 17:35:03.821 27235 27392 I vrsubmit: RecvSubmitFrame, frame id 3576, timestamp 11431896.401273
08-29 17:35:03.821 27235 27391 I vrsubmit: VRDisplayExternal::SubmitFrame frame id 3576
08-29 17:35:03.841 27235 27258 I vrsubmit: SubmitFrame type: 9
08-29 17:35:03.841 27235 27258 I vrsubmit: SendSubmitFrame, type 7, frame id 3577, timestamp 11431916.667733
08-29 17:35:03.842 27235 27392 I vrsubmit: RecvSubmitFrame, frame id 3577, timestamp 11431917.484296
08-29 17:35:03.842 27235 27391 I vrsubmit: VRDisplayExternal::SubmitFrame frame id 3577
08-29 17:35:03.861 27235 27258 I vrsubmit: SubmitFrame type: 9
08-29 17:35:03.861 27235 27258 I vrsubmit: SendSubmitFrame, type 7, frame id 3578, timestamp 11431936.914715
08-29 17:35:03.862 27235 27392 I vrsubmit: RecvSubmitFrame, frame id 3578, timestamp 11431937.339090
08-29 17:35:03.862 27235 27391 I vrsubmit: VRDisplayExternal::SubmitFrame frame id 3578
08-29 17:35:03.877 27235 27258 I vrsubmit: SubmitFrame type: 9
08-29 17:35:03.877 27235 27258 I vrsubmit: SendSubmitFrame, type 7, frame id 3579, timestamp 11431952.738935
[...]

PR #453, opened 2018-08-31
Enable Oculus Go extra latency mode when entering immersive mode.

daoshengmu:
Per #433.

We need to enable ATW to handle stale frames for some situation like WebVR, video playback. As Oculus GO SDK mentioned, asynchronous time warp (ATW) is the best way to handle this[1].

[1] Latency, https://developer.oculus.com/documentation/mobilesdk/latest/concepts/mobile-vrapi/

### DIFF
--- a/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
+++ b/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
@@ -307,6 +307,16 @@ DeviceDelegateOculusVR::SetRenderMode(const device::RenderMode aMode) {
   if (aMode == m.renderMode) {
     return;
   }
+  if (aMode == device::RenderMode::Immersive) {
+    // Enable extra latency mode. We have stale frames when running
+    // WebVR immersive mode especially for video playback. I have checked
+    // VRAPI_EXTRA_LATENCY_MODE_ON is better than VRAPI_EXTRA_LATENCY_MODE_DYNAMIC
+    // when running video playback.
+    vrapi_SetExtraLatencyMode(m.ovr, ovrExtraLatencyMode::VRAPI_EXTRA_LATENCY_MODE_ON);
+  } else {
+    // Turn it off after exiting from the immersive mode.
+    vrapi_SetExtraLatencyMode(m.ovr, ovrExtraLatencyMode::VRAPI_EXTRA_LATENCY_MODE_OFF);
+  }
   m.renderMode = aMode;
   m.SetRenderSize(aMode);
   vrb::RenderContextPtr render = m.context.lock();


app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -539,6 +539,11 @@ DeviceDelegateOculusVR::EnterVR(const crow::BrowserEGLContext& aEGLContext) {
     vrapi_SetClockLevels(m.ovr, 4, 4);
     vrapi_SetPerfThread(m.ovr, VRAPI_PERF_THREAD_TYPE_MAIN, gettid());
     vrapi_SetPerfThread(m.ovr, VRAPI_PERF_THREAD_TYPE_RENDERER, gettid());
+    // Enable asynchronous time warp (ATW).
```
MortimerGoro: nit: As I understand, ATW is always enabled. This set's a extra latency mode for ATW, doesn't enable ATW

MortimerGoro:
I tried this on other demos (e.g. Playcanvas WebVR lab) and it reduces the stale frames.

Should we enable this only while in WebVR? Because we normally have 0 stale frames in the 2D Browser scene

daoshengmu:
Is there a status for me to know I am running a WebVR website?

MortimerGoro:
@daoshengmu yes, you can change it in` DeviceDelegateOculusVR::SetRenderMode()`

MortimerGoro:
I checked if Oculus Browser and Samsung Internet enable this. I think it's the `Lat` parameter in the logcat performance logs. (Lat = 0: disabled, Lat = 1: Enabled, Lat = 2: Dynamic)

Samsung Internet enables VRAPI_EXTRA_LATENCY_MODE_ON in their home scene. While on WebVR it says it's disabled but we think that the stats are bad because they always show 60fps even in demos with tons of reprojection.

It seems that Oculus Browser always has it disabled.

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -307,6 +307,13 @@ DeviceDelegateOculusVR::SetRenderMode(const device::RenderMode aMode) {
   if (aMode == m.renderMode) {
     return;
   }
+  if (aMode == device::RenderMode::Immersive) {
```
MortimerGoro: You also need to disable extra latency mode when the render mode is set back to Standalone

daoshengmu:
We probably could turn it off after fixing the video playback performance issue. But, for now, the latency mode can solve our stale frame issue.

I am wondering why Samsung Internet chooses to turn it off for WebVR immersive mode? The immersive mode is usually the case that would have performance issue.

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
daoshengmu: Thanks for correction.

bluemarvin:
Adding more latency seems like a band-aid fix. Where this latency in WebGL is coming from is far more important to figure out.

daoshengmu:
For the case of 360 video, we only have one WebGL draw call. The stale frame is because of waiting for the video frame. The rAF takes 8 ms. Avg. frame render time is 18ms. They are very close to the stats of the SDK document. I believe we definitely have WebGL performance issue, but WebGL 360 video has used SharedSurface so it shouldn’t be the main reason.

bluemarvin:
Video is a known issue on the Go. I thought we were getting stale frames even when the video decoder was not part of the render loop.

MortimerGoro:
We can use those ones to test stale frames without the video being involved:
- https://developer.playcanvas.com/en/tutorials/webvr-lab/
- https://aframe.io/examples/showcase/sky/

daoshengmu:
I have checked the links
- https://developer.playcanvas.com/en/tutorials/webvr-lab/
- https://aframe.io/examples/showcase/sky/

are improved a lot after enabling the extra latency mode. I feel the performance issue mostly comes from the immersive mode especially when rotating the viewport. I am going to investigate deeply the code path of immersive mode.

cvan:
is this good to be merged? or should it be closed?

MortimerGoro:
@fernandojsg Could you test the impact of this on all the engines using the benchmarking tools? Please rebase the branch to compare with the same GV used in master

philip-lamb:
Per discussion with @daoshengmu we'll drop this approach and focus on other specific performance-related cases elsewhere.

Issue #443, opened 2018-08-29
Expose history to the content feed


fernandojsg:
As proposed for the bookmarks (https://github.com/MozillaReality/FirefoxReality/issues/442), but for the history. 
The content feed should be able to remove content from the list. It doesn't necessarily need to be a two-ways api as we could just blacklist some urls when deleting them from the content feed on the localstorage.

keianhzo:
GV is getting a History API https://bugzilla.mozilla.org/show_bug.cgi?id=1494713. When this is closed we should open a new issue to track that and update.

philip-lamb:
https://bugzilla.mozilla.org/show_bug.cgi?id=1494713 is done.

cvan:
See discussion from the original PR from prior to the implementation of the GeckoView History API: https://github.com/MozillaReality/FirefoxReality/pull/457#issuecomment-447424317

This bug needs attention re: Privacy, Security, and UX. We may be able to close this. But even if we conclude that the Content Feed should feature History (instead of accessible from `about:history`, à la `about:bookmarks` / the Bookmarks button in the Tray), the work needs to be done in #614 to first use the android-components code.

jvonitter:
tracking this in #759

PR #457, opened 2018-08-31
Implement history and expose it to content feed

MortimerGoro:
Example URL
`https://webxr.today/?history=W3sidSI6ImFib3V0OmJsYW5rIiwibiI6IiIsInQiOjE1MzU3Mjc1ODE5NDR9LHsidSI6Imh0dHBz%0AOlwvXC9hZnJhbWUuaW9cLyIsIm4iOiJBLUZyYW1lIOKAkyBNYWtlIFdlYlZSIiwidCI6MTUzNTcy%0ANzQxODc5OX0seyJ1IjoiaHR0cHM6XC9cL3dlYnZyLmluZm9cL3NhbXBsZXNcLyIsIm4iOiJXZWJW%0AUiAtIFNhbXBsZXMiLCJ0IjoxNTM1NzI3NDU5NjQ0fSx7InUiOiJodHRwczpcL1wvYWZyYW1lLmlv%0AXC9leGFtcGxlc1wvIiwibiI6IkhlbGxvIFdlYlZSIOKAkyBBLUZyYW1lIiwidCI6MTUzNTcyNzQy%0AMjQ3Nn0seyJ1IjoiaHR0cHM6XC9cL3d3dy5nb29nbGUuY29tXC9zZWFyY2g%2FcT13ZWJ2ciUyMHNh%0AbXBsZXMmaWU9dXRmLTgmb2U9dXRmLTgma2V5d29yZD1maXJlZm94LWItYWImc2VhcmNoYmFyPWZp%0AcmVmb3gtYiIsIm4iOiJ3ZWJ2ciBzYW1wbGVzIC0gQnVzY2FyIGNvbiBHb29nbGUiLCJ0IjoxNTM1%0ANzI3NDUwNDEzfV0%3D%0A`

If we decode history parameter from urlEncoded base64

`[{"u":"about:blank","n":"","t":1535727581944},{"u":"https:\/\/aframe.io\/","n":"A-Frame – Make WebVR","t":1535727418799},{"u":"https:\/\/webvr.info\/samples\/","n":"WebVR - Samples","t":1535727459644},{"u":"https:\/\/aframe.io\/examples\/","n":"Hello WebVR – A-Frame","t":1535727422476},{"u":"https:\/\/www.google.com\/search?q=webvr%20samples&ie=utf-8&oe=utf-8&keyword=firefox-b-ab&searchbar=firefox-b","n":"webvr samples - Buscar con Google","t":1535727450413}]`

Fixes https://github.com/MozillaReality/FirefoxReality/issues/443

Notes: To clear the history the content feed can filter based on timestamps

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/History.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/History.java
@@ -80,13 +80,17 @@ public class History implements GeckoSession.ProgressDelegate {
         save();
     }
 
+    public String getHistoryPath() {
+        return mContext.getFilesDir().getAbsolutePath() + "/" + FILENAME;
+    }
+
     private void save() {
         JSONArray array = new JSONArray();
         for (Item item: mItems.values()) {
             array.put(item.toJSON());
         }
 
-        try (FileWriter file = new FileWriter(mContext.getFilesDir().getAbsolutePath() + "/" + FILENAME)) {
+        try (FileWriter file = new FileWriter(getHistoryPath())) {
             file.write(getJSONString());
         }
         catch (Exception ex) {
@@ -95,7 +99,7 @@ public class History implements GeckoSession.ProgressDelegate {
     }
 
     private void restore() {
-        try (BufferedReader reader = new BufferedReader(new FileReader(mContext.getFilesDir().getAbsolutePath() + "/" + FILENAME))) {
+        try (BufferedReader reader = new BufferedReader(new FileReader(getHistoryPath()))) {
             StringBuilder sb = new StringBuilder();
             String line = reader.readLine();
 
@@ -140,18 +144,20 @@ public class History implements GeckoSession.ProgressDelegate {
         return result;
     }
 
+    private static class SortItemsByTimestamp implements Comparator<Item> {
+        @Override
+        public int compare(Item o1, Item o2) {
+            return (int)(o1.timestamp - o2.timestamp);
+        }
+    }
+
     private void checkMaxSize() {
         if (mItems.size() <= MAX_ITEMS) {
             return;
         }
 
         ArrayList<Item> list = new ArrayList<>(mItems.values());
-        list.sort(new Comparator<Item>() {
-            @Override
-            public int compare(Item o1, Item o2) {
-                return (int)(o1.timestamp - o2.timestamp);
-            }
-        });
+        list.sort(new SortItemsByTimestamp());
 
         for (Item item: list) {
             mItems.remove(key(item.url));
@@ -171,14 +177,11 @@ public class History implements GeckoSession.ProgressDelegate {
     @Override
     public void onPageStop(GeckoSession geckoSession, boolean b) {
         String uri = SessionStore.get().getUriFromSession(geckoSession);
-        if (uri == null || uri.length() == 0) {
-            return;
-        }
-        if (SessionStore.get().isHomeUri(uri)) {
+        if (uri == null || uri.length() == 0 || SessionStore.get().isHomeUri(uri)) {
             return;
         }
         if (geckoSession.getSettings().getBoolean(GeckoSessionSettings.USE_PRIVATE_MODE)) {
-            // Do not store private mode history
+            // Do not store history in Private Mode.
             return;
         }
 
@@ -205,6 +208,5 @@ public class History implements GeckoSession.ProgressDelegate {
     public void onSecurityChange(GeckoSession geckoSession, SecurityInformation securityInformation) {
 
     }
-
 
 }
--- a/app/src/common/shared/org/mozilla/vrbrowser/History.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/History.java
@@ -1,0 +1,210 @@
+package org.mozilla.vrbrowser;
+
+import android.content.Context;
+import android.net.Uri;
+import android.util.Base64;
+import android.util.Log;
+
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.mozilla.geckoview.GeckoSession;
+import org.mozilla.geckoview.GeckoSessionSettings;
+
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.HashMap;
+
+public class History implements GeckoSession.ProgressDelegate {
+    private HashMap<String, Item> mItems;
+    private static final int MAX_ITEMS = 30;
+    private static final String LOGTAG = "VRB";
+    private static final String FILENAME = "history.json";
+
+    private Context mContext;
+
+    public static class Item {
+        String url;
+        String title;
+        long timestamp;
+
+        Item() {
+            timestamp = System.currentTimeMillis();
+        }
+
+        Item(JSONObject aObject) {
+            url = aObject.optString("u", "");
+            title = aObject.optString("n", "");
+            timestamp = aObject.optLong("t", System.currentTimeMillis());
+        }
+
+        JSONObject toJSON() {
+            JSONObject obj = new JSONObject();
+            try {
+                obj.put("u", url);
+                obj.put("n", title);
+                obj.put("t", timestamp);
+            }
+            catch (JSONException ex) {
+            }
+            return obj;
+        }
+    }
+
+    public History(Context aContext) {
+        mContext = aContext;
+        mItems = new HashMap<>();
+        SessionStore.get().addProgressListener(this);
+        restore();
+    }
+
+    public void updateContext(Context aContext) {
+        mContext = aContext;
+    }
+
+    public void release() {
+        SessionStore.get().removeProgressListener(this);
+    }
+
+    public String encodedJSON() {
+        String value = getJSONString();
+        String base64 = Base64.encodeToString(value.getBytes(), 0);
+        return Uri.encode(base64);
+    }
+
+    public void clear() {
+        mItems.clear();
+        save();
+    }
+
+    private void save() {
+        JSONArray array = new JSONArray();
+        for (Item item: mItems.values()) {
+            array.put(item.toJSON());
+        }
+
+        try (FileWriter file = new FileWriter(mContext.getFilesDir().getAbsolutePath() + "/" + FILENAME)) {
+            file.write(getJSONString());
+        }
+        catch (Exception ex) {
+            Log.e(LOGTAG, "Error saving history", ex);
+        }
+    }
+
+    private void restore() {
+        try (BufferedReader reader = new BufferedReader(new FileReader(mContext.getFilesDir().getAbsolutePath() + "/" + FILENAME))) {
+            StringBuilder sb = new StringBuilder();
+            String line = reader.readLine();
+
+            while (line != null) {
+                sb.append(line);
+                sb.append("\n");
+                line = reader.readLine();
+            }
+            try {
+                JSONArray array = new JSONArray(sb.toString());
+                mItems.clear();
+                for (int i = 0; i < array.length(); ++i) {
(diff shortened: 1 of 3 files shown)


MortimerGoro:
I'd prefer use javascript to inject the data but there is no way to do that in GV

MortimerGoro:
cc @cvan can web user webxr.today as a local page? (downloading the html on the fly)

There are security concerns about the GET query with this sensitive data

cvan:
@MortimerGoro: this is the perfect use case to use hashes (`#`) instead of query-strings (`?`).

hashes are not submitted to servers, so there will be no lots of any data passed after a hash.

sound good?

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -425,6 +448,14 @@ public String getHomeUri() {
         if (mRegion != null) {
             result = SessionStore.HOME_WITHOUT_REGION_ORIGIN + "/?region=" + mRegion;
         }
+        if (mHistory != null) {
+            if (!result.contains("?")) {
+                result += "/?";
+            } else {
+                result += "&";
+            }
+            result += "history=" + mHistory.encodedJSON();
```
cvan: I would remove the above four lines and change this to #history

MortimerGoro:
@cvan good point! I updated the PR to use the hash

app/src/common/shared/org/mozilla/vrbrowser/History.java
```diff
@@ -0,0 +1,210 @@
...
+
+    }
+
+    @Override
+    public void onPageStop(GeckoSession geckoSession, boolean b) {
+        String uri = SessionStore.get().getUriFromSession(geckoSession);
+        if (uri == null || uri.length() == 0) {
+            return;
+        }
+        if (uri.toLowerCase().startsWith(SessionStore.HOME_WITHOUT_REGION_ORIGIN)) {
```
cvan: I made a [helper method](https://github.com/MozillaReality/FirefoxReality/blob/5b8a3340/app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java#L457-L459) for this purpose. you can replace this with `if (SessionStore.get().isHomeUri(uri)) {`

app/src/common/shared/org/mozilla/vrbrowser/History.java
```diff
@@ -0,0 +1,210 @@
...
+
+    @Override
+    public void onPageStart(GeckoSession geckoSession, String s) {
+
+    }
+
+    @Override
+    public void onPageStop(GeckoSession geckoSession, boolean b) {
+        String uri = SessionStore.get().getUriFromSession(geckoSession);
+        if (uri == null || uri.length() == 0) {
```
cvan: I don't know Java at all, but can you do `uri.isEmpty()`?

app/src/common/shared/org/mozilla/vrbrowser/History.java
```diff
@@ -0,0 +1,210 @@
...
+        return result;
+    }
+
+    private void checkMaxSize() {
+        if (mItems.size() <= MAX_ITEMS) {
+            return;
+        }
+
+        ArrayList<Item> list = new ArrayList<>(mItems.values());
+        list.sort(new Comparator<Item>() {
```
cvan: can we move this callback to a private method?

app/src/common/shared/org/mozilla/vrbrowser/History.java
```diff
@@ -0,0 +1,210 @@
...
+        JSONArray array = new JSONArray();
+        for (Item item: mItems.values()) {
+            array.put(item.toJSON());
+        }
+        return array.toString();
+    }
+
+    private String key(String aUri) {
+        String result = aUri.toLowerCase();
+        int index = result.indexOf('#');
```
cvan: it'd be nice if we could use some helper/util method for extracting the hash instead of doing fragile string substring matching and slicing.

out of curiosity, does Java have any URI parsing libraries (like [JavaScript's `URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) or [Node's `URL`](https://nodejs.org/api/url.html))?

app/src/common/shared/org/mozilla/vrbrowser/History.java
```diff
@@ -0,0 +1,210 @@
...
+                sb.append(line);
+                sb.append("\n");
+                line = reader.readLine();
+            }
+            try {
+                JSONArray array = new JSONArray(sb.toString());
+                mItems.clear();
+                for (int i = 0; i < array.length(); ++i) {
+                    Item item = new Item(array.getJSONObject(i));
+                    if (item.url != null && item.url.length() > 0) {
```
cvan: another place where it feels like it'd be nice to have a `isStringEmpty` helper method

app/src/common/shared/org/mozilla/vrbrowser/History.java
```diff
@@ -0,0 +1,210 @@
...
+        save();
+    }
+
+    private void save() {
+        JSONArray array = new JSONArray();
+        for (Item item: mItems.values()) {
+            array.put(item.toJSON());
+        }
+
+        try (FileWriter file = new FileWriter(mContext.getFilesDir().getAbsolutePath() + "/" + FILENAME)) {
```
cvan: as this path is reused below in the `restore` method, can we store it in some private class property?

app/src/common/shared/org/mozilla/vrbrowser/History.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    @Override
+    public void onProgressChange(GeckoSession geckoSession, int i) {
+
+    }
+
+    @Override
+    public void onSecurityChange(GeckoSession geckoSession, SecurityInformation securityInformation) {
+
```
cvan: it'd probably be wise to check `isSecure`, in case https://webxr.today gets MITM'd or the cert expires or something - in that case, we should not pass the hash to the page

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -370,6 +380,19 @@ public String getUriFromSession(GeckoSession aSession) {
         return "";
     }
 
+    public String getTitleFromSession(GeckoSession aSession) {
+        Integer sessionId = getSessionId(aSession);
```
cvan: perhaps could these five lines be replaced with a DRY method for `getStateFromSession`? I see this code repeated elsewhere.

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -32,7 +32,7 @@ public static SessionStore get() {
         }
         return mInstance;
     }
-    private static final String HOME_WITHOUT_REGION_ORIGIN = "https://webxr.today";
+    public static final String HOME_WITHOUT_REGION_ORIGIN = "https://webxr.today";
```
cvan: you can keep this `private` if you use the `isHomeUri` helper method, per my comment above

app/src/common/shared/org/mozilla/vrbrowser/History.java
MortimerGoro: Sadly Java doesn't that a function like that. I miss it a lot...

cvan: darn, sad

app/src/common/shared/org/mozilla/vrbrowser/History.java
```diff
@@ -0,0 +1,210 @@
...
+        ArrayList<Item> list = new ArrayList<>(mItems.values());
+        list.sort(new Comparator<Item>() {
+            @Override
+            public int compare(Item o1, Item o2) {
+                return (int)(o1.timestamp - o2.timestamp);
+            }
+        });
+
+        for (Item item: list) {
+            mItems.remove(key(item.url));
```
cvan: not sure we need to strip the hash, but probably helps - there's probably a better way of de-duping sites in Firefox for iOS/Android/desktop. but this seems fine for now.

app/src/common/shared/org/mozilla/vrbrowser/History.java
```diff
@@ -0,0 +1,210 @@
...
+
+    }
+
+    @Override
+    public void onPageStop(GeckoSession geckoSession, boolean b) {
+        String uri = SessionStore.get().getUriFromSession(geckoSession);
+        if (uri == null || uri.length() == 0) {
+            return;
+        }
+        if (SessionStore.get().isHomeUri(uri)) {
```
cvan: thanks :+1:

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -82,12 +83,15 @@ private SessionStore() {
         mPrivateSessionsStack = new ArrayDeque<>();
     }
 
-    public void clearListeners() {
+    public void onDestroy() {
```
cvan: :+1: good change, for consistency

app/src/common/shared/org/mozilla/vrbrowser/History.java
cvan: I handled this in commit 3d7272b where the Home URI gets set in `SessionStore`: https://github.com/MozillaReality/FirefoxReality/pull/457/files#diff-70a8baac173fb5a93995a46725b70c73R440

cvan:
just rebased from merged PR #462 (for updating region in the Home URI)

bluemarvin:
Can this be landed or is there still something still outstanding?

cvan:
@bluemarvin it is finished. just wanted a code review from one of you to make sure I am not doing anything super silly.

app/src/common/shared/org/mozilla/vrbrowser/History.java
```diff
@@ -0,0 +1,212 @@
...
+    private String key(String aUri) {
+        String result = aUri.toLowerCase();
+        int index = result.indexOf('#');
+        if (index >= 0) {
+            result = result.substring(0, index);
+        }
+        return result;
+    }
+
+    private static class SortItemsByTimestamp implements Comparator<Item> {
```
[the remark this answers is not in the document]
cvan: oh, that's a much better idea - thanks :+1:

keianhzo:
@cvan Sorry for the delay code looks good and I want to land it but I can't see any history entries in the home after brwosing for some time and I get this in the console:

`InvalidStateError: An attempt was made to use an object that is not, or is no longer, usable
TypeError: recentEl is null, can't access property "classList" of it[Learn `More]``

Could you look into that before landing?

bluemarvin:
Since we don't have a way to clear or edit history, I would wait to land this until that is possible.

cvan:
> @cvan Sorry for the delay code looks good and I want to land it but I can't see any history entries in the home after browsing for some time and I get this in the console:

I fixed that a few hours ago.

> Since we don't have a way to clear or edit history, I would wait to land this until that is possible.

and yes, I'd agree with Randall here, that we should wait until we have the UI all in FxR before we land this. I can simulate deletion/exclusion from the New-Tab page via JS, but it's probably not a wise choice to expose this functionality without complete user control in the FxR chrome UI.

thanks, everyone. if it's cool, let's leave this open for a bit; I'll rebase it against `master` every so often to not keep it stale. or we can close it. whichever y'all feel is best.

keianhzo:
@cvan If I understand this correctly we are waiting for FxR UI for this to be finished. What kind of UI elements are you referring to? for adding/removing history entries?

bluemarvin:
The GV history API has landed. But I don't know that we want history in the content feed? Is this still the plan or are we going to handle history like bookmarks?

MortimerGoro:
Closing this because we can use the new history android-component to implement this

PR #462, opened 2018-09-03
Update the home page content when there is a region update

### LINKED ISSUE #409: Update Home URI when region (`country_code`) changes from Geolocation API calls (for search engine)
## Hardware
Any

## Steps to Reproduce

1. Launch FxR with PR #402
2. Notice the new New-Tab/Home page correctly loads, but because the Geolocation API call hasn't yet finished for the search-engine code (added in issue #248), we don't include `/?region=<region>` in the Home URI
initially empty (i.e., https://webxr.today)

## Current Behavior
- Upon first load, there is no region in the Home URI: https://webxr.today
- If you click the Home button, you'll be taken to the appropriate Home URI with a region: https://webxr.today/?region=us

## Expected Behavior
- Upon first load, the Home URI contains the user's region: https://webxr.today https://webxr.today/?region=us

## Possible Solution
Whenever the region changes (in the search code), we should use some mechanism, such as `postMessage` to communicate the new region to the Home URI (https://webxr.today), so the page can dynamically update the region from within the content page.

keianhzo:
Fixes #409 Update homepage when there is a region update

We have to deal with this in two different places to make sure the homepage if refreshed in case the homepage has been already been loaded when the region is updated and in case the homepage is not yet loaded so we refresh it after it's loaded.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
@@ -32,7 +32,7 @@ public class SessionStore implements GeckoSession.NavigationDelegate, GeckoSessi
         }
         return mInstance;
     }
-    private static final String HOME_WITHOUT_REGION_ORIGIN = "https://webxr.today";
+    private static final String HOME_WITHOUT_REGION_ORIGIN = "https://webxr.today/";
     public static final int NO_SESSION_ID = -1;
 
     private LinkedList<GeckoSession.NavigationDelegate> mNavigationListeners;
@@ -416,14 +416,18 @@ public class SessionStore implements GeckoSession.NavigationDelegate, GeckoSessi
 
     public void setRegion(String aRegion) {
         Log.d(LOGTAG, "SessionStore setRegion: " + aRegion);
-        aRegion = aRegion != null ? aRegion.toLowerCase() : "worldwide";
-        mRegion = aRegion;
+        mRegion = aRegion != null ? aRegion.toLowerCase() : "worldwide";
+
+        // There is a region update and the home is already loaded
+        if (mCurrentSession != null && isHomeUri(getCurrentUri())) {
+            mCurrentSession.loadUri(SessionStore.HOME_WITHOUT_REGION_ORIGIN + "#region=" + mRegion);
+        }
     }
 
     public String getHomeUri() {
         String result = SessionStore.HOME_WITHOUT_REGION_ORIGIN;
         if (mRegion != null) {
-            result = SessionStore.HOME_WITHOUT_REGION_ORIGIN + "/?region=" + mRegion;
+            result = SessionStore.HOME_WITHOUT_REGION_ORIGIN + "?region=" + mRegion;
         }
         return result;
     }
@@ -509,6 +513,7 @@ public class SessionStore implements GeckoSession.NavigationDelegate, GeckoSessi
         if (mCurrentSession == null) {
             return;
         }
+
         if (aUri == null) {
             aUri = getHomeUri();
         }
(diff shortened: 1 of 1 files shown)


app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -416,14 +416,18 @@ public void setCurrentSession(int aId) {
 
     public void setRegion(String aRegion) {
         Log.d(LOGTAG, "SessionStore setRegion: " + aRegion);
-        aRegion = aRegion != null ? aRegion.toLowerCase() : "worldwide";
-        mRegion = aRegion;
+        mRegion = aRegion != null ? aRegion.toLowerCase() : "worldwide";
+
+        // There is a region update and the home is already loaded
+        if (mCurrentSession != null && isHomeUri(getCurrentUri())) {
+            mCurrentSession.loadUri(SessionStore.HOME_WITHOUT_REGION_ORIGIN + "#{'region':'" + mRegion + "'}");
```
cvan: feel free to change this to `#region=mRegion`; I have code in place for parsing `#history=` already, so it'll be easy to accept

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -719,6 +724,11 @@ public void onLocationChange(GeckoSession aSession, String aUri) {
         for (GeckoSession.NavigationDelegate listener: mNavigationListeners) {
             listener.onLocationChange(aSession, aUri);
         }
+
+        // The homepage finishes loading after the region has been updated
+        if (mRegion != null && aUri.equalsIgnoreCase(SessionStore.HOME_WITHOUT_REGION_ORIGIN)) {
```
cvan: you can replace the latter part with `isHomeUri(aUri)`

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -719,6 +724,11 @@ public void onLocationChange(GeckoSession aSession, String aUri) {
...
+            aSession.loadUri(SessionStore.HOME_WITHOUT_REGION_ORIGIN + "#{'region':'" + mRegion + "'}");
```
cvan: same here

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
keianhzo: I use equalsIgnoreCase instead of the startsWith from isHomeUri to avoid infinite loading when the #region= is loaded as it also starts with the HOME_WITHOUT_REGION_ORIGIN string.

cvan: ah, good call

cvan: could you change `isHomeUri` to something like this:

```diff
diff --git a/app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java b/app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
index 6d6ea25..9ff011a 100644
--- a/app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
@@ -432,8 +432,14 @@ public class SessionStore implements GeckoSession.NavigationDelegate, GeckoSessi
         return result;
     }
 
-    public Boolean isHomeUri(String aUri) {
-        return aUri != null && aUri.toLowerCase().startsWith(SessionStore.HOME_WITHOUT_REGION_ORIGIN);
+    public Boolean isHomeUri(String aUri, boolean aExactMatch) {
+        if (aUri == null || aUri.length == 0) {
+            return false;
+        }
+        if (aExactMatch === true) {
+            return aUri.equalsIgnoreCase(SessionStore.HOME_WITHOUT_REGION_ORIGIN);
+        }
+        return aUri.toLowerCase().startsWith(SessionStore.HOME_WITHOUT_REGION_ORIGIN);
     }
 
     public String getCurrentUri() {
```

bluemarvin: I'm generally against booleans in APIs. When reading the code, it is really hard to understand what the boolean in `isHomeUri("some.uri", false)` means. I would prefer two functions `isHomeUri()` and `startsWithHomeUri()`.

cvan: yeah, that's better - thanks 👍will change later

PR #468, opened 2018-09-04
Switch from local test signing to Mozilla signing

### LINKED ISSUE #408: Officially sign Firefox Reality builds with a Mozilla Android key
## Hardware
All

## Current Behavior
Release APKs are signed with a test key.

## Expected Behavior
Sign with an official Mozilla Android key. This is required to ship in stores.

## Possible Solution
1. Request an Autograph services edge key
2. Add signing steps to the taskcluster builds referencing that secret edge key

larsbergstrom:
Fixes #408

Before landing this, I'll need to get the official production `AUTOGRAPH_EDGE_TOKEN` into the taskcluster secrets. This works locally and produces a build that passes signing and installs & runs cleanly on the Go.

I'd still love any suggestions for fixups!

This uses the default recommended CI method ("just use `curl`") the webops team recommended to me, linked from:
https://mozilla.github.io/application-services/docs/applications/signing-android-apps.html

---
This change is [Reviewable](https://reviewable.io/reviews/mozillareality/firefoxreality/468)

### DIFF
--- a/tools/taskcluster/sign.sh
+++ b/tools/taskcluster/sign.sh
@@ -1,6 +1,4 @@
 #!/bin/sh
-python tools/taskcluster/fetch_secret.py -s project/firefoxreality/preview-keystore -d -o keystore.jks -n store
-python tools/taskcluster/fetch_secret.py -s project/firefoxreality/keystore-password -o keystore_password -n password
-python tools/taskcluster/fetch_secret.py -s project/firefoxreality/key-password -o key_password -n password
-python tools/taskcluster/sign_apk.py -s keystore.jks -p keystore_password -k key_password -a FirefoxRealityTestKey
+python tools/taskcluster/fetch_secret.py -s project/firefoxreality/autograph_token -o token -n password
+python tools/taskcluster/sign_apk.py -t token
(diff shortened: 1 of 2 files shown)


bluemarvin:
I'm not sure we want to get rid of the unofficial-key signing. For Fx Focus, all builds from pushes to master are signed with a unofficial key and only release candidate builds are signed with the official mozilla key.

larsbergstrom:
We don't really have a release candidate pipeline/tooling. Or are you proposing that I just locally request/push signing when we have a build that we want officially signed? I don't really care and it's easier in some ways but it makes for some horrific busfactor (I'm the only one with a signing key, e.g. if we need to address a CVE).

bluemarvin:
I'm saying we add official signing. It will still be a taskcluster event. I think the Fx Focus release build task is commented out in our taskcluster file. I'm saying we do what they do. I'm just not sure what they do to trigger a release build. I can ask.

bluemarvin:
Okay, I understand how to trigger a release build. I can try modifying the taskcluster file. I'm not clear what else need to be added to taskcluster so we can get the apk signed with the official key.

PR #482, opened 2018-09-06
Update onLoadError to return page inplace of load it

bluemarvin:


keianhzo:
Doesn't seem to be working. Maybe only URIs can be returned?


<details open>


```
09-07 11:28:30.525 8310-8330/org.mozilla.vrbrowser W/GeckoLoadURIDelegate: Failed to parse URI '<?xml version="1.0" encoding="UTF-8"?>
    <!DOCTYPE html>
    <!-- This Source Code Form is subject to the terms of the Mozilla Public
       - License, v. 2.0. If a copy of the MPL was not distributed with this
       - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
    
    <html xmlns="http://www.w3.org/1999/xhtml">
```


</details>

bluemarvin:
Let me talk to @snorp as he is the one who suggested the changes.

snorp:
Yeah, you have to return a URI. You can use a data URI in this case.

`"data:text/html," + theHtml` should work

bluemarvin:
I've encountered issues with the safe browsing pages getting stuck in an infinite loop. Still investigating.

keianhzo:
@bluemarvin I've updated the branch. Loading as base64 data URI seems to work. Not sure what you refer with "safe browsing pages getting stuck" but let me know if my latest commit fixes it.

bluemarvin:
What I was seeing was us getting into an infinite loop where onLoadError() was called over and over again. Your patch seems to work.

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -727,7 +714,7 @@ public void onLocationChange(GeckoSession aSession, String aUri) {
 
         // The homepage finishes loading after the region has been updated
         if (mRegion != null && aUri.equalsIgnoreCase(SessionStore.HOME_WITHOUT_REGION_ORIGIN)) {
-            aSession.loadUri(SessionStore.HOME_WITHOUT_REGION_ORIGIN + "#region=" + mRegion);
+            aSession.loadUri("javascript:window.location.replace('" + getHomeUri() + "');");
```
cvan: curious about the thinking behind this.

this might cause unintended/confusing refreshes/redirects, but it may probably best we handle this on the server, so as to not confusing URLs where the `?region=<mRegion>` does not match the `#region=<mRegion>`.

I was thinking we could do this to avoid the refresh/redirect:

```java
aSession.loadUri("javascript:window.location.hash = (window.location.hash ? window.location.hash + '&region=' : '#') + " + mRegion + "');
```

thoughts?

app/src/common/shared/org/mozilla/vrbrowser/ui/NavigationURLBar.java
```diff
@@ -111,6 +111,8 @@ public void setURL(String aURL) {
                 return;
             else if (aURL.startsWith("resource:") || SessionStore.get().isHomeUri(aURL))
                 aURL = "";
+            else if (aURL.startsWith("data:") && SessionStore.get().isCurrentSessionPrivate())
```
cvan: what if I want to load my own, a different, data URI in Private Browsing mode? I do this often on desktop (e.g., `data:text/html,&times;`). admittedly, I wouldn't expect this to be done by a non-developer, but this will certainly cause a minor bug.

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -90,15 +90,6 @@ public void clearListeners() {
         mTextInputListeners.clear();
     }
 
-    public static final String LOCAL_URI = "data:";
-
-    public static boolean isLocalPage(String aURI) {
```
cvan: 👍good call on removing. we don't want to make that assumption

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
bluemarvin: loadUri doesn't support `javascript:` URLs, that's why it was changed.

bluemarvin: I guess it isn't supposed to work with remote pages but if it is working I guess it's okay.

app/src/common/shared/org/mozilla/vrbrowser/ui/NavigationURLBar.java
keianhzo: Good call, we could add some logic to more properly handle the private mode page data URL masking without affecting other data URLs.

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
keianhzo: @cvan Not sure if I get it. The problem with URL loading (being ? or #) is that although there is no server request they get commited to the session history. What's the drawback of the window.location.replace? I thought that might help refreshing locally without making server requests or adding history entries.

Using:
```
aSession.loadUri("javascript:window.location.hash = (window.location.hash ? window.location.hash + '&region=' : '#') + " + mRegion + "');
```

will add a history entry.

PR #489, opened 2018-09-07
FxR Logo is no longer head locked on start up. Fixes #487

### LINKED ISSUE #487: Loading icon should not be headlocked
## Hardware
Oculus Go

## Steps to Reproduce
1. Launch Firefox Reality from the library

## Current Behavior
The loading icon is headlocked (moves with your head).

## Expected Behavior
The loading icon is in a fixed position.

## Details
This is a ship blocker / will result in app rejection if we don't fix it.

bluemarvin:


bluemarvin:
I'm just going to land this since it is a small change and blocking the oculus store review.

PR #507, opened 2018-09-11
Support for select prompt

### LINKED ISSUE #234: Implement UI for <select> fields
Hardware: All
Steps to Reproduce:
1) Go to: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select
2) Scroll to Examples.
3) Click on selection drop down list.

Actual Behavior:
Nothing happens.

Expected Behavior:
A selection list is presented and the user may select an item in the list.

Recommendation:
FxR need to implement UI for the various selection types a web page may present.

keianhzo:
Fixes #234 Added support for the select prompt.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
@@ -23,7 +23,10 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.util.*;
 
-public class SessionStore implements GeckoSession.NavigationDelegate, GeckoSession.ProgressDelegate, GeckoSession.ContentDelegate, GeckoSession.TextInputDelegate, GeckoSession.TrackingProtectionDelegate {
+public class SessionStore implements GeckoSession.NavigationDelegate, GeckoSession.ProgressDelegate,
+        GeckoSession.ContentDelegate, GeckoSession.TextInputDelegate, GeckoSession.TrackingProtectionDelegate,
+        GeckoSession.PromptDelegate {
+
     private static SessionStore mInstance;
     private static final String LOGTAG = "VRB";
     public static SessionStore get() {
@@ -41,6 +44,7 @@ public class SessionStore implements GeckoSession.NavigationDelegate, GeckoSessi
     private LinkedList<GeckoSession.ContentDelegate> mContentListeners;
     private LinkedList<SessionChangeListener> mSessionChangeListeners;
     private LinkedList<GeckoSession.TextInputDelegate> mTextInputListeners;
+    private LinkedList<GeckoSession.PromptDelegate> mPromptListeners;
 
     public interface SessionChangeListener {
         void onNewSession(GeckoSession aSession, int aId);
@@ -76,6 +80,7 @@ public class SessionStore implements GeckoSession.NavigationDelegate, GeckoSessi
         mContentListeners = new LinkedList<>();
         mSessionChangeListeners = new LinkedList<>();
         mTextInputListeners = new LinkedList<>();
+        mPromptListeners = new LinkedList<>();
 
         mSessions = new LinkedHashMap<>();
         mSessionsStack = new ArrayDeque<>();
@@ -256,6 +261,14 @@ public class SessionStore implements GeckoSession.NavigationDelegate, GeckoSessi
         mTextInputListeners.remove(aListener);
     }
 
+    public void addPromptListener(GeckoSession.PromptDelegate aListener) {
+        mPromptListeners.add(aListener);
+    }
+
+    public void removePromptListener(GeckoSession.PromptDelegate aListener) {
+        mPromptListeners.remove(aListener);
+    }
+
     public class SessionSettings {
         public boolean multiprocess = SettingsStore.getInstance(mContext).isMultiprocessEnabled();
         public boolean privateMode = false;
@@ -280,6 +293,7 @@ public class SessionStore implements GeckoSession.NavigationDelegate, GeckoSessi
         state.mSession.getSettings().setInt(GeckoSessionSettings.USER_AGENT_MODE, aSettings.userAgentMode);
         state.mSession.setNavigationDelegate(this);
         state.mSession.setProgressDelegate(this);
+        state.mSession.setPromptDelegate(this);
         state.mSession.setContentDelegate(this);
         state.mSession.getTextInput().setDelegate(this);
         state.mSession.setPermissionDelegate(mPermissionDelegate);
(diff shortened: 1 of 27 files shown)


bluemarvin:
@thenadj The radio buttons do not make any sense. Can they please be removed?

keianhzo:
@bluemarvin the only place were they make sense for me is when it's a multiple select, we could remove them from the single options select and leave it for the multiple.

bluemarvin:
@keianhzo I actually disagree. To me, radio buttons represent an exclusive selection (i.e. selecting one button unselects the currently selected button). Oculus browser uses check marks but I don't know that we even need that. For multiple selection, typically you just leave each item highlighted as its selected.

bluemarvin:
BTW, how do you even do multiple selection? I couldn't figure it out or maybe the sample I tried didn't support it?

keianhzo:
@bluemarvin actually you are right, I'll go with the checks or nothing, it's what others are using anyway, we shouldn't reinvent the wheel here.

Multiple select: https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select_multiple

bluemarvin:
@keianhzo Thanks, I think if we can replace the radio buttons with something that makes more senses, this looks good to me.

thenadj:
@keianhzo @bluemarvin Let's remove the radio buttons. We based it on this (see attached)from Samsung Internet not completely understanding where this control would be used and how. 

dropdown

keianhzo:
@thenadj @bluemarvin it seems that Samsung Internet, Fennec and Chrome are using radio buttons. Fennec and Chrome also use checks for multiple. Most of implementations I've seen didn't have checks or radio buttons. Let's see what @cvan's has to say before making any further changes.

bluemarvin:
I would be okay using it for single selection if it was used to show the current selection. Right now, I don't think any are selected and the window is closed before the button changes state. It is just a column of empty circles.

cvan:
> I would be okay using it for single selection if it was used to show the current selection. Right now, I don't think any are selected and the window is closed before the button changes state. It is just a column of empty circles.

yeah, I agree with @thenadj here that we should remove the radio buttons.

thenadj:
The state is supposed to change to a filled in circle on select. See UIS-45 for reference. Also, we're using a dialog rather than a dropdown, so the functionality is slightly different than in a traditional dropdown. 

[UIS-45 HTML Fields & Prompts_001.pdf](https://github.com/MozillaReality/FirefoxReality/files/2372727/UIS-45.HTML.Fields.Prompts_001.pdf)

thenadj:
Your call, I could go either way.

cvan:
pasting some screenshots and proposal based on your UIS-45. will check in a few min.

cvan:
done. check these [`<select>` states](https://github.com/MozillaReality/FirefoxReality/files/2372942/fxr-select.pdf), and see if they align with what everyone expects from a `<select>` field in FxR.

([View/Comment on Google Drive](https://drive.google.com/file/d/13t7TTXIrNGBDeWrm5vTxY2CzTaZtApuy/view?usp=sharing))

thenadj:
@cvan Are you suggesting to use checkmarks instead of radio buttons for selected items? Please follow the visual style from UIS-45. The visual design in the PDF you attached is not consistent with the rest of the UI (i.e. we don't have blue hover and the font looks different). I assume that's just a wireframe.

cvan:
> @cvan Are you suggesting to use checkmarks instead of radio buttons for selected items?

Yep! We agree that radio buttons don't make sense for `<select>` fields. `OK` and `Cancel` also do not make sense that's not how `<select>` fields work in browsers; it adds state complexity and unexpected behaviour for users (e.g., was my selection saved or not, do I have to also click this button to dismiss it, and so on).

I'm suggesting we use the same checkmark and blue hover as used in the `<select>` fields ([example](https://mixedreality.mozilla.org/fxr-html-tests/select.html)) à la all the desktop browsers.

image
image
image

> Please follow the visual style from UIS-45. The visual design in the PDF you attached is not consistent with the rest of the UI (i.e. we don't have blue hover and the font looks different). I assume that's just a wireframe.

Heh, not suggesting we alter the font. FWIW, I had UIS-45 open while I was mocking this up. And I referred to the [UIS-7 Color Palette](https://trello.com/c/TDEHC92c/55-uis-7-color-palette) and [UIS-30 Firefox Reality Logo - Colors](https://trello.com/c/7BJVI0RJ/115-uis-30-firefox-reality-logo-colors).

Selection as Blue is a browser convention. I'm not sure why FxR should diverge here. When you highlight text on a web page or in the URL bar in FxR, that's blue too.

In terms of fixing the design for implementation, if my mocks do not address the needs, please update UIS-45 for @keianhzo.

thenadj:
@cvan @keianhzo We're not using blue as a hover or selection color in FxR. It might be standard for desktop but our VR interaction palette is different. As for Cancel/OK, we need those in there. The dropdown is actually a dialog floating in the middle of the screen and there needs to be a way to opt out if a user doesn't want to make a selection at all. Attached is the updated UIS-45 (...002) with updated colors and checkmark. 

[UIS-45 HTML Fields & Prompts_002.pdf](https://github.com/MozillaReality/FirefoxReality/files/2373279/UIS-45.HTML.Fields.Prompts_002.pdf)

keianhzo:
@thenadj @bluemarvin @cvan I've updated the PR:
- Replaced Radios for Checks
- Single choice mode: Checks are removed but I kept the "Cancel" button only to send back the default choice if the user doesn't choose any.
- Multiple mode: Checks are shown and both the "Ok" and "Cancel" buttons are shown. Ok, send the current selection, cancel send the default selection.

bluemarvin:
@keianhzo For single select, I think we need to include the check mark so the user knows what is currently selected or what is the default selection.

keianhzo:
Good catch, fixed.

bluemarvin:
@keianhzo For me, the very first time the selection box is clicked, there is no check but is there is on the second time.

cvan:
that was quick. thanks!

a few issues I noticed:

image

when the `<select>` is opened, and I click on another form field on the page, the `<select>` doesn't get dismissed.

cvan:
upon choosing a selection in the `<select>`, it'd be nice if the checkmark changed before getting dismissed. currently, if you hold and click you can see it change, but if you click and release quickly, you can't.

fxr-select-release

cvan:
it's a nit, but it confused me for a bit: an `<option>` may still appear selected even after you uncheck all the options. I believe this is caused by when the `<select>` is opened, we check the first in the `<select multiple>` modal, which probably is causing the `<option>` to be focused on the page.

perhaps we should blur the `<option>` when it's unselected and the user has pressed `OK` and the `<select multiple>` modal has been dismissed.

fxr-select-multiple

keianhzo:
@bluemarvin @cvan thanks for the review. I've addressed all of the above. 

@cvan When you click on another select, the current dialog is refreshed with the options of the new clicked select. We could also just make the current dialog to dismiss but think I prefer the current behavior as if you click on another dialog you'll probably want to display that one.

bluemarvin:
But that did make it easy to test it 🙂

keianhzo:
Sorry for that (⊙_◎)

bluemarvin:
Let's land this and handle any outstanding issues in a follow up.

Issue #358, opened 2018-08-13
On New Tab page/Home, show placeholder text "Search with Google or enter address" when the URL bar is empty


cvan:
## Steps to Reproduce

1. Launch Firefox Reality on a fresh install (or press the Home button)
2. Notice the URL bar is empty, with no placeholder text (i.e., `input[placeholder="…"]`)

## Current Behavior
The URL bar is empty on the New Tab/Home page. As it's a rectangle below the browser, it's unclear that the URL bar is in fact a URL bar (and also duals as a search bar for Google).

![screenshot of New Tab/Home page in Firefox Reality](https://user-images.githubusercontent.com/203725/44060524-cd605260-9f09-11e8-9707-b62b1d948d4e.png)

## Expected Behavior
When the user first launches Firefox Reality or presses the Home button, the user should know that rectangle below the browser window (i.e., the URL bar) is a URL bar and can also accept search.

![screenshot of URL bar on Firefox desktop](https://user-images.githubusercontent.com/203725/44060328-30a8993c-9f09-11e8-885a-62d5f1aeb23e.png)

## Possible Solution

This is what the Search form looks like on Firefox desktop's New Tab page:

![image](https://user-images.githubusercontent.com/203725/44060617-155d7b10-9f0a-11e8-95d7-c940c3b62459.png)

Based on that, I mocked this up:

### default without any user action
![image](https://user-images.githubusercontent.com/203725/44060582-fb241dc6-9f09-11e8-9ad5-1d7f31e00763.png)

### user has focused field
![image](https://user-images.githubusercontent.com/203725/44060642-2aa1811a-9f0a-11e8-8e59-c931b7c06ce6.png)

### user has typed text into the field
![image](https://user-images.githubusercontent.com/203725/44060680-47af56b0-9f0a-11e8-9885-673d63488aac.png)

PR #513, opened 2018-09-12
Always position URL bar caret at the beginning when focusing

### LINKED ISSUE #476: Position of caret in text of URL should reset to beginning after focus + blur
## Hardware
- All likely, but tested on Oculus Go
- FxR on `master` (current: 3f8a2ed)

## Steps to Reproduce
1. Load FxR to a page with a long URL (I used the Voice Search in the URL bar to do a Google Search for "360 Videos")
2. Click on the URL bar
3. Notice the URL bar is highlighted
4. Notice the end of the URL is visible but not the beginning of the URL
5. Exit the URL bar (e.g., by clicking on the page or exiting the keyboard)
6. Notice the end of the URL is still visible, so you can't see the protocol+host+beginning of the URL

## Current Behavior
- Upon focusing the URL bar, the text is focused but the caret is set to the *end* of the URL bar.
- Upon exiting the URL bar, the caret is still set to the *end* of the URL bar.

fxr_url_focus

## Expected Behavior
- Upon focusing the URL bar, the text is focused but the caret is set to the **beginning** of the URL bar upon additional click.
- Upon exiting the URL bar, the caret is reset to the **beginning** of the URL bar.

## Possible Solution
See above.

## Context
This causes confusion in the user experience and possibly security implications since the protocol is not shown. But not a total dealbreaker. I expect most folks to not encounter this, but it should be fixed.

keianhzo:
Fixes #476 Always position URL bar caret at the beginning when focusing

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/NavigationURLBar.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/NavigationURLBar.java
@@ -13,7 +13,9 @@ import android.text.TextWatcher;
 import android.text.style.ForegroundColorSpan;
 import android.util.AttributeSet;
 import android.util.TypedValue;
+import android.view.GestureDetector;
 import android.view.KeyEvent;
+import android.view.MotionEvent;
 import android.view.View;
 import android.view.animation.Animation;
 import android.view.animation.AnimationUtils;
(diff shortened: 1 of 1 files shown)


cvan:
P.S. I spent a few minutes to see if I could easily add placeholder text when the URL bar is empty and not focused and on the Home page/`about:` pages. Issue #358. Here's my WIP branch: https://github.com/MozillaReality/FirefoxReality/compare/search-placeholder?expand=1

It's close, but I'm getting a bit confused about how to store the real URL vs. the placeholder, and whether this logic should live in `NavigationBarWidget` instead of in `NavigationURLBar`. If there are more pressing items, we can wait. When you launch FxR today, the URL bar is empty and it looks awkward and confusing, since there's no search icon or anything else in the URL bar. Let me know if either of you have ideas for a quick fix.

keianhzo:
@cvan I've added a fix for that in #522 I've implemented the placeholder text using a hint that automatically handles the placeholder text when the text is empty.

PR #521, opened 2018-09-13
Fixed showing microphone icon when there is text in the URL

### LINKED ISSUE #514: Microphone icon shows when clear icon should show in URL bar
## Steps to Reproduce

1. Enter URL into URL bar
2. Clear it by pressing the `x` button in the URL bar
3. The `x` is replaced with the microphone button.
4) Press the microphone button and do a voice search.

## Current Behavior

Even thought there is now a URL in the URL bar and it has focus, the microphone button is still visible.

## Expected Behavior

The clearing `x` button should be visible.

## Possible Solution

Make sure the button state is switched when there is text in URL bar and it has focus.

keianhzo:
Fixes #514 Fixed showing microphone icon when there is text in the URL after a voice search

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/NavigationURLBar.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/NavigationURLBar.java
@@ -73,8 +73,8 @@ public class NavigationURLBar extends FrameLayout {
         mURL.setOnFocusChangeListener(new OnFocusChangeListener() {
             @Override
             public void onFocusChange(View view, boolean b) {
-                if (b && mURL.getText().length() > 0) {
-                    showVoiceSearch(false);
+                if (b) {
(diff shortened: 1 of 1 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/NavigationURLBar.java
```diff
@@ -73,8 +73,8 @@ public boolean onEditorAction(TextView aTextView, int actionId, KeyEvent event)
         mURL.setOnFocusChangeListener(new OnFocusChangeListener() {
             @Override
             public void onFocusChange(View view, boolean b) {
-                if (b && mURL.getText().length() > 0) {
-                    showVoiceSearch(false);
+                if (b) {
```
bluemarvin: In a follow up, we should rename that `b` variable to something else.

PR #540, opened 2018-09-18
Add telemetry for uri count.

daoshengmu:
Update `mozilla.components:telemetry` to 0.23 for using large size of histograms.

Follow the requirement for the product side to understand URI count as Focus team (https://github.com/mozilla-mobile/focus-android/issues/3320#issuecomment-419091837) and add the histogram to know the avg. time spent for the page loading.

@pocmo Please help review the telemetry part. I believe we need a data review as well.


app/src/common/shared/org/mozilla/vrbrowser/utils/UrlUtils.java
```diff
@@ -0,0 +1,160 @@
...
+                    url1.regionMatches(true, 0, url2, 0, url2.length());
+        } else if (lengthDifference == -1) {
+            return url2.charAt(url2.length() - 1) == '/' &&
+                    url2.regionMatches(true, 0, url1, 0, url1.length());
+        }
+
+        return false;
+    }
+
+    public static String stripCommonSubdomains(@Nullable String host) {
```
bluemarvin: This function appears to be the only one used in this file. Please don't just copy and paste whole files into the code base just to use one function.

app/src/common/shared/org/mozilla/vrbrowser/utils/UrlUtils.java
```diff
@@ -0,0 +1,160 @@
...
+        } else if (host.startsWith("mobile.")) {
+            start = 7;
+        } else if (host.startsWith("m.")) {
+            start = 2;
+        }
+
+        return host.substring(start);
+    }
+
+    public static boolean isLocalizedContent(@Nullable String url) {
```
bluemarvin: I don't think this function actually does anything right now. If it is supposed to filter out local content like error pages and the home page, it needs to do a lot more than just watch for `about:blank`

app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
```diff
@@ -129,5 +174,24 @@ private static void browseEvent() {
         // TODO: Working on autocomplete result.
         event.queue();
     }
+
+    @UiThread
+    public static void addLoadToHistogram(String uri, Long newLoadTime) {
+        domainMap.add(UrlUtils.stripCommonSubdomains(URI.create(uri).getHost()));
```
bluemarvin: This seems to be striping just the common leading parts of domains but leaving all the user data and page info? If it is a domain map shouldn't it end at the first `/`?

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -837,6 +843,14 @@ public void onPageStop(GeckoSession aSession, boolean b) {
         }
 
         state.mIsLoading = false;
+        long elapsedLoad = SystemClock.elapsedRealtime() - startLoadTime;
+        if (elapsedLoad > MIN_LOAD_TIME
+                && !state.mUri.equals(getHomeUri())
+                && !state.mUri.equals(HOME_WITHOUT_REGION_ORIGIN)
```
cvan: we don't want to track the Home page? is there a reason why?

app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
```diff
@@ -19,31 +21,51 @@
...
+import java.util.HashSet;
+import static java.lang.Math.toIntExact;
 
 
 public class TelemetryWrapper {
     private final static String APP_NAME = "FirefoxReality";
+    private final static String LOGTAG = "VRB";
+    private final static int HISTOGRAM_SIZE = 200;
+    private final static int BUCKET_SIZE_MS = 100;
+    private final static int HISTOGRAM_MIN_INDEX = 0;
```
cvan: I move this line above for consistency/alphabetical order

app/src/common/shared/org/mozilla/vrbrowser/utils/UrlUtils.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * Is the given string a URL or should we perform a search?
+     *
+     * TODO: This is a super simple and probably stupid implementation.
+     */
+    public static boolean isUrl(String url) {
+        String trimmedUrl = url.trim();
+        if (trimmedUrl.contains(" ")) {
```
cvan: this most certainly could be triggered as a false positive by a space in unencoded hash or query-string values

app/src/common/shared/org/mozilla/vrbrowser/utils/UrlUtils.java
```diff
@@ -0,0 +1,160 @@
...
+                        scheme.startsWith("javascript") ||
+                        scheme.startsWith("about"));
+    }
+
+    public static boolean isSupportedProtocol(@Nullable final String scheme) {
+        return scheme != null && (isPermittedResourceProtocol(scheme) || scheme.startsWith("error"));
+    }
+
+    public static boolean isInternalErrorURL(final String url) {
+        return "data:text/html;charset=utf-8;base64,".equals(url);
```
cvan: we need to figure out a better way of identifying an error URL. there can be data URIs that look like this that aren't error URIs.

app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
daoshengmu: I am not sure I understand what you mean. For the example of `https://m.facebook.com/?refsrc=https%3A%2F%2Fwww.facebook.com%2F&_rdr`, `getHost()` will return `m.facebook.com` then `UrlUtils.stripCommonSubdomains ` helps us strip it to `facebook.com`.

app/src/common/shared/org/mozilla/vrbrowser/utils/UrlUtils.java
daoshengmu: Making `isLocalizedContent ()` to check if it is `PRIVATE_BROWSING_URI, HOME_WITHOUT_REGION_ORIGIN, getHomeUri, and about:blank`

app/src/common/shared/org/mozilla/vrbrowser/utils/UrlUtils.java
daoshengmu: I have removed unused functions.

daoshengmu:
yes. I have tested. For example `http://aframe.io/examples/showcase/videosphere/`, it only puts `aframe.io` to the hash table. Besides, we will not upload user data.  `//We only upload the domain and URI counts to the probes without including users' URI info.`

bluemarvin:
> yes. I have tested. For example `http://aframe.io/examples/showcase/videosphere/`, it only puts `aframe.io` to the hash table. Besides, we will not upload user data. `//We only upload the domain and URI counts to the probes without including users' URI info.`

Sorry, I miss read the code.

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
```diff
@@ -156,6 +161,13 @@ public void dumpAllState(Integer sessionId) {
         dumpAllState(getSession(sessionId));
     }
 
+    private boolean isLocalizedContent(@Nullable String url) {
+        return url != null && (url.equals("about:blank")
+                || url.equals(PRIVATE_BROWSING_URI)
+                || url.equals(HOME_WITHOUT_REGION_ORIGIN)
+                || url.equals(getHomeUri()));
+    }
```
bluemarvin: will `url.equals(HOME_WITHOUT_REGION_ORIGIN)` ever catch a case `url.equals(getHomeUri()))` does not?

Additionally, I agree with @cvan, do we really want to filter the home page since it is a web page like any other content? I don't know.

But maybe instead of filtering `about:blank`, maybe we should be filtering `url.startsWith("about:");` ?

app/src/common/shared/org/mozilla/vrbrowser/utils/UrlUtils.java
daoshengmu: I have removed it at my new update. Thanks!

app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
daoshengmu: done

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
daoshengmu: agree!

app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
daoshengmu: I have added `url.startsWith("about:") || url.startsWith("data:")` for filtering. Also adding `uri.getHost()` to check other unwanted uri.

PR #550, opened 2018-09-19
Correct ExternaVR shutdown

### LINKED ISSUE #538: Correct ExternaVR shutdown
The ExternaVR struct never gets released from the Gecko side. But when the activity is destroyed when FxR gets back grounded, the structure gets deleted. Then when a new activity starts, gecko is using the old deleted external structure and it crashes, VR doesn't work, or other undefined behavior.

We are using this hack as a workaround now: https://github.com/MozillaReality/FirefoxReality/commit/26cf8416c6a548029ff59dbc5dacde65c4c90cfa

We should implement a correct ExternalVR shutdown from both FXR and Gecko sides

MortimerGoro:
Fixes #538

Related Gecko PR: https://phabricator.services.mozilla.com/D6236

### DIFF
--- a/app/src/main/cpp/ExternalVR.cpp
+++ b/app/src/main/cpp/ExternalVR.cpp
@@ -6,7 +6,6 @@
 #include "ExternalVR.h"
 #include "VRBrowser.h"
 
-#include "vrb/ConcreteClass.h"
 #include "vrb/Matrix.h"
 #include "vrb/Quaternion.h"
 #include "vrb/Vector.h"
@@ -121,6 +120,7 @@ private:
 namespace crow {
 
 struct ExternalVR::State {
+  static ExternalVR::State * sState;
   mozilla::gfx::VRExternalShmem data;
   mozilla::gfx::VRSystemState system;
   mozilla::gfx::VRBrowserState browser;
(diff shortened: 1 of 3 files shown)


app/src/main/cpp/ExternalVR.cpp
```diff
@@ -420,6 +420,26 @@ ExternalVR::StopPresenting() {
...
+ExternalVR::Shutdown() {
+  m.system.displayState.shutdown = true;
+  m.system.displayState.mSuppressFrames = true;
+  PushSystemState();
+
+  // Wait until Gecko has processed the shutdown message
+  // Otherwise the shmem data may be deallocated before Gecko processes it, leading to undefined behaviour
+  Wait wait(m.data.browserMutex, m.data.browserCond);
+  wait.Lock();
+  m.PullBrowserStateWhileLocked();
```
bluemarvin: This got me thinking about how to deal with the PullBrowserStateWhileLocked() issue. I wonder if we could use a closure to fix this issue where the code would look something like:
```
void PullBrowserState(std::function<bool(const BrowserState&)>& aFunc, float aWait = 0.0f) {
  wait.lock();
  copy_data();
  while(!aFunc(m.browserState)) {
    if (!wait.DoWait(aWait)) {
     break;
    }
    copy_data()
  }
}

**Comment by bluemarvin** [comment]
I applied the GV patch and pulled this and built and I don't think it is working. STR on the Go:
1) Go to https://tonite.dance
2) Enter immersive mode.
3) Press the home button.
4) Exit FxR
5) Relaunch FxR

Expected: FxR launches with https://tonite.dance as the currently loaded page not in immersive mode. But the user is able to enter immersive mode.
Actual: Black screen and the Go is hung requiring a system reboot, or FxR crashes, or FxR fails to enter immersive mode.

**Comment by bluemarvin** [comment]
I'm seeing this in the log:
`09-19 14:34:35.860  2437  3092 W VRB     : FXR Shutdown wait timeout`

**Comment by MortimerGoro** [comment]
I changed the approach to use the static struct we talked about in slack.

Even with that sometimes the next relaunch was a black screen. I found that sometimes there is a deadlock in `onDestroy` when the browser widget is released (on the `display.SurfaceDestroyed()` call, which stops the compositor using a sync() call). This only happens when the activity is destroyed while presenting immersive. 

It seems that there is a race condition where the immersive mode is not paused before `onDestroy` arrived, probably because the frame loop is stopped before the event sent from `onPause` is processed. I made the `exitImmersive() `call sync in `onPause` to fix the issue. This

**Comment by bluemarvin** [comment]
@MortimerGoro I assume this no longer requires the Gecko patch?

**Comment by MortimerGoro** [comment]
@bluemarvin yes, the GV patch is no longer required. I just set the shutdown flag for correctness.

File: app/src/main/cpp/ExternalVR.cpp [hunk 2]
```diff
@@ -179,6 +190,8 @@ struct ExternalVR::State {
   }
 };
 
+ExternalVR::State * ExternalVR::State::sState = nullptr;
+
 ExternalVRPtr
 ExternalVR::Create() {
   return std::make_shared<vrb::ConcreteClass<ExternalVR, ExternalVR::State> >();
```
bluemarvin: Since the state is no longer instanced with the class this should be:
```
return std::make_shared<ExternalVR>();
```

And `#include "vrb/ConcreteClass.h"` can be removed.

bluemarvin:
I'm seeing a dead lock when restarting:
STR:
1) Go to https://tonite.dance
2) Enter immersive mode
3) Press home button on controller
4) Press exit button
5) Launch FxR again
6) Goto 2)

Expected: Should be able to do this indefinitely
Actual: After going in and out several times, FxR will hang on launch and just show a black screen after step 5)

bluemarvin:
Can no longer reproduce hang issue.

PR #560, opened 2018-09-20
Crash service support

keianhzo:
Crash handler support.


app/src/common/shared/org/mozilla/vrbrowser/CrashReporterService.java
```diff
@@ -0,0 +1,76 @@
...
+                public void run() {
+                    boolean otherProcessesFound = false;
+                    for (final ActivityManager.RunningAppProcessInfo info : activityManager.getRunningAppProcesses()) {
+                        if (pid != info.pid) {
+                            otherProcessesFound = true;
+                        }
+                    }
+
+                    if (!otherProcessesFound) {
+                        Intent restartIntent = intent.setClass(CrashReporterService.this, VRBrowserActivity.class);
```
bluemarvin: I don't understand this part. From the docs intent.setClass() returns it self. So I don't know why you need to reset all the values it already has.

```
intent.setClass(CrashReporterService.this, VRBrowserActivity.class).addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
startActivity(intent);
```

Should be the same?

bluemarvin:
Unfortunately we can't land this until we figure out a work around for the Go crash on first run issue.

app/src/common/shared/org/mozilla/vrbrowser/CrashReporterService.java
keianhzo: The original intent it targeted to the CrashReporterService so we need to retarget it for the VRBActivity but you are right we don't need to reset the rest of the values. Changed that.

bluemarvin:
@keianhzo I pushed some fixes for Android P to this PR and made the crash reporter setting only grayed out on Oculus. Please take a look.

keianhzo:
@bluemarvin what's the issue with Android P? There is an infinite while loop in the latest committhat keeps on restarting the app whenever it's is closed after the first crash. We should always let the crash service finish after handling the crash intent.

bluemarvin:
I posted a fix for gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1493227
We should probably wait to land this until the fix is in Gecko or we have an alternative.

keianhzo:
@bluemarvin added a fix for Android P. I still can see issues but only in API level 26.

bluemarvin:
> @bluemarvin added a fix for Android P. I still can see issues but only in API level 26.

What issues are you seeing with 26? My S8 is running O right now.

keianhzo:
@bluemarvin it seems that system applies quite aggressive restrictions when the battery saver is on and the crash reporter service is killed upon second execution.

keianhzo:
@bluemarvin If it's working fine I guess we could land this and file a follow-up for testing in battery saving scenarios as I think this will only affect to Daydream "low end" clip-ins (S8).

MortimerGoro:
We need to wait for https://phabricator.services.mozilla.com/D7282 to land this

PR #567, opened 2018-09-24
Added support for emu x86

keianhzo:
Added support for android x86 emulator. Ideally we could add support for flavor dimensions and have targets for abi, platforms and channels (release, beta, nightly) so we can easily test all combinations. We can do that in the future.


bluemarvin:
This change looks like the noApi will now only build for x86? I still use the noApi for testing on arm.

keianhzo:
@bluemarvin Fixed that and added multi flavor support as it's a more standard way of dealing with this. Haven't added support for channels as I'm not really sure we need that yet and it generates a lot of variants.

bluemarvin:
this looks great. I wish I could do a test push to master to see if taskcluster has any problems with this before landing it for real.

bluemarvin:
I can build all the versions of the apk with:
```
./gradlew --no-daemon clean assembleGooglevr assembleNoapi assembleWavevr assembleSvr assembleOculusvr
```

But when I try to run noapix86 on an emulator by pressing the green play button in AndroidStudio I get the following error:
```
> Cannot build for ABI: x86 supported ABIs are: armeabi-v7a
```

keianhzo:
Fixed the MacOS ABI filtering issue, should work fine in all platforms now.

PR #578, opened 2018-09-26
Fix spinner during link traversal

MortimerGoro:
This PR fixes the spinner during link traversal and for the first frame after link traversal.

We need to land this to make link traversal work reliably: https://phabricator.services.mozilla.com/D6921 but this PR is not blocked by that GV patch

I use this sample for testing: https://webvr.info/samples/test-vr-links.html


bluemarvin:
I re-triggered the task. Looks like a gradle failure to pull down an AAR.

PR #579, opened 2018-09-26
Add support for MSAA selection in dev setting. Default MSAA to 2x.

MortimerGoro:


bluemarvin:
There should be a line between the the UA setting and MSAA setting.
screenshot_20180926_120129

bluemarvin:
Did the missing divider get added?

MortimerGoro:
Yes, forgot to comment that here

PR #588, opened 2018-09-28
Throttle hover events to 10Hz

bluemarvin:


app/src/main/cpp/Controller.cpp
```diff
@@ -80,6 +80,7 @@ Controller::Reset() {
   numAxes = 0;
   leftHanded = false;
   inDeadZone = true;
+  lastHoverEvent = 0.0;
```
MortimerGoro: We also need to add this new member to the copy assignment operator

MortimerGoro:
Code & local test looks good. Just handle the new variable also in the copy/assignment operator

PR #590, opened 2018-10-01
Support intent preferences for benchmarks

MortimerGoro:
Usage:

Ensure that the app is killed (GeckoView doesn't support changing preferences at runtime yet)
`adb shell am force-stop org.mozilla.vrbrowser`

Launch intent with the desided preference values. We have only exposed vr-user gesture and timer precision for now.

`adb shell am start -n  org.mozilla.vrbrowser/.VRBrowserActivity --es url "https://aframe.io" --ez dom.vr.require-gesture false --ez privacy.reduceTimerPrecision false`

Note, extra parameters don't work when using the android.intent.action.VIEW. That's why I added the optional `url` parameter.

`adb shell am start -a android.intent.action.VIEW -d "https://aframe.io" org.mozilla.vrbrowser/.VRBrowserActivity`

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/SessionStore.java
@@ -7,6 +7,7 @@ package org.mozilla.vrbrowser;
 
 import android.content.Context;
 import android.graphics.Rect;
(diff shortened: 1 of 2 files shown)


fernandojsg:
Looking great! thanks :)

PR #601, opened 2018-10-02
Update telemetry app name to be FLAVOR_platform.

daoshengmu:
We changed the behavior of BuildConfig.FLAVOR recently. We should update our telemetry app name to be FLAVOR_platform.


bluemarvin:
There was an update to taskcluster that broke a lot of repositories. I've pushed a fix to master. Rebase your patch and push it to the PR again.

daoshengmu:
it works now. Thanks

PR #604, opened 2018-10-05
Support added for loading local files through file:// handler (loadUri/reload)

### LINKED ISSUE #594: FxR is unable to load locally stored files
## Hardware

All
## Steps to Reproduce

1. Upload an HTML file to `/sdcard`
2. Attempt to open file in FxR with the path `file:///sdcard/somefile.html`

## Current Behavior

FxR reports access denied.
## Expected Behavior

FxR should prompt the user for permission then open the file.
## Possible Solution

FxR needs the user to grant permission to access local files. Currently FxR doesn't not prompt the user and requires that they grant permission from the system UI. Just like the microphone prompt, users should be prompted for this permission.

keianhzo:
Fixes #594 Added support for general app permissions (outside Gecko) handling and support for loading content from external storage upon loadUri/reload.

See https://github.com/MozillaReality/FirefoxReality/issues/594#issuecomment-427354841

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/PermissionDelegate.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/PermissionDelegate.java
@@ -1,8 +1,11 @@
 package org.mozilla.vrbrowser;
 
+import android.Manifest;
 import android.app.Activity;
 import android.content.Context;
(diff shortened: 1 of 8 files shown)


keianhzo:
Found the right solution using async GeckoResult so closing this one.

PR #605, opened 2018-10-05
Support added for loading local files through file:// handler

### LINKED ISSUE #594: FxR is unable to load locally stored files

keianhzo:
Fixes #594 Added support for general app permissions (outside Gecko) handling and support for loading content from external storage upon onLoadRequest.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/PermissionDelegate.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/PermissionDelegate.java
@@ -1,8 +1,10 @@
 package org.mozilla.vrbrowser;
 
+import android.Manifest;
 import android.app.Activity;
 import android.content.Context;
 import android.content.pm.PackageManager;
+import android.support.annotation.NonNull;
 import android.util.Log;
 
 import org.mozilla.geckoview.GeckoSession;
@@ -54,7 +56,7 @@ public class PermissionDelegate implements GeckoSession.PermissionDelegate, Widg
         }
     }
 
-    private void handleContentPermission(final String aUri, final PermissionWidget.PermissionType aType, final Callback aCallback) {
+    public void handlePermission(final String aUri, final PermissionWidget.PermissionType aType, final Callback aCallback) {
         if (mPermissionWidget == null) {
             mPermissionWidget = new PermissionWidget(mContext);
             mPermissionWidget.getPlacement().parentHandle = mParentWidgetHandle;
@@ -122,7 +124,7 @@ public class PermissionDelegate implements GeckoSession.PermissionDelegate, Widg
             return;
         }
 
-        handleContentPermission(aUri, type, callback);
+        handlePermission(aUri, type, callback);
     }
 
     @Override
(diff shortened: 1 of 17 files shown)


app/src/common/shared/org/mozilla/vrbrowser/PermissionDelegate.java
```diff
@@ -1,8 +1,11 @@
 package org.mozilla.vrbrowser;
 
+import android.Manifest;
 import android.app.Activity;
 import android.content.Context;
 import android.content.pm.PackageManager;
+import android.support.annotation.NonNull;
+import android.support.v4.content.ContextCompat;
```
MortimerGoro: I don't think that we need to use android.support.vr because our min API is v24.

keianhzo:
@MortimerGoro fixed the ContextCompat issue.
@bluemarvin I have replaced the text for "Will you allow Firefox Reality to read you external storage?" as in this case is the app who is asking for the permission not the web site.

app/src/main/res/values/strings.xml
```diff
@@ -15,6 +15,8 @@
     <string name="permission_location">Will you allow %1$s to access your location?</string>
     <!-- Parameter will be replaced with the app name -->
     <string name="permission_notification">Will you allow %1$s to send notifications?</string>
+    <!-- Parameter will be replaced with the app name -->
+    <string name="permission_read_external_storage">Will you allow %1$s to read you external storage?</string>
```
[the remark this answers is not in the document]
keianhzo: Thanks! Fixed

app/src/common/shared/org/mozilla/vrbrowser/ui/NavigationBarWidget.java
```diff
@@ -394,12 +395,45 @@ public void onCanGoForward(GeckoSession aSession, boolean canGoForward) {
...
+            Uri uri = Uri.parse(aUri);
+            if (uri.getScheme().equals("file")) {
+                if (!mWidgetManager.isPermissionGranted(android.Manifest.permission.READ_EXTERNAL_STORAGE)) {
+                    mWidgetManager.requestPermission(
+                            aUri,
+                            android.Manifest.permission.READ_EXTERNAL_STORAGE,
+                            new GeckoSession.PermissionDelegate.Callback() {
+                                @Override
+                                public void grant() {
+                                    result.complete(false);
```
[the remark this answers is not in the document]
keianhzo: The GeckoResult completion value tells Gecko whether or not we are going to handle the request. Returning true will mean that Gecko wouldn't either load the page or show the error page so we need to return false in both cases.

app/src/main/res/values/strings.xml
```diff
@@ -15,6 +15,8 @@
     <string name="permission_location">Will you allow %1$s to access your location?</string>
     <!-- Parameter will be replaced with the app name -->
     <string name="permission_notification">Will you allow %1$s to send notifications?</string>
+    <!-- Parameter will be replaced with the app name -->
+    <string name="permission_read_external_storage">Will you allow %1$s to read your external storage?</string>
```
bluemarvin: I guess I still find the wording confusing because the FxR dialog isn't actually granting permission, it is only asking the user if they want FxR to ask for permission. Then a system dialog is show (at least on Oculus) that requests that the user grant permission.

cvan: what is the benefit/rationale to introducing the initial permission prompt?

ideally, we should follow the [Permissions API standard](https://w3c.github.io/permissions/#examples), correct?

keianhzo: The [Android recommended flow](https://developer.android.com/training/permissions/requesting#explain) for permissions request includes showing a dialog so you have a chance to explain to the users why you need that permissions before they are prompted by the system.

In our case there are two different sets of permissions, the ones requested by the website and the ones requested by the app. 
Site permissions are requested per site. Gecko will grant permission to each site based on our custom dialog response. The system dialog is only shown once if the permission is already granted.
App permissions are only shown once for the whole app so for example if the user wants to access the internal storage to load a page it will prompted only once in case the user grants the permission.
In both cases we show the same dialog (that requires redesign btw), we just change the requester:
"Will you allow X to Y?" where X can be either "Firefox Reality" or a Website URL.

Oculus Browser does the same thing as us. It shows a banner at the bottom of the page saying that "xyz.com want's to use your location" if you say "Ok" the system dialog is shown.

Samsung Internet request permissions upon first start and doesn't let you use the browser unless you grant both external storage and location (for all sites) which is not the recommended flow as you should only request permissions when the are needed.

cvan: thanks for writing this all up. this is all helpful information.

ah, I understand now. and, you're right, Samsung Internet does prompt upon a fresh install -> launch for the File read permissions.

image

image

see my [review comments](https://github.com/MozillaReality/FirefoxReality/pull/605#pullrequestreview-163591570), though, because I have the accepted the File read permissions.

another question: does Android conflate the "Storage" permission to mean both read _and_ **write** access? because when I granted the Android permissions, I noticed that both permissions got granted (from running `adb shell dumpsys package org.mozilla.vrbrowser`):

    ```
          runtime permissions:
            android.permission.READ_EXTERNAL_STORAGE: granted=true
            android.permission.WRITE_EXTERNAL_STORAGE: granted=true
    ```

I expected it to set only `android.permission.READ_EXTERNAL_STORAGE`.

keianhzo: Don't have an explanation for this but for some reason it seems to be Oculus related. It is not happening in the noapi or Daydream flavors. I've checked the merged manifests and I can't find a reason so it might be something that Oculus is doing under the hood.

keianhzo:
@cvan 

1) 2) This is Gecko behaviour, it seems that for:
- file:/// if permission is granted it return a ERROR_FILE_NOT_FOUND onLoadError
- File:///sdcard/ (that gets redirected to file:///storage/emulated/0 and Gecko does nothing so it seems to be a Gecko issue

In both cases above I think it should return an ERROR_FILE_NOT_FOUND. I'm going to open a GV issue to address that issue.

3) Added an issue for this one #611

2) 3) I'll ask Nadja for a new design for this dialog as this UI is quite old, that should fix the white border and icon issues #612

5) This might be an issue and even could pause immersive mode. Will test and open an issue with the result.

6) If the user clicks on "Don't ask again" the only way of reenabling the permission is manually via the app settings or uninstalling the app. There is also no way of knowing if the user click on that checkbox.

keianhzo:
@cvan I've updated the code to dim the environment when a permission prompt is shown. I'll address the rest of the issues in the other follow-ups.

PR #607, opened 2018-10-09
Refactored Developer Options code

### LINKED ISSUE #580: Refactor developer options code
I think that the code can be simplified a lot creating some templates for the different type of options (e.g. radio options, toggles) and some mapping classes. That way we'd only need an array of settings and adding new ones or changing order would be trivial.

It might be interesting to also switch to a ListView to improve scrolling performance.

keianhzo:
Fixes #580 Refactored developer options code. Haven't replaced the scrollview for a listview as we don''t have so many elements as to affect the performance and would require some more work.


keianhzo:
Thanks @MortimerGoro! Default MSAA should be ok now

app/src/main/res/values/strings.xml
```diff
@@ -35,7 +35,7 @@
     <string name="developer_options_events">Input Events</string>
     <string name="developer_options_display_density">Display Density:</string>
     <string name="developer_options_window_size">Window Size:</string>
-    <string name="developer_options_window_size_by">by</string>
+    <string name="developer_options_by">By</string>
```
cvan: the L10n string should be like `Window Size: {width} × {height}` or `Window Size: {width} by {height}`

we can't use the word `By` by itself, because the ordering of words doesn't make sense in all other languages

app/src/main/res/values/strings.xml
```diff
@@ -51,10 +51,10 @@
     <string name="developer_options_ua_vr">VR</string>
     <string name="developer_options_reset">Reset Developer Settings</string>
     <string name="developer_options_reset_button">Reset</string>
-    <string name="developer_options_env">Environment</string>
-    <string name="developer_options_env_meadow">Meadow</string>
-    <string name="developer_options_env_cave">Cave</string>
-    <string name="developer_options_env_void">Void</string>
+    <string name="developer_options_environments">Environment</string>
+    <string name="developer_options_env_meadow">meadow</string>
```
cvan: why ar these lowercased but the `Pointer Color` selections, for example, are not?

keianhzo:
@cvan Fixed those two, thanks!

bluemarvin:
Looks like this is ready to land. Any outstanding issues can be addressed in a follow up issue.

PR #609, opened 2018-10-10
Fixed focus handling for VoiceSearch

keianhzo:
Focus improvements. Fixes focus handling for the voice search widget.


app/src/common/shared/org/mozilla/vrbrowser/ui/NavigationBarWidget.java
```diff
@@ -90,7 +90,7 @@ public void run() {
         mBackButton.setOnClickListener(new OnClickListener() {
             @Override
             public void onClick(View v) {
-                requestFocusFromTouch();
+                mBackButton.requestFocusFromTouch();
```
MortimerGoro: could you use `v.requestFocusFromTouch()` instead of the variable heare and in all the other buttons? This will avoid some duplicated references, a object with a listener with holds himself

keianhzo:
@MortimerGoro Yep, makes sense. Fixed.

PR #619, opened 2018-10-12
Hide home page redirect javascript URLs

### LINKED ISSUE #617: `javascript:window.location.replace…` appears in URL bar upon fresh install
## Hardware
Oculus Go

## Steps to Reproduce
1. Uninstall Firefox Reality (`adb uninstall org.mozilla.vrbrowser`).
2. Reinstall FxR.
3. Launch FxR.
4. Notice that as the URL is loading, `javascript:window.location.replace…` appears before being hidden after a few seconds.

## Current Behavior
As the URL is loading, `javascript:window.location.replace…` appears before being hidden after a few seconds.

image

## Expected Behavior
As the URL is loading, the `Search the Web or enter address` hint/placeholder text should appear immediately while the Home/Start page is loaded.

keianhzo:
Fixes #617  Hide all javascript: urls from the URL bar

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/SessionStore.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/SessionStore.java
@@ -90,6 +90,7 @@ public class SessionStore implements GeckoSession.NavigationDelegate, GeckoSessi
         boolean mIsInputActive;
         GeckoSession.ProgressDelegate.SecurityInformation mSecurityInformation;
         String mUri;
+        String mPreviousUri;
         String mTitle;
         boolean mFullScreen;
         GeckoSession mSession;
@@ -495,6 +496,18 @@ public class SessionStore implements GeckoSession.NavigationDelegate, GeckoSessi
         return result;
     }
 
+    public String getPreviousUri() {
+        String result = "";
+        if (mCurrentSession != null) {
+            State state = mSessions.get(mCurrentSession.hashCode());
+            if (state == null) {
+                return result;
+            }
+            result = state.mPreviousUri;
+        }
+        return result;
+    }
+
     public boolean isInputActive(int aSessionId) {
         SessionStore.State state = mSessions.get(aSessionId);
         if (state != null) {
@@ -798,6 +811,7 @@ public class SessionStore implements GeckoSession.NavigationDelegate, GeckoSessi
             return;
         }
 
+        state.mPreviousUri = state.mUri;
         state.mUri = aUri;
 
         for (GeckoSession.NavigationDelegate listener: mNavigationListeners) {
--- a/app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
@@ -210,30 +210,35 @@ public class TelemetryWrapper {
         if (uri == null)
             return;
 
-        URI uriLink = URI.create(uri);
-        if (uriLink.getHost() == null) {
-            return;
+        try {
+            URI uriLink = URI.create(uri);
+            if (uriLink.getHost() == null) {
+                return;
+            }
+
+            domainMap.add(UrlUtils.stripCommonSubdomains(uriLink.getHost()));
+            numUri++;
+
+            long elapsedLoad = SystemClock.elapsedRealtime() - startLoadPageTime;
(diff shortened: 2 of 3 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/NavigationURLBar.java
```diff
@@ -135,6 +135,8 @@ else if (aURL.startsWith("resource:") || SessionStore.get().isHomeUri(aURL))
                 aURL = "";
             else if (aURL.startsWith("data:") && SessionStore.get().isCurrentSessionPrivate())
                 aURL = "";
+            else if (aURL.startsWith("javascript:"))
```
cvan: nah, we should not always hide them. if a user enters a URL like this or gets redirected/clicked, it should show in the URL bar. can we hide it for only when the Home page is loaded?

keianhzo: Fixed

app/src/common/shared/org/mozilla/vrbrowser/ui/NavigationURLBar.java
```diff
@@ -135,7 +135,7 @@ else if (aURL.startsWith("resource:") || SessionStore.get().isHomeUri(aURL))
                 aURL = "";
             else if (aURL.startsWith("data:") && SessionStore.get().isCurrentSessionPrivate())
                 aURL = "";
-            else if (aURL.startsWith("javascript:"))
+            else if (aURL.startsWith("javascript:") && SessionStore.get().isHomeUri(aURL))
```
cvan: not sure how this could work because `aURL` cannot start with `javascript:` and also start with `https://webxr.today/`

keianhzo:
@cvan Should be fine now. For some reason I couldn't see the "javascript:" string with the previous patch.

app/src/common/shared/org/mozilla/vrbrowser/ui/NavigationURLBar.java
```diff
@@ -135,6 +135,8 @@ else if (aURL.startsWith("resource:") || SessionStore.get().isHomeUri(aURL))
                 aURL = "";
             else if (aURL.startsWith("data:") && SessionStore.get().isCurrentSessionPrivate())
                 aURL = "";
+            else if (aURL.startsWith("javascript:") && aURL.contains(SessionStore.get().getHomeUri()))
```
cvan: this fixes the bug, but it could trigger false positives. can we check that the previous URL started with the Home URI (`SessionStore.HOME_WITHOUT_REGION_ORIGIN`), and, if so, any URL changes or redirect (e.g., [`"javascript:window.location.replace('" + getHomeUri() + "')"`](https://github.com/MozillaReality/FirefoxReality/blob/81cbd41/app/src/common/shared/org/mozilla/vrbrowser/browser/SessionStore.java#L457)) gets correctly masked (i.e., `aURL remains ""`).

is there a way to set a flag for the active session before the [`loadUri` Gecko call here](https://github.com/MozillaReality/FirefoxReality/blob/81cbd41/app/src/common/shared/org/mozilla/vrbrowser/browser/SessionStore.java#L457) that will tell us "keep the URL empty when we are navigating on the Home page"?

cvan:
also, can you change the commit/PR message to reflect what's being changed here (i.e., this is specific to hiding the `javascript:window.location.replace(…)` redirect for the Home/Start URI

keianhzo:
Fixed.

app/src/common/shared/org/mozilla/vrbrowser/ui/NavigationURLBar.java
keianhzo: > is there a way to set a flag for the active session before the loadUri Gecko call here that will tell us "keep the URL empty when we are navigating on the Home page"?

There is no way of doing that in Gecko ATM.

> this fixes the bug, but it could trigger false positives. can we check that the previous URL started with the Home URI (SessionStore.HOME_WITHOUT_REGION_ORIGIN), and, if so, any URL changes or redirect (e.g., "javascript:window.location.replace('" + getHomeUri() + "')") gets correctly masked (i.e., aURL remains "").

Which are the false positives? Can you give me some example to reproduce? That will only leave the URL empty if it's any javascript URL that includes the home URL.

keianhzo:
@cvan Thanks for the URL I've added a getPreviousUri to deal with this. Also the URl have helped unveil a crash in Telemetry so I've also added a fix for that.

PR #498, opened 2018-09-10
Support for ServoSession

paulrouget:
This is an early PR to support Servo. It toggles from Gecko to Servo when the homebutton is pressed (just as a temporary trigger). It overrides GeckoSession and a few other classes.

I'd like to get an early feedback to know if this approach is sensible.

@MortimerGoro Can I ask you to take a look?

PR #632, opened 2018-10-16
Only show the servo button for few websites

paulrouget:
This is for after #498. We only want to show the button for some urls.
I moved the button to the right of the url bar to prevent the UI to "move" too much as the user navigates.

@jdm can you look at the list and tell me if there's anything else you want to add in there?


MortimerGoro:
Merging this in `servo`branch. We'll land the changes in servo branch to master after v1.1.

PR #633, opened 2018-10-16
Adding support for speech library 1.0.5 containing multi-lingual capabilities and the possibility to set a tag do identify the product using the speech API.

andrenatal:


andrenatal:
Yes, we need, but the intent of this PR is to land support only for the API's new version. To add support for new languages we'll need to design and develop some UI so the user could choose the language he wants so I'd do it in two different PRs.

larsbergstrom:
Would the user select the language or would we just pick up the system language? Or do you mean in the keyboard/voice system we should select the language? Actually, I'll open up a separate design issue for that...

andrenatal:
Speech API's version 1.0.5 also introduces two new methods:
```     
            mMozillaSpeechService.storeSamples(bool);
            mMozillaSpeechService.storeTranscriptions(bool);
``` 

The former tells the backend to save the audio sample, and the latter to save the transcribed sentence.

andrenatal:
@larsbergstrom The way I'd do in VoiceFill is to create a panel so the user could pick his preferred language, so maybe we can do the same on Fxr. 

Take my situation for example: my system locale and language is English, but I prefer to do voice search in Brazilian-Portuguese. 

In addition to that, there's the issue that the UA can return some very weird locale codes (my Firefox returns `en-US,en;q=0.5` for example), and if we pass that to the backend, it will return a header error since we validate the languages against this set:  https://github.com/mozilla/speech-proxy/blob/master/languages.json

So for these reasons, just pass the the system's locale code over to the API without any validation can be problematic and lead to issues.

andrenatal:
Ok, taskcluster is now passing. So this is good to merge whenever you guys decide so.

Issue #643, opened 2018-10-18
Error pages refactor to use A-C browser-errorpages component


keianhzo:
## Hardware
All

## Current Behavior
All the existing work for error pages (code + strings) has been committed to the A-C project and we are just waiting for a Fix to be released in the next A-C release to migrate error pages to A-C. As soon as that's ready error pages can be migrated. 

This is required by the I10n team and #642 so we make sure we are not doing I10n work twice.

## Possible Solution
Migrate to A-C browser-errorpages component.

## Error Logs and Stack Traces
<details open>

```
```

</details>

keianhzo:
Closing, already merged.

PR #626, opened 2018-10-15
Add localizations notes to the application strings. Fixes #625 #624

bluemarvin:
@Delphine This is a work in progress. Would you look at what I have so far and make sure they are adequate? @cvan Do you think you could add the LOCALIZATION NOTEs to the localpages.xml file?

PR #644, opened 2018-10-18
Refactored FxR error pages to A-C browser-errorpages component

### LINKED ISSUE #642: Make Error pages localizable
So I've had a few conversations with a number of people, and I get different information on what needs to be localized for error pages.

Seems like you guys have Error pages here: 
* https://github.com/MozillaReality/FirefoxReality/blob/master/app/src/main/res/values/localpages.xml
but then there's also Android-Components here:
* https://github.com/mozilla-mobile/android-components/blob/master/components/browser/errorpages/src/main/res/values/strings.xml

The Android-Components ErrorPages need to be localized no matter what, so since that's the first info I got (ie., that it was also needed for Reality), we started working on making it localizable. The work is, I believe, basically done (they fixed strings and added localization comments).

My question is: since these strings are basically the same, could you backport this work here? Isn't the plan to use a-c across mobile products as much as possible anyways?
If we can't do that, I'll open another issue and the same work will have to be done on the localpages.xml

also CC @pike

keianhzo:
Fixes #642 #643  Refactored FxR error pages to A-C browser-errorpages component.

**Note**: Private browsing  in not available in A-C so we still need to localize PB related string at localpages.xml

cc\ @Delphine for review


app/src/main/res/values/localpages.xml
```diff
@@ -7,207 +7,6 @@
     <string name="errorpage_title">Problem loading page</string>
```
cvan: these are also defined in the A-C strings. can we use those instead?

keianhzo: Good catch. Thanks!

keianhzo:
@cvan @Delphine I've addressed all the issues above.

app/src/main/res/values/localpages.xml
```diff
@@ -3,217 +3,12 @@
...
-<p>This web page at %1$s has been reported as a deceptive site and has been blocked based on your security preferences.</p>
-     ]]></string>
-
     <!-- Private Browsing -->
     <string name="private_browsing_title">Private Browsing</string>
     <string name="private_browsing_body"><![CDATA[
 <h1>Private Browsing</h1>
 
-<p>When you browse in a Private Window, Firefox Reality <strong>does not save</strong>:</p>
+<p>When you browse in a Private Window, %appName% <strong>does not save</strong>:</p>
```
Delphine: I don't think that %appName% is supported  by our l10n tooling (see [this comment](https://github.com/MozillaReality/FirefoxReality/pull/626#discussion_r225403197)). Could you please use %1$S instead?
Otherwise I think this is good to go :) thanks!

cvan:
> Fixes #642 #643 Refactored FxR error pages to A-C browser-errorpages component.

can you include ` (fixes #642 and #643)` to the message of the Pull Request and/or parent commit? thanks!

app/src/main/res/values/localpages.xml
keianhzo: Fixed.

app/src/common/shared/org/mozilla/vrbrowser/utils/InternalPages.java
```diff
@@ -234,7 +163,8 @@ public static String createErrorPage(Context context,
 
         html = html
                 .replace("%pageTitle%", context.getString(R.string.private_browsing_title))
-                .replace("%pageBody%", context.getString(R.string.private_browsing_body))
+                .replace("%pageBody%", context.getString(R.string.private_browsing_body)
+                        .replace("%appName%", context.getString(R.string.app_name)))
```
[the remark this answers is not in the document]
keianhzo: Fixed

app/src/main/res/values/localpages.xml
```diff
@@ -222,7 +17,7 @@
     <li>temporary files</li>
 </ul>
 
-<p>Private Browsing <strong>doesn\'t make you anonymous</strong> on the Internet. Your employer or Internet service provider can still know what page you visit.</p>
+<p>Private Browsing <strong>does not make you anonymous</strong> on the Internet. Your employer or Internet service provider can still know what page you visit.</p>
```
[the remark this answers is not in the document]
keianhzo: Fixed

PR #652, opened 2018-10-22
Oculus Go Rendering Optimizations

kearwood:
- Reduced FOV to 85%, increasing effective pixel density
- Decreased resolution of frame buffer to 1.0x the VRAPI's suggested value.
- Increased MSAA from 2x to 4x

Results in reducation of memory bus utilization by 1GB/s.


MortimerGoro:
We can close this because finally we were able to reduce the resolution of frame buffer to 1.0x implementing TimeWarp Layers

PR #654, opened 2018-10-22
Enable Oculus Fixed Foveated Rendering

kearwood:
A value of 2 was selected through experimentation on the Oculus Go. (0 = none, 3 = max).

Ideally, this would be a user-select-able value; however, a default of 2 seems reasonable as I could not notice any artifacts in the peripheral vision.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/DeveloperOptionsWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/DeveloperOptionsWidget.java
@@ -14,6 +14,7 @@ import android.widget.RadioGroup;
 import org.mozilla.vrbrowser.R;
 import org.mozilla.vrbrowser.browser.SessionStore;
 import org.mozilla.vrbrowser.browser.SettingsStore;
+import org.mozilla.vrbrowser.BuildConfig;
 import org.mozilla.vrbrowser.audio.AudioEngine;
 import org.mozilla.vrbrowser.ui.views.UIButton;
 import org.mozilla.vrbrowser.ui.settings.ButtonSetting;
@@ -41,6 +42,7 @@ public class DeveloperOptionsWidget extends UIWidget {
     private RadioGroupSetting mPointerColorRadio;
     private RadioGroupSetting mUaModeRadio;
     private RadioGroupSetting mMSAARadio;
+    private RadioGroupSetting mFoveatedRadio;
 
     private SingleEditSetting mDensityEdit;
     private SingleEditSetting mDpiEdit;
@@ -126,6 +128,15 @@ public class DeveloperOptionsWidget extends UIWidget {
         mMSAARadio = findViewById(R.id.msaa_radio);
         mMSAARadio.setOnCheckedChangeListener(mMSSAChangeListener);
         setMSAAMode(mMSAARadio.getIdForValue(msaaLevel), false);
+
+        mFoveatedRadio = findViewById(R.id.foveated_radio);
+        if (BuildConfig.FLAVOR_platform == "oculusvr") {
+            int foveatedLevel = SettingsStore.getInstance(getContext()).getFoveatedLevel();
+            mFoveatedRadio.setOnCheckedChangeListener(mFoveatedChangeListener);
+            setFoveatedLevel(mFoveatedRadio.getIdForValue(foveatedLevel), false);
+        } else {
+            mFoveatedRadio.setVisibility(View.GONE);
+        }
 
         mDensityEdit = findViewById(R.id.density_edit);
         mDensityEdit.setFirstText(Float.toString(SettingsStore.getInstance(getContext()).getDisplayDensity()));
@@ -242,6 +253,13 @@ public class DeveloperOptionsWidget extends UIWidget {
         @Override
         public void onCheckedChanged(RadioGroup radioGroup, int checkedId, boolean doApply) {
             setMSAAMode(checkedId, true);
+        }
+    };
+
+    private RadioGroupSetting.OnCheckedChangeListener mFoveatedChangeListener = new RadioGroupSetting.OnCheckedChangeListener() {
+        @Override
+        public void onCheckedChanged(RadioGroup radioGroup, int checkedId, boolean doApply) {
+            setFoveatedLevel(checkedId, true);
         }
     };
 
@@ -411,6 +429,18 @@ public class DeveloperOptionsWidget extends UIWidget {
         if (doApply) {
             SettingsStore.getInstance(getContext()).setMSAALevel((Integer)mMSAARadio.getValueForId(checkedId));
             showRestartDialog();
+        }
+    }
+
+    private void setFoveatedLevel(int checkedId, boolean doApply) {
+        mFoveatedRadio.setOnCheckedChangeListener(null);
+        mFoveatedRadio.setChecked(checkedId, doApply);
+        mFoveatedRadio.setOnCheckedChangeListener(mFoveatedChangeListener);
+
+        SettingsStore.getInstance(getContext()).setFoveatedLevel((Integer)mFoveatedRadio.getValueForId(checkedId));
+
+        if (doApply) {
+            mWidgetManager.updateFoveatedLevel();
         }
     }
 
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -102,6 +102,21 @@
     <string name="developer_options_msaa_2">2x</string>
     <!-- This string is used to label the MSAA radio button that enables four times MSAA in immersive mode. -->
     <string name="developer_options_msaa_4">4x</string>
+    <!-- This string is used to label radio buttons for setting fixed-foveated-rendering (FFR) level.
+         Higher values of this option result in larger parts of the peripheral vision being rendered
+         at lower resolutions than the content in the center of the view.  These higher values reduce
+         quality to increase performance.  The Oculus implementation of fixed-foveated-rendering is
+         described in more detail here:
+         https://developer.oculus.com/blog/optimizing-oculus-go-for-performance/-->
+    <string name="developer_options_foveated">Foveation Level</string>
+    <!-- This string is used to label the Foveated radio button that disables fixed-foveated-rendering in immersive mode. -->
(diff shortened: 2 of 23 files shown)


app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -172,6 +172,10 @@ struct DeviceDelegateOculusVR::State {
     vrapi_SetPropertyInt(&java, VRAPI_BLOCK_REMOTE_BUTTONS_WHEN_NOT_EMULATING_HMT, 0);
     // Reorient the headset after controller recenter.
     vrapi_SetPropertyInt(&java, VRAPI_REORIENT_HMD_ON_CONTROLLER_RECENTER, 1);
+    // Enable fixed-foveated rendering.
+    // TODO - This should be user configurable. VRAPI_SYS_PROP_FOVEATION_AVAILABLE
+    //        can be checked to determine if the option should be presented.
+    vrapi_SetPropertyInt(&java, VRAPI_FOVEATION_LEVEL, 2);
```
MortimerGoro: I think it's better to enable this only for the chrome scene. You can do that via`DeviceDelegateOculusVR::SetRenderMode()`

According to the spec foveated rendering for WebVR should be up to the JS code right? via WebGL extensions or WebXR when available.

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -172,6 +172,10 @@ struct DeviceDelegateOculusVR::State {
     vrapi_SetPropertyInt(&java, VRAPI_BLOCK_REMOTE_BUTTONS_WHEN_NOT_EMULATING_HMT, 0);
     // Reorient the headset after controller recenter.
     vrapi_SetPropertyInt(&java, VRAPI_REORIENT_HMD_ON_CONTROLLER_RECENTER, 1);
+    // Enable fixed-foveated rendering.
+    // TODO - This should be user configurable. VRAPI_SYS_PROP_FOVEATION_AVAILABLE
```
MortimerGoro: Check `VRAPI_SYS_PROP_FOVEATION_AVAILABLE` before calling `vrapi_SetPropertyInt(&java, VRAPI_FOVEATION_LEVEL, 2);`

app/src/main/res/values/strings.xml
```diff
@@ -102,6 +102,16 @@
     <string name="developer_options_msaa_2">2x</string>
     <!-- This string is used to label the MSAA radio button that enables four times MSAA in immersive mode. -->
     <string name="developer_options_msaa_4">4x</string>
+    <!-- This string is used to label radio buttons for setting fixed-foveated-rendering -->
```
cvan: can we add a note that explains in a little bit more detail what this means. I assume this string will be difficult to translate for most.

app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/DeveloperOptionsWidget.java
```diff
@@ -127,6 +128,15 @@ public void onClick(View view) {
         mMSAARadio.setOnCheckedChangeListener(mMSSAChangeListener);
         setMSAAMode(mMSAARadio.getIdForValue(msaaLevel), false);
 
+        int foveatedLevel = SettingsStore.getInstance(getContext()).getFoveatedLevel();
+        mFoveatedRadio = findViewById(R.id.foveated_radio);
+        if (!true) { // FINDME!! KIP!! HACK!!
```
bluemarvin: This should be:
`if (BuildConfig.FLAVOR_platform != "oculusvr") {`

app/src/main/res/values/strings.xml
```diff
@@ -102,6 +102,16 @@
...
+    <string name="developer_options_foveated">Foveated Level</string>
+    <!-- This string is used to label the Foveated radio button that disables fixed-foveated-rendering in immersive mode. -->
+    <string name="developer_options_foveated_disabled">Disabled</string>
+    <!-- This string is used to label the Foveated radio button that enables low-level fixed-foveated-rendering. -->
+    <string name="developer_options_foveated_1">1</string>
```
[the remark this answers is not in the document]
kearwood: I can only really notice the difference between #2 and #3.  With #3, you can see the edges having a slightly lower resolution when you move the headset so something with contrast is near the top or sides.

I originally wanted to use "low", "medium", and "high" but the longer strings broke the UI causing overlaps.  Perhaps we could improve this later with some changes to the developer options panel layout?

app/src/noapi/cpp/DeviceDelegateNoAPI.cpp
```diff
@@ -154,6 +154,18 @@ DeviceDelegateNoAPI::ReleaseControllerDelegate() {
...
 
+void
+DeviceDelegateNoAPI::SetFoveatedLevel(const uint32_t aLevel) {
+  // Not Supported
+}
+
+
+int32_t
+DeviceDelegateNoAPI::GetMaxFoveatedLevel() {
+  // Return 0 to indicate that it is not supported
```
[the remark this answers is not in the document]
kearwood: I've removed the GetMaxFoveatedLevel() related code and replaced it with a simple check for the Oculus build in DeveloperOptionsWidget.java.  My original implementation was not thread safe, and would take a lot longer to fix.  I'd like to see a better mechanism for selectively enabling platform specific UI in the options widget.

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -386,6 +386,21 @@ DeviceDelegateOculusVR::ReleaseControllerDelegate() {
...
+  vrapi_SetPropertyInt(&m.java, VRAPI_FOVEATION_LEVEL, aLevel);
+}
+
+int32_t
+DeviceDelegateOculusVR::GetMaxFoveatedLevel() {
+  int available = vrapi_GetSystemPropertyInt(&m.java, VRAPI_SYS_PROP_FOVEATION_AVAILABLE);
+  if (available == VRAPI_TRUE) {
+    return 3;
+  }
+  return 0;
```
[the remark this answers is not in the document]
kearwood: This code as been removed.  It wasn't clear how we plan to make the options dialogue dynamic to support other headsets that also support foveated rendering, so I opted for a simpler solution for now in DeveloperOptionsWidget.java

app/src/main/res/values/strings.xml
```diff
@@ -102,6 +102,16 @@
     <string name="developer_options_msaa_2">2x</string>
     <!-- This string is used to label the MSAA radio button that enables four times MSAA in immersive mode. -->
     <string name="developer_options_msaa_4">4x</string>
+    <!-- This string is used to label radio buttons for setting fixed-foveated-rendering -->
+    <string name="developer_options_foveated">Foveated Level</string>
```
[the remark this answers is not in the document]
kearwood: I've added some more comments.  I hope it helps!

I suspected "foveated" would be a non-localized term.

kearwood:
Sorry for the multiple pushes..  This final one seems to be working well.  I've tested on Oculus Go and HTC Vive Focus.  (On the Vive Focus, the foveated option disappears as intended)

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
kearwood: Agree, we should probably have this preference affect the browser's 2d modes separately.

There is no defined API for setting foveated level in the WebXR spec yet.  Unlike MRS (Mixed resolution shading) and LMS (Lens Matched Shading), this kind of fixed-foveated-rendering does not need changes to the content to function.  Turning it on manually for immersive content could make the difference between being able to see some immersive sites and it not working.  Perhaps we should give the user choice in this case.

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
kearwood: Initially (my earlier patch), I was attempting to check VRAPI_SYS_PROP_FOVEATION_AVAILABLE before enabling the UI, but it would require many more changes (eg, getting the value asynchronously from the render thread while building the UI).  Setting this value when it is not supported will simply be a NULL operation.  I've hidden the UI when not using the Oculus API, but perhaps we could enhance this at a later time, together with better adaptive UI selections for other rendering features?

kearwood: I imagine the only case right now where VRAPI_SYS_PROP_FOVEATION_AVAILABLE  would return false would be for GearVR / clip-in headsets using the Oculus API.  In those cases with the current patch, the option would appear but not have any effect.

MortimerGoro:
@kearwood could you rebase this PR with master? The other change required is disable/enable foveated when the user starts/exits a WebVR session. You could do that in SetRenderMode() based on the device::RenderMode::StandAlone or device::RenderMode::Immersive parameter

philip-lamb:
@kearwood It'd be great to get this into 1.1.3.

MortimerGoro:
@kearwood Do you have time to work on this? I can work in the remaining bits if you want

philip-lamb:
> The other change required is disable/enable foveated when the user starts/exits a WebVR session. You could do that in SetRenderMode() based on the device::RenderMode::StandAlone or device::RenderMode::Immersive parameter

I'm still not clear on why we'd not just apply fixed foveation mode all the time?

MortimerGoro:
> I'm still not clear on why we'd not just apply fixed foveation mode all the time?

The WebXR spec says: implementations should not enable foveated rendering or performance optimizations that may negatively affect the correctness of an application unless the application explicitly opts-in to that behavior.

philip-lamb:
> The WebXR spec says: implementations should not enable foveated rendering or performance optimizations that may negatively affect the correctness of an application unless the application explicitly opts-in to that behavior.

True, except that we don't have WebXR yet, and according to @kearwood above, there is no defined API for setting foveated level in the WebXR spec yet. I think we should make the choice in favour of performance now and offer the user an option to enable/disable when WebXR lands.

MortimerGoro:
PR moved to https://github.com/MozillaReality/FirefoxReality/pull/1064

Issue #651, opened 2018-10-22
Infinite restart loop on startup crashes


MortimerGoro:
## Steps to Reproduce
- Easy to test by creating a startup crash  (e.g. a error in `navigation_bar.xml`)

## Current Behavior
- Infinite restart loop

## Expected Behavior
- Try some times but not forever

## Possible Solution
- Detect the number of consecutive crashes and stop restarting the app after some threshold.

cvan:
@MortimerGoro opened PR #1202. per @bluemarvin the code change is blocked on https://github.com/MozillaReality/FirefoxReality/pull/1202#issuecomment-492024166.

MortimerGoro:
Already fixed in #1202

PR #656, opened 2018-10-23
Add a user property to disable app relaunch on crash

MortimerGoro:
I find the restart annoying while developing because the logs/stacktraces get back on the log history and may be lost.

Related issue: https://github.com/MozillaReality/FirefoxReality/issues/651


bluemarvin:
Turning off the crash reporting in debug will make testing and working on crash reporting more painful. Maybe you can add config that can be set in `user.properties`? Something like `disableCrashRestart=true`? Then it isn't just a debug vs release thing?

MortimerGoro:
I like the `user.properties` approach. I'll update the PR.

MortimerGoro:
PR updated with the user.properties solution

PR #661, opened 2018-10-24
Add strings for video mode

MortimerGoro:
Add these before waiting for the video mode landing so they can be translated earlier

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -166,6 +166,8 @@
     <string name="cancel_button">Cancel</string>
     <!-- This string labels a button that are used to approve an action. -->
     <string name="ok_button">OK</string>
+    <!-- This string labels a button to go back to the previous menu or exits the current mode -->
+    <string name="back_button">Back</string>
     <!-- This string is displayed in a dialog after an application crash. -->
     <string name="crash_dialog_heading">We had an issue and crashed</string>
     <!-- This string explains why Mozilla would like the user to allow the crash data to be uploaded.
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -166,6 +166,8 @@
     <string name="cancel_button">Cancel</string>
     <!-- This string labels a button that are used to approve an action. -->
     <string name="ok_button">OK</string>
+    <!-- This string labels a button that are used to approve an action. -->
```
keianhzo: Maybe "This string labels a button to go back to the previous menu or exit the current mode"?

app/src/main/res/values/strings.xml
```diff
@@ -181,4 +183,42 @@
...
+    <!-- This string is displayed when selecting a background brightness mode -->
+    <string name="brightness_mode_dark">Dark</string>
+    <!-- This string is displayed when selecting a background brightness mode -->
+    <string name="brightness_mode_void">Void</string>
+    <!-- This string labels a button to open the Video projection mode selector-->
+    <string name="video_mode_tooltip">Video Projection</string>
+    <!-- This string is displayed when selecting a video playback mode -->
+    <string name="video_mode_3d_side">3D Side by Side</string>
+    <!-- This string is displayed when selecting a video playback mode -->
+    <string name="video_mode_360">360</string>
```
keianhzo: Shouldn't numbers be in the non_I10n?

MortimerGoro: I wasn't sure all the languages use `180` or `360` to refer to that kind of videos.

keianhzo:
Let's the I10n team decide about the numbers when they check.

app/src/main/res/values/strings.xml
cvan: numerals can be localised, so yeah we should keep. a better L10n note might say `360` refers to `360°` (or `360-degree` videos)

PR #662, opened 2018-10-24
Improve background dim API

### LINKED ISSUE #657: Closing settings panel doesn't disable dimming
## Hardware

All
## Steps to Reproduce

1. Open settings panel
2. Close setting panel

## Current Behavior

The environment stays dimmed.
## Expected Behavior

The environment should no longer be dimmed.

MortimerGoro:
- Better API to set up the background dim value and go back to the previous ones. The current code breaks a lot.
- Fixes #657
- Provide more flexibility for the configurable brightness in the Video mode.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WidgetManagerDelegate.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WidgetManagerDelegate.java
@@ -19,6 +19,8 @@ public interface WidgetManagerDelegate {
         // Indicates that the user has clicked on the world, outside of any UI widgets
         void onWorldClick();
     }
+    float DEFAULT_DIM_BRIGHTNESS = 0.25f;
+
     int newWidgetHandle();
     void addWidget(@NonNull Widget aWidget);
     void updateWidget(@NonNull Widget aWidget);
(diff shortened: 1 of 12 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WidgetManagerDelegate.java
```diff
@@ -25,8 +27,9 @@
     void removeUpdateListener(@NonNull UpdateListener aUpdateListener);
     void pushBackHandler(@NonNull Runnable aRunnable);
     void popBackHandler(@NonNull Runnable aRunnable);
-    void fadeOutWorld();
-    void fadeInWorld();
+    void pushWorldBrightness();
+    void setWorldBrightness(float aBrightness);
```
bluemarvin: Do we need separate `setWorldBrightness()` and `pushWorldBrightness()` functions? Are there instances where we want to set the brightness and not have it push the current state?

MortimerGoro: all the cases should call `setWorldBrightness` after `pushWorldBrightness`. So we can add the parameter to setWorldBrightness. I'll update the PR

MortimerGoro:
I fixed the mentioned cases:
- I added a key for each brigthness setter/push so if a widget back in the hierarchy changes it's brightness it doesn't affect the current one (e.g. a tab changing from private to normal while a dialog in on top)
- Crash dialog was failing because there is a FadeIn() after the splash is shown. Now that fadeIn uses the correct final brightness.

PR #663, opened 2018-10-24
Improve widget depth sorting

MortimerGoro:
The depth sorting we are using for the widgets fails in some situations for the video mode UI (e.g. a widget on the right side vs one in the center but farther in z plane). This PR improves the depth sorting using distance to plane instead of distance to node position.

### DIFF
--- a/app/src/main/cpp/BrowserWorld.cpp
+++ b/app/src/main/cpp/BrowserWorld.cpp
@@ -871,11 +871,21 @@ BrowserWorld::DrawWorld() {
   m.externalVR->SetCompositorEnabled(true);
   m.device->SetRenderMode(device::RenderMode::StandAlone);
   vrb::Vector headPosition = m.device->GetHeadTransform().GetTranslation();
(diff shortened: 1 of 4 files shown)


app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -1109,6 +1119,28 @@ BrowserWorld::DistanceToNode(const vrb::NodePtr& aTargetNode, const vrb::Vector&
...
+}
+
+float
+BrowserWorld::DistanceToPlane(const WidgetPtr& aWidget, const vrb::Vector& aPosition, const vrb::Vector& aDirection) const {
+  if (!aWidget) {
+    return -1.0f;
+  }
+  vrb::Vector result;
+  bool inside;
+  float distance;
```
bluemarvin: nit: I would prefer `inside` and `distance` have default values.

MortimerGoro: done

PR #667, opened 2018-10-25
add `Bleeding-edge developer APKs` section at top of README

cvan:
if there's a better place, I'm all ears


cvan:
> Need to add an intermediary step to click `Firefox Reality for Android - Build - Master update` before clicking `Run Artifacts`

good catch. done, thanks.

PR #670, opened 2018-10-25
Do not call anything after session.close()

paulrouget:
This is a problem with servo sessions. `stop` and `setActive` can't work after `close` is called.


MortimerGoro:
It makes sense to call `setActive` and `stop` before `onClose` on GV too

Issue #695, opened 2018-10-30
Typo in Private Browsing splash page


cvan:
## Hardware
All

## Steps to Reproduce
1. Launch Firefox Reality
2. Click the **Private Browsing** icon in the Tray
3. Once the splash page loads, notice the typo

## Current Behavior
> Private Browsing **doesn’t save make you anonymous** on the Internet.

![image](https://user-images.githubusercontent.com/203725/47756946-4d341f00-dc61-11e8-8c91-6e2beba1daf2.png)

## Expected Behavior
> Private Browsing **doesn’t make you anonymous** on the Internet.

## Possible Solution
Remove `save ` from the sentence.

cvan:
fixed in PR #696

PR #696, opened 2018-10-30
fix typo in Private Browsing splash page (fixes issue #695)

cvan:


cvan:
> The curly quote doesn't render correctly.

@keianhzo: interesting, good catch. do you know why?

```sh
% file -I app/src/main/res/values/localpages.xml

app/src/main/res/values/localpages.xml: text/xml; charset=utf-8
```

keianhzo:
@cvan I guess it's probably because the page is missing the:

`<meta http-equiv="Content-Type" content="text/html; charset=utf-8">`

tag. We should add that for both the private browsing and the error pages at res/raw.

cvan:
> @cvan I guess it's probably because the page is missing the:
> 
> `<meta http-equiv="Content-Type" content="text/html; charset=utf-8">`
> 
> tag. We should add that for both the private browsing and the error pages at res/raw.

great catch, you're right. I filed #700 to update the pages accordingly. thanks!

otherwise, this PR good to go?

PR #702, opened 2018-11-01
WIP Layers implementation

kearwood:
Do not commit yet...  WIP


MortimerGoro:
Thanks Kip for the initial research. Here is the complete implementation: https://github.com/MozillaReality/FirefoxReality/pull/778

I had to change the API a lot to support everything and straight render to Android Surface textures instead of FBOs

PR #725, opened 2018-11-06
180/360/3D video support

MortimerGoro:


MortimerGoro:
Depends on https://phabricator.services.mozilla.com/D9026

cvan:
as `360` is probably most common, can we have that be the first option, instead of `3D Side by Side`?

cvan:
I had some compilation errors when I tried to rebase it locally against `master`. if you can rebase and update, that'd be awesome :smile: thanks!

thenadj:
Fine to change order to show 360 as first option. UIS won't be updated for that. Please make adjustments accordingly.

PR #704, opened 2018-11-01
Make VR User Agent the default. Fixes #478

bluemarvin:

PR #727, opened 2018-11-06
Change UA settings storage key

MortimerGoro:
Changed the key name so users who update the app without uninstalling it get the VR UA by default.


MortimerGoro:
True: https://github.com/MozillaReality/FirefoxReality/pull/704/files

Closing this

PR #730, opened 2018-11-07
Fixes #729, add localizations to FirefoxReality

### LINKED ISSUE #729: Localization of Target Locales
Localization of the following target locales:
Simplified Chinese (zh-CN), Traditional Chinese (zh-TW), French (fr), Italian (it), German (de), Spanish (es-ES), Japanese (ja), Korean (ko)

Target locales strings: https://drive.google.com/drive/folders/1_wWZvljcIKS7HyU1s69GU0OpZw1GZkKP?usp=sharing

@keianhzo Can you please help put each of the files in the correct directory?

Once this is done, we will need localized builds for testing.

Pike:
Let's get the ball rolling on #729 by adding the files from gdrive to a PR.

CC @delphine


Pike:
OK, that doesn't work that way.

The build fails on bundling the a-c strings. Hopefully we can resolve this inside fxr?

Looping on @pocmo, too.

pocmo:
What is `androidcomponents_strings.xml` exactly? Shouldn't those strings come with the component? Or is this a workaround until we can get them into the a-c repo?

Pike:
Yes, that was my naive attempt to get the strings in before we can localize the properly in a-c.

pocmo:
> Yes, that was my naive attempt to get the strings in before we can localize the properly in a-c.

Okay, in this case we can as a workaround disable the `ExtraTranslation` lint check. Try adding this here:
https://github.com/MozillaReality/FirefoxReality/blob/master/app/build.gradle#L64

Pike:
OK, now we have a green build. We'll need some actual testing of the app.

Sadly, the APK from the builds on PRs don't get exposed as artifacts.

app/src/main/res/values-de/androidcomponents_strings.xml
```diff
@@ -0,0 +1,260 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<resources>
+    <!-- The default document title of an error page (i.e., the contents of the `<title>` tag). -->
+    <string name="mozac_browser_errorpages_page_title">Seiten-Ladefehler</string>
```
cvan: I am glad these are localised - yay!

correct me if I am wrong, @keianhzo, these XML files should stay checked in to the android-components repo, correct?

cvan:
@pocmo @bluemarvin @keianhzo with Android Studio and the command line, how can I change and test locales? (à la `export LC_ALL="en_US.UTF-8"`) once I figure out how to change the locales, I can take screenshots and videos manually. (later we can get fancy with [automation like Focus for Android](https://github.com/mozilla-mobile/focus-android/wiki/l10n-Screenshot-Generation).)

pocmo:
@cvan Builds will always contain all locales. Switching the locale on your device/emulator should be enough to see other strings.

keianhzo:
@cvan Oculus Go gets the locales from the linked device after a factory reset and you can't change that unless you do another factory reset so I always test that using the emulator or a Daydream enabled phone just changing the device locale.

keianhzo:
@Pike As talked in Slack we might need to move values-jp to values-ja

Delphine:
All strings have been updated by agency, and can be found here:
https://drive.google.com/drive/u/1/folders/1_wWZvljcIKS7HyU1s69GU0OpZw1GZkKP?ogsrc=32

They correspond  in each case to:
* localpages2.xml
* strings2.xml

Issue #598, opened 2018-10-02
Consider removing clip-in support from Daydream store?


larsbergstrom:
Just wanted to test the waters here - we'd only accidentally enabled clip-in support in the Daydream store, and I'd like to propose removing it. Reasons:
- There are very few users
- Is totally untested for product usage and has a ton of one-off bugs
- Does not have a smooth user workflow for enabling the device

The backup here is that people can use the Sideload instructions on the wiki if they are developers (who are the main people I'm concerned about leaving behind with this change).

Note that it would still be available for use on the Lenovo Mirage AIO, as was originally intended.

Thoughts? 

cc @bluemarvin @avrignaud @MortimerGoro @keianhzo

larsbergstrom:
Related issue to the prompts:
https://github.com/MozillaReality/FirefoxReality/issues/583#issuecomment-426299847

avrignaud:
I'd agree - we prioritized and tested standalone headsets, and aren't seeing many users (at all) on the clip-ins. Also, the Pixel1 isn't a suitable device due to perf, which is causing some additional confusion in market.

nancyhang:
Will reach out to Ritu to turn the clip-ins support off. We will support standalone Daydream

larsbergstrom:
This is done. Thanks, @nancyhang!

RangerMauve:
Hi, I think it would be useful to have this mentioned somewhere in the README so that people won't be as confused when they can't install it on their Android phones.

Would a PR for the README update be welcome?

avrignaud:
Please, feel free! Thank you!
-- 
Andre Vrignaud — Head of Mixed Reality Platform Strategy | Emerging
Technologies | Mozilla
avrignaud@mozilla.com | +1 (206) 245-6067 <%2B1.206.245.6067> | Skype:
andrevrignaud

PR #731, opened 2018-11-07
Added note about Daydream on phones

RangerMauve:
As per https://github.com/MozillaReality/FirefoxReality/issues/598#issuecomment-436711115, I added a note to say why the APK isn't available on Daydream phones yet.


RangerMauve:
Awesome, thank you!

Issue #747, opened 2018-11-12
Remove dependency from `org.mozilla.gecko`


agi90:
At GeckoView, are trying to hide `org.mozilla.gecko` from consumers, as that package is not intended to be used outside of GeckoView. To do that we need to remove all existing dependencies from `org.mozilla.gecko` in FxR. I'll send a patch soon for this.

See also https://bugzilla.mozilla.org/show_bug.cgi?id=1499762

PR #753, opened 2018-11-13
Remove most dependencies from org.mozilla.gecko (#747)

agi90:
`org.mozilla.gecko` is private to geckoview and will hidden removed soon.

Remaining dependencies after this commit:

- GeckoVRManager, will be made available to GV consumers
- GeckoProfile, will add prefs to GV
- PanZoomController, will be moved to org.mozilla.geckoview


bluemarvin:
@agi90 The global exception handler is being used so just deleting it doesn't really fix the issue.

bluemarvin:
The FxR code has changed enough that I think it better just to close this and try again. There seem to be even more `.gecko` dependancies now.

PR #760, opened 2018-11-15
Fixes #347: Bookmarks support

### LINKED ISSUE #347: UR Finding: Users expect to be able to save bookmarks
UR Finding: Users expect to be able to save bookmarks. Explore implementing Favorite/Bookmark support.

keianhzo:
Fixes #347 #782 Bookmarks support. This requires merging also https://github.com/MozillaReality/vrb/pull/54. Updated deps to Jetpack. 

At the moment this only overlays the Bookmarks panel on top of the Browser window. we will need to do some more work for Surface sharing in a follow up.


app/schemas/org.mozilla.vrbrowser.db.BookmarkDatabase/1.json
```diff
@@ -0,0 +1,61 @@
+{
```
cvan: nice! :+1:

app/schemas/org.mozilla.vrbrowser.db.BookmarkDatabase/1.json
```diff
@@ -0,0 +1,61 @@
...
+          },
+          {
+            "fieldPath": "url",
+            "columnName": "url",
+            "affinity": "TEXT",
+            "notNull": false
+          },
+          {
+            "fieldPath": "timestamp",
+            "columnName": "timestamp",
```
cvan: can we rename this as `added`? perhaps we'll want to later add `lastVisited` or `lastModified`, etc.

cvan: (ignore this if these column names are consistent with those Gecko…)

can we rename this date field `timestamp` to `added`? perhaps we'll want to later add `lastVisited` or `lastModified`, etc.

keianhzo:
@cvan There some of outstanding issues pending fixes in master that affect this PR, so for now you can just use: 

https://drive.google.com/open?id=1lMxOLU79EoTnJgBjnpRYd-3edfkhIEbi

Avoid Bookmarks resizing, there still issues pending solve that affect this and the URL bar.

If you want to build the app you'll need to:
- Comment this line to avoid duplicated back button events https://github.com/MozillaReality/FirefoxReality/blob/453bd4a214a43c03b3775b8945500cafdf85709b/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java#L520
- Ese the following VRB branch: https://github.com/keianhzo/vrb/tree/bookmarks

cvan:
1. Press the **Private Browsing** icon in the Tray.
2. Press the **Bookmarks** icon in the Tray.
3. Notice the Tray state:

image

This is related to #774, etc. But figured I'd bring it up in this issue.

<hr>

Can you swap the **Bookmarks** icon with the **Settings** icon?

cvan:
> Avoid Bookmarks resizing, there still issues pending solve that affect this and the URL bar.

gotcha, I noticed the issues. will those be fixed before landing this PR and the other PRs? or later?

keianhzo:
@cvan I'd rather to land and fix it later as is something that's already broken in master and this way we could get bookmarks tested as much as possible in the meantime.

cvan:
> @cvan I'd rather to land and fix it later as is something that's already broken in master and this way we could get bookmarks tested as much as possible in the meantime.

that sounds good to me. after this other PRs get merged and I have a few moments, I'll file follow-up issues for cleanup.

thanks again for getting this done for 1.1. this is a super awesome feature that our users will be glad to have!

cvan:
tested the latest changes. can we file for the Resize Mode? and then it sounds like we're good to go! :+1:

Issue #754, opened 2018-11-13
VR video playback mode has artifacts


caseyyee:
## Hardware
Oculus Go
FXR: eaa4fc5

## Steps to Reproduce
1. Goto Youtube and watch any 360 or 180 content.
2. Fullscreen and Select a VR projection mode, you can see this in any of the modes.
3. Notice especially along top and bottom parts of the video, there is a shimmering/stairstepping/distortion effect when viewing the video.

You can see this in the grass of this video, the effect is much more pronounced in-headset:

![video-artifacts](https://user-images.githubusercontent.com/1560884/48444792-ce9fad00-e749-11e8-9b8c-56b11d6b6f7a.gif)

MortimerGoro:
https://github.com/MozillaReality/FirefoxReality/pull/778 fixes this on Oculus.

This artifact didn't happen on the first version (without the layers). So we still need to fix this for the other platforms

cvan:
note: fixed in PR #778 and #853

PR #778, opened 2018-11-19
Implement TimeWarp Layers

MortimerGoro:


app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
```diff
@@ -421,14 +422,52 @@ public void onFrameAvailable(SurfaceTexture surfaceTexture) {
...
             if (widget == null) {
                 MotionEventGenerator.dispatch(mRootWidget, aDevice, aPressed, aX, aY);
-
+            } else if (widget == mBrowserWidget && mBrowserWidget.getBorderWidth() > 0) {
+
+                final float extra = mBrowserWidget.getBorderWidth() * 2.0f;
+                final float rx = aX / mBrowserWidget.getTextureWidth();
+                final float ry = aY/ mBrowserWidget.getTextureHeight();
+                final float x = rx * (mBrowserWidget.getTextureWidth() - extra);
+                final float y = ry * (mBrowserWidget.getTextureHeight() - extra);
```
bluemarvin: Since the pixels are still the same size, I don't think you need to scale them? I think:
```
final int border = mBrowserWidget.getBorderWidth();
MotionEventGenerator.dispatch(widget, aDevice, aPressed, x - border, y - boarder);
```
Since we are just shifting the window by `border` number of pixels. The compositor doesn't know there is a border so pixel size should be unaffected.

app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/BrowserWidget.java
```diff
@@ -146,15 +167,45 @@ public void setSurfaceTexture(SurfaceTexture aTexture, final int aWidth, final i
...
+        } else {
+            mDisplay.surfaceDestroyed();
+        }
+        if (aFirstDrawCallback != null) {
+            aFirstDrawCallback.run();
+        }
+    }
+
+    private void callSurfaceChanged() {
+        mDisplay.surfaceChanged(mSurface, mBorderWidth, mBorderWidth, mWidth - mBorderWidth * 2, mHeight - mBorderWidth * 2);
```
bluemarvin: I wonder if we should be making the surface larger for the border? Or are you requesting a size larger than you want the content size to be? Basically, if we want our content to be 200x300 with a 1 pixel border, we should be requesting a surface size of 202x302.

app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/MediaControlsWidget.java
```diff
@@ -318,9 +318,7 @@ public void onPlaybackRateChange(MediaElement mediaElement, double rate) {
 
     @Override
     public void onFullscreenChange(MediaElement mediaElement, boolean fullscreen) {
-        if (!fullscreen) {
-            hide();
-        }
+
```
bluemarvin: nit: extra line?

app/src/main/cpp/DeviceDelegate.h
```diff
@@ -11,8 +11,11 @@
 #include "Device.h"
 #include "ControllerDelegate.h"
 #include "GestureDelegate.h"
+#include "VRLayer.h"
 
 #include <memory>
+#include <functional>
+#include <jni.h>
```
bluemarvin: What needs `jni.h` in this file?

app/src/main/cpp/DeviceDelegate.h
```diff
@@ -11,8 +11,11 @@
 #include "Device.h"
 #include "ControllerDelegate.h"
 #include "GestureDelegate.h"
+#include "VRLayer.h"
 
 #include <memory>
+#include <functional>
```
bluemarvin: What needs `functional`? I'm missing any changes that require it.

app/src/main/cpp/VRLayer.h
```diff
@@ -0,0 +1,149 @@
...
+#include "Device.h"
+
+#include <memory>
+#include <functional>
+#include <jni.h>
+
+namespace crow {
+
+class VRLayer;
+typedef std::shared_ptr<VRLayer> VRLayerPtr;
```
bluemarvin: I'm starting to think we should add a `CrowForward.h` or something similar.

caseyyee:
Getting some strange smearing in awesomebar results:
screen

MortimerGoro:
@caseyyee layers require a transparent border. That means that we lost the border on the awesome bar in some situation.
STR? I didn't get it while testing awesome bar

caseyyee:
Hmm yeah, doesn't seem to be occuring for me after restart anymore. Will keep an eye out.

Also looks like issue #754 no longer seems to be an issue.

app/src/main/cpp/DeviceDelegate.h
MortimerGoro: It was there for some initial API test, but I moved everything to VRLayer.h. It's not needed anymore.

app/src/main/cpp/DeviceDelegate.h
MortimerGoro: It was there for some initial API test, but I moved everything to VRLayer.h. It's not needed anymore.

app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -908,14 +929,15 @@ void
 BrowserWorld::DrawWorld() {
   m.externalVR->SetCompositorEnabled(true);
   m.device->SetRenderMode(device::RenderMode::StandAlone);
+  if (m.fadeAnimation) {
+    m.fadeAnimation->UpdateAnimation();
```
bluemarvin: m.fadeAnimation seems like a good candidate for vrb::Updatable, that way you can use RenderContext::GetFrameDelta() to make the animations time based. If you agree, just file as a follow up.

app/src/main/cpp/Device.h
```diff
@@ -38,6 +38,11 @@ struct EyeRect {
     mHeight = aRect.mHeight;
     return* this;
   }
+
+  bool IsDefault() const {
```
bluemarvin: This function doesn't seem to be used anywhere?

app/src/main/cpp/Quad.cpp
```diff
@@ -236,6 +259,18 @@ Quad::SetBackgroundColor(const vrb::Color& aColor) {
...
+    /*
+     * We disable the background when using layers because there is a difficult to fix glitch.
+     * We can create 2 nodes to simulate the background color on the sides of the ScaleToFit mode.
+     * It works while resizing, but at the end, when the actual GV surface is really changed
+     * there is some SurfaceTexture update delay that causes a glitch. Background color becomes
+     * transparent or the background color is shown on top of the GV quad surface.
+     * This glitch doesn't happen with normal scenegraph code because the background color is shown
+     * on the back of the Quad surface. But with Layers we can't do that.
+     */
+    return;
```
bluemarvin: I'm not sure I understand the issue but it sounds like we were drawing a blank quad while resizing. I'm surprised this didn't cover the layer since the content layer is drawn first? Feel free to file a follow up if there is still an out standing issue.

app/src/main/cpp/VRLayer.h
```diff
@@ -0,0 +1,149 @@
...
+#include "vrb/gl.h"
+#include "vrb/MacroUtils.h"
+#include "vrb/Forward.h"
+
+#include "Device.h"
+
+#include <memory>
+#include <functional>
+#include <jni.h>
+
```
bluemarvin: Looks like functional and jni.h are left overs here to? I didn't see then being used any where.

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -23,14 +25,38 @@
...
 #include "VrApi.h"
 #include "VrApi_Helpers.h"
 #include "VrApi_Input.h"
 #include "VrApi_SystemUtils.h"
 
+#include "VRBrowser.h"
+
 namespace crow {
 
+static ovrMatrix4f ovrMatrixFrom(const vrb::Matrix& aMatrix) {
```
bluemarvin: This looks like a good candidate for a member function in vrb::Matrix. Something like `CopyTransposed(float[4][4] aTarget)`? As a follow up if you agree.

caseyyee:
Woohoo!  Thanks guys :)

PR #788, opened 2018-11-20
Refactor controller pointer code. Support layers for pointer.

MortimerGoro:
- Refactor controller pointer code
- Improve quality of the pointer when layers are enabled
- Fixed depth problem when the pointer is shown on a widget that is back to another widget (e.g. dialog)

### DIFF
--- a/app/src/main/cpp/BrowserWorld.cpp
+++ b/app/src/main/cpp/BrowserWorld.cpp
@@ -15,6 +15,7 @@
 #include "LoadingAnimation.h"
 #include "Skybox.h"
 #include "SplashAnimation.h"
+#include "Pointer.h"
 #include "Widget.h"
 #include "WidgetPlacement.h"
 #include "Quad.h"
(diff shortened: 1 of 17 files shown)


app/src/main/cpp/Controller.h
```diff
@@ -32,6 +35,7 @@ struct Controller {
   float scrollDeltaX;
   float scrollDeltaY;
   vrb::TransformPtr transform;
+  PointerPtr pointer;
```
bluemarvin: Need to add this to `operator=` in Controller.cpp

app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -290,12 +290,20 @@ BrowserWorld::State::UpdateControllers(bool& aRelayoutWidgets) {
       resizingWidget.reset();
     }
 
+    if (controller.pointer) {
+      controller.pointer->SetVisible(hitWidget.get() != nullptr);
+      controller.pointer->SetHitWidget(hitWidget);
+    }
+    if (hitWidget) {
+      vrb::Matrix translate = vrb::Matrix::Translation(vrb::Vector(hitPoint.x(), hitPoint.y(), 0.001f));
+      controller.pointer->SetTransform(hitWidget->GetTransform().PostMultiply(translate));
```
bluemarvin: This `if (hitWidget) {` needs to be in the `if(controller.pointer) {` block since it calls `pointer->SetTransform()`

PR #790, opened 2018-11-20
Add support for UI texture scaling

MortimerGoro:
We are using pretty big textures now for UI due to the Android screen density. VR doesn't have the same density so we may be using more than needed. This PR allows to set a specific texture scaling per widget. The "big" scaling is done on the UI renderer only after each render due to layout changes instead of every frame on the GPU

I set a default value of 0.75f. There may be some better formula taking z into account but it's a starting point. I saw some improvements on the aliasing of the navigation bar on Oculus with Layers enabled (In addition to reduced texture sizes). It should have a similar aliasing effects to creating mipmaps if we are able to set the right value here.

I'd like some more testing on all the platforms. Thoughts?

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
@@ -474,8 +474,10 @@ public class VRBrowserActivity extends PlatformActivity implements WidgetManager
             }
             widget.setSurfaceTexture(aTexture, aWidth, aHeight);
             // Add widget to a virtual display for invalidation
-            if (((View)widget).getParent() == null) {
-                mWidgetContainer.addView((View) widget, new FrameLayout.LayoutParams(aWidth, aHeight));
+            View view = (View) widget;
+            if (view.getParent() == null) {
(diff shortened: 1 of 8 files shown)


keianhzo:
@MortimerGoro I'm seeing an artifact if you turn off/on the headset, it seems like the texture scaling is not correct after scaling.

See the gaps between the tray buttons below.

scaling2

MortimerGoro:
Yes, I reproduced the gap in the tray.

I'll investigate this later. There is no rush to land this for v1.1

bluemarvin:
@MortimerGoro This might be a good one for 1.1.1 release?

MortimerGoro:
I still need to research some of vertical gap issues. Probably not in time for v1.1.1

MortimerGoro:
I fixed the gaps on the tray and rebased all the code. This is ready for review again

bluemarvin:
On fresh start, the URL bar is clipped on Vive Focus.
screen_shot_1902142708

bluemarvin:
I'm not longer able to reproduce the issue. It was 100% reproducible before on Focus. I think there have been several system updates so maybe that fixed something.

PR #798, opened 2018-11-21
Fix stuck in fullscreen mode when exiting private browsing. Fixes #739

### LINKED ISSUE #739: Stuck in fullscreen mode when exiting private browsing
## Hardware
Oculus go

## Steps to Reproduce
1. Enter private browsing mode
2. Navigate to a page with Fullscreen (eg: Youtube, watch video fullscreen)
3. Press the ( X ) icon above browser window to close private browsing mode.

## Current Behavior
When returning back to regular browser, we're stuck in fullscreen mode. 
Note that URL bar is present when navigating to another page, though no other controls are.

## Expected Behavior
Exit out of full-screen mode, returns user back to regular browsing.

See video here:
https://imgur.com/a/jjpxkyF

Screenshot:
scrcpy_2018-11-08_15-04-49

MortimerGoro:
Fixes #739 

I also found that the `X` button of private mode was only visible after a browser window change. The worldWidth was set to `Infinity` and the 'X' widget got more size than needed on the browser widget updat call. Fixed this issues too in this PR


cvan:
@MortimerGoro: on `master`, I don't see any more `(×)` Close button at the top of the window anymore. if that's true, in which commit did it go away? or am I missing something?

PR #808, opened 2018-11-21
Fixes #804 Update strings for Japanese v1.1 version

### LINKED ISSUE #804: [l10n] Update strings for Japanese v1.1 version
Japanese localizers have updated the localized strings for v1.1.
Please[ find attached the files ](https://drive.google.com/drive/u/1/folders/1y6DoKigRSElVY0il4kIPN4fuqEheiDRh)that we'll need to use instead of current versions of both localpages.xml and strings.xml
thanks!
(CC @nancyhang and @larsbergstrom so they are aware)

keianhzo:
Fixes #804 Update strings for Japanese v1.1 version

@delphine @Pike for review

### DIFF
--- a/app/src/main/res/values-ja/FirefoxReality_localpages.xml
+++ b/app/src/main/res/values-ja/FirefoxReality_localpages.xml
@@ -12,7 +12,7 @@
     <!-- This string is used as the body of the Private Browsing page that is opened when the
     Private Browsing Mode button from the tray is clicked. -->
     <string name="private_browsing_body"><![CDATA[
-<p>%1$0.sプライベートウィンドウでは以下を<strong>保存しません</strong>:</p>
+<p>%1$sプライベートウィンドウでは以下を<strong>保存しません</strong>:</p>
 <ul class="two-column">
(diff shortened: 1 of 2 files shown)


app/src/main/res/values-ja/FirefoxReality_localpages.xml
```diff
@@ -12,15 +12,15 @@
     <!-- This string is used as the body of the Private Browsing page that is opened when the
     Private Browsing Mode button from the tray is clicked. -->
     <string name="private_browsing_body"><![CDATA[
-<p>プライベートブラウジング中、%1$s は以下を<strong>保存しません</strong>:</p>
+<p>%1$0.sプライベートウィンドウでは以下を<strong>保存しません</strong>:</p>
```
bluemarvin: Looks like `%1$s` got changed to `%1$0.s`? I thought we were only supporting `%1$s`?

Delphine: Yeah, it should be reverted, that's probably a typo

Issue #751, opened 2018-11-13
UI sounds are now too loud


cvan:
## Hardware
Oculus Go
[`1.0.1`](https://github.com/MozillaReality/FirefoxReality/tree/1.0.1), eaa4fc5

## Steps to Reproduce
1. Exit Firefox Reality.
2. Press the physical Volume-Up button until the volume is at the maximum level.
3. Press the buttons in the browser chrome/UI.

## Current Behavior
Notice the noises are too loud.

## Expected Behavior
The noises should be ~70% quieter.

caseyyee:
Other sound inconsistencies found:
- UI sounds do not play when using the hardware back button on controller.
- Sounds not present with media controls.

UI sounds overall should be much more subtle/quite and it would be really nice to have the option to disable them altogether.  Another idea would be to use haptics instead of sound to provide UI feedback.

cvan:
I have a local branch of removing all logic for these audio files. To simplify things and reduce future tech debt, I will simply prevent the loading of the sounds but keep the listeners as they are. And we can polish the sounds and interactions in later releases. (Haptics would be nice, but there are no haptics in the Oculus Go controllers.)

cvan:
done in PR #813 (merged)

PR #813, opened 2018-11-22
temporarily disable loud sounds for now (fixes issue #751)

cvan:
let me know if there is a preferred way of disabling (commenting out) these sounds until we approach a better user response (possibly involving haptics and so on).


MortimerGoro:
I'd add a `setEnabled(boolean aEnabled)` method to audio engine. That way we can also use it to include a settings toggle in the future.

keianhzo:
Fixed merge issues and added and enable method and a setting.

PR #814, opened 2018-11-22
Improve layer blending. Fix beam black halo.

MortimerGoro:


MortimerGoro:
@bluemarvin I'm merging this. You can add follow-ups if you find something

PR #815, opened 2018-11-22
Fix black quad on the browser window before the first composite.

MortimerGoro:


MortimerGoro:
@bluemarvin I'm merging this. You can add follow-ups if you find something.

Issue #742, opened 2018-11-09
Pressing clear in the URL bar closes the keyboard


cvan:
## Hardware
Oculus Go
7c8b949

## Steps to Reproduce
1. Enter a URL, such as http://example.com
2. Click the × icon in the far-right corner of the URL bar
3. Notice the keyboard unexpectedly gets dismissed

## Current Behavior
The keyboard unexpectedly gets dismissed:

![fxr-clear-url-2](https://user-images.githubusercontent.com/203725/48245904-203ee500-e3a2-11e8-8d1a-b9ebdb710965.gif)

## Expected Behavior

The keyboard stays open.

cvan:
handled in PR #830

PR #830, opened 2018-11-26
keep keyboard open upon pressing Clear icon in URL bar (fixes issue #742)

cvan:
so I can better understand how my fix fixes the bug, can you share the context of PR #609?

specifically, in the `mMicrophoneListener` method, what does the `view.requestFocusFromTouch()` call do?

the reason I ask: on `master`, I was able to reproduce the same bug randomly twice (but I still cannot find reliable STR): I pressed the Voice Search icon in the URL bar, my Voice Search was processed and completed, but then the keyboard opened (even though I was not previously interacting with the keyboard, and the keyboard was closed). ideas? if it was a bug/oversight from PR #609, I can address this in this PR. otherwise, I'll be filing an issue.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/views/NavigationURLBar.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/views/NavigationURLBar.java
@@ -419,7 +419,6 @@ public class NavigationURLBar extends FrameLayout {
             mAudio.playSound(AudioEngine.Sound.CLICK);
         }
 
-        view.requestFocusFromTouch();
         mURL.getText().clear();
     };
 


keianhzo:
@cvan By default android only triggers focus change events for some widgets that consume user input, like EditText for example. In our context we need to handle focus change events in other cases like:

1. User opens the settings dialog
2. User clicks on the URL bar back button

By default Android won't deliver focus change events in that case but we need to get them so we know the settings widget should be closed. So that's what we do with **requestFocusFromTouch**: trigger focus change events for those widgets that don't trigger them by default.

All the focus change events are handled at https://github.com/MozillaReality/FirefoxReality/blob/a9da13e7a43c7947ed50b16566df8aba483ad9ff/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java#L170 

And sent to all the widget that are subscribed to the focus change events.

Issue #802, opened 2018-11-21
YouTube 360 not displaying properly using VR video modes


caseyyee:
## Hardware
Go FXR: 453bd4a

## Steps to Reproduce
1. Navigate to 360 Youtube video eg (https://www.youtube.com/watch?v=qUxMdYhipvQ)
2. Enter fullscreen
3. Select 360 projection 

## Current Behavior
Video is present, but Youtube 2D UI also comes along for the ride.

## Expected Behavior
Properly formatted 360 presentation.

![yt-fullscreen-360](https://user-images.githubusercontent.com/1560884/48864099-602da100-ed80-11e8-9fb9-932f864f5412.gif)

caseyyee:
Appears to be fixed.

PR #831, opened 2018-11-26
Fix duplicated surface changed calls with layers

MortimerGoro:


MortimerGoro:
Helps with https://github.com/MozillaReality/FirefoxReality/issues/802

MortimerGoro:
I'm merging this so it's available the final apk for v1.1 testing.

PR #853, opened 2018-11-27
Improve VR video quality on platforms not using layers

MortimerGoro:


MortimerGoro:
Fixes #754

bluemarvin:
The taskcluster failure was not a build failure but a VM failure. I re-triggered and it passed. Merging.

PR #856, opened 2018-11-27
Don't remember fullscreen resize on restart

### LINKED ISSUE #839: Browser window size is set to fullscreen size on restart.
## Hardware

Focus

## Steps to Reproduce

1. Make video fullscreen.
2. Manually resize the window with the drag handles.
3. Press home button to exit the app.
4. Relaunch the application

## Current Behavior

Browser window is now using the fullscreen window size.

## Expected Behavior

Browser window should still have the same size from before exiting the application.

MortimerGoro:
Fixes  #839

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/BrowserWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/BrowserWidget.java
@@ -47,6 +47,7 @@ public class BrowserWidget extends UIWidget implements SessionStore.SessionChang
     private int mBorderWidth;
     Runnable mFirstDrawCallback;
     private boolean mIsInVRVideoMode;
+    private boolean mSaveResizeChanges;
 
     public BrowserWidget(Context aContext, int aSessionId) {
         super(aContext);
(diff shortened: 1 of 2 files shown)


bluemarvin:
Shouldn't both browser window size and fullscreen window size be saved seperately? Maybe just a scale factor for fullscreen since we resize based on video aspect ratio? Probably should be considered when the window resize refactor work is done.

cvan:
> Shouldn't both browser window size and fullscreen window size be saved seperately? Maybe just a scale factor for fullscreen since we resize based on video aspect ratio? Probably should be considered when the window resize refactor work is done.

@caseyyee ^ thoughts?

caseyyee:
I like the option of scaling the fullscreen window to any ratio. It's not just for video, even if that's the main use case.  As far as remembering the size goes, I think some proportion of the browser window size might work, but I can see this being a annoying inconsistency from a UX perspective as well.  Safest bet for now would be just to have a fixed starting size for fullscreen.  It's how it works now, and I think it's pretty alright.

bluemarvin:
I'm pretty sure we are resizing the window based on the videos aspect ratio to prevent black bars. Am I wrong? I think fullscreen video resize should maintain the aspect ratio. Otherwise we are wasting pixels. Which would make it different from a normal window resize.

PR #871, opened 2018-11-29
Fixes #671 Tray doesn't indicate private icon on when restoring

### LINKED ISSUE #671: Tray doesn't indicate private icon on when restoring a private mode session
Also the background is not dimmed. It seems that the tray didn't receive the listener in that case.

## Steps to Reproduce
1. Open FxR and switch to private mode
2. Click back and exit the browser
3. Reopen it again from the Oculus Library

## Current Behavior
- Private session correctly restored
- Tray icon not enabled
- Background not dimmed

## Expected Behavior
- Tray private mode icon enabled
- Background dimmed

keianhzo:
Fixes #671 Tray doesn't indicate private icon on when restoring


bluemarvin:
@cvan Does this patch cause these other issues or are they preexisting. If they are preexisting and this patch fixes the issue we should land it and file issues for the other bugs you have found.

cvan:
@bluemarvin it seems to regress, introduce bugs that did not appear before

PR #873, opened 2018-11-29
Oculus touch controller

daoshengmu:
In order to support all the current functions in FxR, we need to follow MozillaReality/FirefoxReality-android-third-party#2 to upgrade SDK to 1.19. In this implementation, I notice the projection matrix from `predictedTracking.eye` is fovX: 95, foxY: 99, But if we use `VRAPI_SYS_PROP_SUGGESTED_EYE_FOV_DEGREES_X` to get fov, that will be 104 and 106 degree. So, we have to make it consistent between camera.projectionMatrix and the main eye buffer layer. 

Besides, the controllerTransform is wrong when doing 6 DOF orientation. It should rotate transform by the elbow instead of by the origin. This controller has the same button layout as *Oculus Rift Touch* but different looking. So, we will keep working on it at following bugs to let it support multiple controllers and has the same button and axis mapping for WebVR content as *Oculus Rift*.


app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -642,45 +644,59 @@ struct DeviceDelegateOculusVR::State {
...
     }
 
-    if (controllerCapabilities.ControllerCapabilities & ovrControllerCaps_HasPositionTracking) {
-      auto & position = tracking.HeadPose.Pose.Position;
-      controllerTransform.TranslateInPlace(vrb::Vector(position.x, position.y, position.z));
-    } else {
-      controllerTransform = elbow->GetTransform(hand, head, controllerTransform);
-    }
-
+    controllerTransform = elbow->GetTransform(hand, head, controllerTransform);
```
bluemarvin: 6Dof controllers should not require an elbow model.

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -642,45 +644,59 @@ struct DeviceDelegateOculusVR::State {
...
-
-    const float trackpadX = controllerState.TrackpadPosition.x / (float)controllerCapabilities.TrackpadMaxX;
-    const float trackpadY = controllerState.TrackpadPosition.y / (float)controllerCapabilities.TrackpadMaxY;
-    float scrollX = trackpadX;
-    float scrollY = trackpadY;
-    if (trackpadTouched && !trackpadPressed) {
-      controller->SetTouchPosition(0, scrollX, scrollY);
+    const int32_t kNumAxes = 2;
+    bool triggerPressed, triggerTouched = false;
+    bool trackpadPressed, trackpadTouched = false;
```
bluemarvin: `triggerPressed` and `trackpadPressed` both need to be initialized.

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
daoshengmu: got it. it works after add `kAverageHeight` as we give to the headset position.

daoshengmu:
@bluemarvin  the 6DoF head tracking is supported from my patch. Because we use `m.predictedTracking.HeadPose.Pose` to get the headset position that is 6Dof data. Both of 6Dof headset and controllers are workable base on this.

In terms of two controllers, I already have a WIP patch in local, it has been near to be done. I will send you a PR shortly.

bluemarvin:
@daoshengmu Please wait for @MortimerGoro to review before landing. He wrote the original implementation so I would like his feedback first. Thanks.

daoshengmu:
Update using `deviceType` to check the controller type in `DeviceDelegateOculusVR::SetControllerDelegate()` because `controllerCapabilities` is not yet ready until calling `UpdateControllerID()`.

Issue #860, opened 2018-11-27
Keyboard stays open after entering a URL


cvan:
## Hardware
Oculus Go
9cf66ae

## Steps to Reproduce

1. Press the URL bar
2. Enter an address
3. Notice the URL bar stays open even after the page is loaded

## Current Behavior
The URL bar remains upon after a URL is entered / a page is requested.

## Expected Behavior
The URL bar closes after a URL is entered / a page is requested.

## Context
Not sure if this is a regression I added in PR #830. I haven't dug deep enough yet.

keianhzo:
Probably related or duplicate of #846

cvan:
Yep, I notice it after a crash prompt, search suggestions, and a few others. Probably the same fix for all?

keianhzo:
Fixed with ##889 Closing

Issue #849, opened 2018-11-27
Keyboard appears upon a crash


cvan:
## Hardware

Oculus Go
f5e522d

## Steps to Reproduce

1. Load the URL `about:crashparent`
2. Notice upon relaunch, the keyboard appears opened

## Context
Related: issue #846

## Error Logs and Stack Traces

<details open>

```
```

</details>

keianhzo:
Probably related or duplicate of #846

keianhzo:
Fixed with #889 Closing.

PR #889, opened 2018-12-14
Fixes #846 #849 #860 Fixes keyboard focus issues

### LINKED ISSUE #846: Keyboard should not open upon a successful Voice Search
## Hardware

Oculus Go
f5e522d

## Steps to Reproduce

1. Press the Voice Search button (the microphone icon) in the URL bar
2. Say a phrase to trigger a Voice Search
3. Notice after the results are processed and the search results page loads, the keyboard opens

## Current Behavior
The keyboard opens after a Voice Search.

## Expected Behavior
The keyboard does not open after a Voice Search.

keianhzo:
Fixes #846 #849 #860 Avoid URL EditText to gain focus when focus changes

### DIFF
--- a/app/src/main/res/layout/navigation_url.xml
+++ b/app/src/main/res/layout/navigation_url.xml
@@ -17,7 +17,9 @@
 
     <FrameLayout
         android:layout_width="match_parent"
-        android:layout_height="match_parent">
+        android:layout_height="match_parent"
+        android:descendantFocusability="beforeDescendants"
+        android:focusableInTouchMode="true">
 
         <mozilla.components.ui.autocomplete.InlineAutocompleteEditText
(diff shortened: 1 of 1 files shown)


keianhzo:
@cvan I've been testing that and it happens but not 100% it seems that under some circumstances GV doesn't call showSoftInput even though the browser view is focused. I don't think is related to this issue and might be GV related so best thing is to open a separate issue.

cvan:
> @cvan I've been testing that and it happens but not 100% it seems that under some circumstances GV doesn't call showSoftInput even though the browser view is focused. I don't think is related to this issue and might be GV related so best thing is to open a separate issue.

filed issue #898 to track this. feel free to add your findings to that issue. thank you.

PR #892, opened 2018-12-14
Enable ETC2 compressed textures. Add asset compressor tool.

MortimerGoro:
…w assets


MortimerGoro:
I need to also convert the controller model images, just forgot they are in different folders, I'll update the PR, but you can start the review already

MortimerGoro:
Already converted controller images

app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -505,14 +505,16 @@ BrowserWorld::InitializeJava(JNIEnv* aEnv, jobject& aActivity, jobject& aAssetMa
     m.loadingAnimation->LoadModels(m.loader);
     m.rootController->AddNode(m.controllers->GetRoot());
     std::string skyboxPath = VRBrowser::GetActiveEnvironment();
+    std::string extension;
     if (VRBrowser::isOverrideEnvPathEnabled()) {
       std::string storagePath = VRBrowser::GetStorageAbsolutePath(INJECT_SKYBOX_PATH);
       if (std::ifstream(storagePath)) {
         skyboxPath = storagePath;
+        extension = ".jpg";
```
bluemarvin: Would be great to either scan or allow user to specify extension. Definitely a followup, I believe there is already an issue on this.

bluemarvin:
Broken on Focus too.

bluemarvin:
Oops, forgot to update VRB, testing again 😔

PR #894, opened 2018-12-14
Fixes #887 Merge localpages.xml and strings.xml

### LINKED ISSUE #887: [l10n] Merge localpages.xml and strings.xml
It would make sense to merge localpages.xml and strings.xml in the same file at this point, like we do for other mobile projects (so, just one file containing all the strings). In fact, there are only two strings left in the localpages.xml file at this point.
Thoughts?
(CC @Pike so he's aware I filed this)

keianhzo:
Fixes #887 Merge localpages.xml and strings.xml


l10n.toml
```diff
@@ -16,8 +16,4 @@ locales = [
 
 [[paths]]
   reference = "app/src/main/res/values/strings.xml"
-  l10n = "app/src/main/res/values-{android_locale}/strings.xml"
-  
-  [[paths]]
-    reference = "app/src/main/res/values/localpages_strings.xml"
-    l10n = "app/src/main/res/values-{android_locale}/localpages_strings.xml"
+  l10n = "app/src/main/res/values-{android_locale}/strings.xml"
```
Pike: Keep the trailing newline in this file?

keianhzo: Fixed

PR #900, opened 2018-12-18
Fix resize mode visible on top of settings

### LINKED ISSUE #895: Dim opacity more when Settings is opened
## Hardware
Oculus Go
1.1, 12eac9b, ef19a34

## Steps to Reproduce

1. Open Bookmarks (from the Tray)
2. Enter Resize Mode (from the far-right icon below the browser window)
3. Use the handlers to resize the window
4. Enter Settings (from the Tray)
5. Notice the opacity of the Settings lets the Resize Mode UI shine through

## Current Behavior
Notice the opacity of the Settings lets the Resize Mode UI shine through.

image

## Expected Behavior
The opacity ought to be dark enough when the Settings is opened that the Resize Mode UI doesn't shine through.

## Context

See these related issues:
- #861
- **https://github.com/MozillaReality/FirefoxReality/issues/671#issuecomment-440494471**
- #774
- #389

MortimerGoro:
Fixes  #895


bluemarvin:
I'm pretty sure this will happen with any dialog and not just the settings. However I'm not actually sure that you can be in resize mode and have another dialog come up. Maybe a better solution is to hide the tray while in resize mode?

keianhzo:
+1 hiding the tray will avoid any conflicts plus we are already hiding it for fullscreen mode.

MortimerGoro:
Done! Ready for review

PR #905, opened 2018-12-20
Fixes #435 Added support for alert/prompt/confirm

### LINKED ISSUE #435: Implement UI for browser-level `alert`/`prompt`/`confirm`
Hardware: All
Steps to Reproduce:

    Go to: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select
    Scroll to Examples.
    Click on selection drop down list.

Actual Behavior:
Nothing happens.

Expected Behavior:
A selection list is presented and the user may select an item in the list.

Recommendation:
FxR need to implement UI for the various selection types a web page may present.

keianhzo:
Fixes #435 Added support for alert/text/confirm prompt dialogs. Prompts are now dynamically resizable depending on the content size with a max height limit.

@MortimerGoro I've noticed a glitch when resizing the widget I guess because the size is updated while the widget is still visible. I didn't have time to dig into that so feel free to update the PR or open a follow-up to fix that.


MortimerGoro:
@keianhzo Check this commit https://github.com/MozillaReality/FirefoxReality/pull/905/commits/2fe9349370d47afab34a5e906ecd1b9ff9ad64be

MortimerGoro:
You can test with https://webxr.sh/dialogues.html

bluemarvin:
Just merge and I’ll test the staging build.

bluemarvin:
Looks good to me.

bluemarvin:
@cvan if you find additional issues file followup.

PR #906, opened 2018-12-20
Exit fullscreen on VR video exit when autoenter is used

MortimerGoro:


bluemarvin:
@cvan @caseyyee is this ready to land?

PR #908, opened 2018-12-20
Fixes #618 Support debug and release builds on device simultaneously

### LINKED ISSUE #618: Support debug and release builds on device simultaneously
I would find this useful when I want to quickly compare against the released version of FxR and I'm working on hacked up and customized development builds locally. It would also mean that I would not have to keep reinstalling the app every time I want to use it for real.

keianhzo:
Fixes #618 Added support for simultaneous debug and release builds int he same device.


bluemarvin:
Will this work correctly on Oculus? I assume when we install the debug version you will have to launch it from adb as the debug version won't be visible in the launcher?

cvan:
> One question, why do we need to duplicate the logos?

to easily differentiate between release and debug versions, if I had to guess.

I was going to create a version with a `Debug` ribbon on the edge:

image

but the grayscale version's good IMO 👍

keianhzo:
@MortimerGoro I change the logo just so it's more obvious that it's a debug build but that can be removed or changed for any other logo.

@rbarker it does work ok Oculus, just the store app will show in the apps shelf and the debug app will show in the unknown sources list. I couldn't find a way for the debug build to show in the apps shelf.

bluemarvin:
I would prefer this be a `user.properties` option and not the default behavior at least for oculus but really for all platforms. Having two different FxR on a head set really complicates testing and development scripts.

MortimerGoro:
+1 for the `user.properties` option.

I updated this PR with a new user property to enable the simultaneous builds

MortimerGoro:
Merging this. The dev sufix is only enabled via user.properties now so it will be harmless for tascluster

PR #917, opened 2019-01-08
Fix broken WindowWidget views when not using layers. Fixes #911

### LINKED ISSUE #911: Bookmarks broken on lenovo mirage & HTC Focus
## Hardware
Reproduced on HTC Focus; reported on Lenovo Mirage
Version 1.1.1 in-stores.

## Steps to Reproduce
1. Visit website
2. Click Star to bookmark
3. View bookmarks page

## Current Behavior
No bookmark added

## Expected Behavior
Bookmark added

MortimerGoro:
Fixes #911

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WindowWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WindowWidget.java
@@ -144,10 +144,17 @@ public class WindowWidget extends UIWidget implements SessionStore.SessionChange
         mView.setVisibility(VISIBLE);
         addView(mView);
         mWidgetPlacement.density =  getContext().getResources().getDisplayMetrics().density;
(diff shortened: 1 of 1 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WindowWidget.java
```diff
@@ -642,9 +653,8 @@ public void onBookmarksHidden() {
     @Override
     public void onBookmarksClicked() {
         if (mBookmarksView.getVisibility() == View.VISIBLE) {
-            SessionStore.get().unstackSession();
-
             unsetView(mBookmarksView);
+            SessionStore.get().unstackSession();
```
bluemarvin: I realize this is existing code from the feature, but why do we need to push an empty session onto the stack to show book marks? It seems that this session is never actually used.

PR #921, opened 2019-01-10
Use proper browserState

paulrouget:
Once this lands: https://bugzilla.mozilla.org/show_bug.cgi?id=1519035 we will need FxR to switch between `servoState` and `geckoState`.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WidgetManagerDelegate.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WidgetManagerDelegate.java
@@ -38,6 +38,7 @@ public interface WidgetManagerDelegate {
     void setTrayVisible(boolean visible);
     void setControllersVisible(boolean visible);
     void setWindowSize(float targetWidth, float targetHeight);
+    void setIsServoSession(boolean aIsServo);
     void keyboardDismissed();
     void updateEnvironment();
     void updatePointerColor();
--- a/app/src/main/cpp/ExternalVR.cpp
+++ b/app/src/main/cpp/ExternalVR.cpp
@@ -121,6 +121,9 @@ namespace crow {
 
 struct ExternalVR::State {
   static ExternalVR::State * sState;
+  pthread_mutex_t* browserMutex;
+  pthread_cond_t* browserCond;
+  mozilla::gfx::VRBrowserState* sourceBrowserState;
   mozilla::gfx::VRExternalShmem data;
   mozilla::gfx::VRSystemState system;
   mozilla::gfx::VRBrowserState browser;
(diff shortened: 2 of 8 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WidgetManagerDelegate.java
```diff
@@ -32,6 +32,7 @@
     void removeUpdateListener(@NonNull UpdateListener aUpdateListener);
     void pushBackHandler(@NonNull Runnable aRunnable);
     void popBackHandler(@NonNull Runnable aRunnable);
+    void pushIsServo(boolean aIsServo);
```
MortimerGoro: I'd rename this to something else like `setIsServo` or `setIsServoSession`. Using `push` prefix normally indicates that there is also a `pop`.

paulrouget:
In `sync_poses`, I block, but I do not copy the state. That means when `synced_frame_data` and `inmediate_frame_data` are called, I use the system mutex before reading the shared memory.

Is that fine, or should I copy the state in `sync_poses` and use that copy in `synced_frame_data` and `inmediate_frame_data`?

app/src/main/cpp/ExternalVR.cpp
```diff
@@ -172,8 +178,12 @@ struct ExternalVR::State {
 
   void PullBrowserStateWhileLocked() {
     const bool wasPresenting = IsPresenting();
-    memcpy(&browser, &data.browserState, sizeof(mozilla::gfx::VRBrowserState));
 
+    if (sourceBrowser == ExternalVR::VRBrowserType::Gecko) {
```
MortimerGoro: The code may be better if we create something like `browserMutex` and `browserCond` references in `ExternalVR::State` and set them in `SetSourceBrowser`. That way we can avoid the ifs in all the Pull/push methods and just use the current ones from the state struct.

app/src/main/cpp/ExternalVR.cpp
```diff
@@ -283,12 +293,24 @@ ExternalVR::PushSystemState() {
 
 void
 ExternalVR::PullBrowserState() {
-  Lock lock(m.data.browserMutex);
+
+  pthread_mutex_t* browserMutex;
+  if (m.sourceBrowser == ExternalVR::VRBrowserType::Gecko) {
```
MortimerGoro: Same comment as above

app/src/main/cpp/ExternalVR.cpp
```diff
@@ -379,9 +401,18 @@ ExternalVR::PushFramePoses(const vrb::Matrix& aHeadTransform, const std::vector<
 
 bool
 ExternalVR::WaitFrameResult() {
-  Wait wait(m.data.browserMutex, m.data.browserCond);
+  pthread_mutex_t* browserMutex;
+  pthread_cond_t* browserCond;
+  if (m.sourceBrowser == ExternalVR::VRBrowserType::Gecko) {
```
MortimerGoro: Same comment about these ifs

app/src/main/cpp/ExternalVR.cpp
```diff
@@ -283,12 +293,24 @@ ExternalVR::PushSystemState() {
...
+  }
+  Lock lock(*browserMutex);
   if (lock.IsLocked()) {
    m.PullBrowserStateWhileLocked();
   }
 }
 
+void ExternalVR::SetSourceBrowser(VRBrowserType aBrowser) {
+  m.Reset();
+  m.sourceBrowser = aBrowser;
```
MortimerGoro: Do the if here and set the mutex & cond references according to the browser type

app/src/main/cpp/ExternalVR.cpp
```diff
@@ -133,16 +134,20 @@ struct ExternalVR::State {
...
+    pthread_cond_init(&data.geckoCond, nullptr);
+    pthread_cond_init(&data.servoCond, nullptr);
   }
 
   ~State() {
     pthread_mutex_destroy(&(data.systemMutex));
-    pthread_mutex_destroy(&(data.browserMutex));
+    pthread_mutex_destroy(&(data.geckoMutex));
+    pthread_mutex_destroy(&(data.servoMutex));
     pthread_cond_destroy(&(data.systemCond));
```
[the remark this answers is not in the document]
paulrouget: I don't now. @kearwood ^

paulrouget:
@MortimerGoro I've updated the PR. I don't think that we need to duplicate systemCond, but still waiting from @kearwood confirmation.

paulrouget:
This change will happen in #936

PR #928, opened 2019-01-18
Set ExternalVR timestamp

MortimerGoro:


app/src/main/cpp/ExternalVR.cpp
```diff
@@ -404,6 +404,10 @@ ExternalVR::PushFramePoses(const vrb::Matrix& aHeadTransform, const std::vector<
     }
   }
 
+  timespec spec;
+  clock_gettime(CLOCK_MONOTONIC, &spec);
+  m.system.sensorState.timestamp = (double)spec.tv_sec + (double)spec.tv_nsec / 1e09f;
```
bluemarvin: Why not just use `RenderContext::getTimeStamp()`?

https://github.com/MozillaReality/vrb/blob/master/include/vrb/RenderContext.h#L29

MortimerGoro:
@bluemarvin can we get this one merged? it should be ready

PR #930, opened 2019-01-24
Add support for curved display

MortimerGoro:
- Curved display is disabled by default. It can be enabled via display settings (I'll create another PR to avoid the required restart)
- Uses Oculus Layers or a geometry fallback if layers are not available
- Resize mode adapted for layers (minor extra width glitch at the bottom resize line, will fix it in a different commit/PR)

Requires https://github.com/MozillaReality/vrb/pull/58 and https://github.com/MozillaReality/vrb/pull/59


MortimerGoro:
Merging this. I'll address the no-restart and the glitch at the bottom resize line on a different PR)

PR #944, opened 2019-02-12
Fix queue event lifecycle on googlevr and noapi platforms

MortimerGoro:
GLSurfaceView.queueEvent() calls dispatched before the surfaceView renderer is ready are dispatched before the activityCreated callback is notified to native. This can cause some nullptr device states (we had to add a if not null workaround) or issues when initializing some data (e.g curdved setting on the first widgets created)

### DIFF
--- a/app/src/googlevr/java/org/mozilla/vrbrowser/PlatformActivity.java
+++ b/app/src/googlevr/java/org/mozilla/vrbrowser/PlatformActivity.java
@@ -17,6 +17,8 @@ import android.view.WindowManager;
 import com.google.vr.ndk.base.AndroidCompat;
 import com.google.vr.ndk.base.GvrLayout;
 
+import java.util.ArrayList;
+
 import javax.microedition.khronos.egl.EGLConfig;
 import javax.microedition.khronos.opengles.GL10;
 
@@ -30,6 +32,8 @@ public class PlatformActivity extends Activity {
     private GvrLayout mLayout;
     private GLSurfaceView mView;
     private static final String FLAT_ACTIVITY_CLASSNAME = "org.mozilla.vrbrowser.BrowserActivity";
+    private ArrayList<Runnable> mPendingEvents;
+    private boolean mSurfaceCreated = false;
 
     private final Runnable activityDestroyedRunnable = new Runnable() {
         @Override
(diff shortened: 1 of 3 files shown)


app/src/googlevr/java/org/mozilla/vrbrowser/PlatformActivity.java
```diff
@@ -170,8 +177,27 @@ void setImmersiveSticky() {
                                 | View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY);
     }
 
+
     void queueRunnable(Runnable aRunnable) {
-        mView.queueEvent(aRunnable);
+        if (mSurfaceCreated) {
+            mView.queueEvent(aRunnable);
+        } else {
+            synchronized (mPendingEvents) {
```
bluemarvin: Are events getting queued off of the main UI thread?

MortimerGoro: `queueRunnable` is always called from the main thread. But the pendingEvents array is iterated also when the surfaceCreated callback is received to notify the pending events

MortimerGoro:
I did some tests removing the lock and adding this to `onSurfaceCreated`:
```
PlatformActivity.this.runOnUiThread(() -> {
  mSurfaceCreated = true;
  notifyPendingEvents();
});
```

But we hit a deadlock on Daydream. `runOnUiThread` is never executed because `onPause` is waiting for a runnable being ran in the GLThread.

bluemarvin:
Would it make more sense to use the same queue we use on the other platforms and just stop using the GLView’s queue?

MortimerGoro:
I think it's a good idea for consistency. I'll run some tests

MortimerGoro:
I did some tests on this branch: https://github.com/MozillaReality/FirefoxReality/commit/3af65da83c7a13804e329c131c361059b8b12ecb

Found two problems:

- We can't correctly dispatch onDestroy() using the vrb::Queue because SurfaceView is already stopped in onPause. Tried with surfaceView.queueEvent only for that method, it works because the thread underlying thread is still alive.
- On the next run after a destroy there is a deadlock in onPause(). Not sure why yet...

Using the lock solution of this PR the life cycle seems good, onDestroy is correctly dispatched and next run after a destroy runs ok.

PR #946, opened 2019-02-13
Use real user height on Oculus 6DOF devices

MortimerGoro:


bluemarvin:
I wonder how this will affect the UI for the user if they are sitting vs. standing? Maybe the UX should be relative to the height at start up and then use the height when recalibrated?

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -680,6 +691,10 @@ struct DeviceDelegateOculusVR::State {
     aHeight = scale * (uint32_t)(vrapi_GetSystemPropertyInt(&java, VRAPI_SYS_PROP_SUGGESTED_EYE_TEXTURE_HEIGHT));
   }
 
+  bool Is6DOF() const {
+    return deviceType >= VRAPI_DEVICE_TYPE_OCULUSQUEST_START && deviceType >= VRAPI_DEVICE_TYPE_OCULUSQUEST_END;
```
cvan: is there a better way to feature detect 6DoF?

daoshengmu:
I will take a try today and give a review. Sorry for waiting.

daoshengmu:
It doesn't work properly on my Oculus 6DOF device. The controller's position is not updated. I move my controllers above my head, but they only be moved a little bit and still below my waist. The current master branch will update the controllers above my head, that's correct. Beside that, the controllers are using a wrong 3d model.

MortimerGoro:
@daoshengmu is the HMD display position right or it only happens with controllers? Feel free to continue the work on this PR because I don't have a device to test this

daoshengmu:
@MortimerGoro The HMD display position is right. I guess it is possible that the 6DOF property of controllers are not be set properly. I will keep working on it then.

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
daoshengmu: @MortimerGoro Sorry for waiting. Here is a small bug makes Oculus device can't recognize this is a 6Dof one, You should replace it with `return deviceType >= VRAPI_DEVICE_TYPE_OCULUSQUEST_START && deviceType <= VRAPI_DEVICE_TYPE_OCULUSQUEST_END;`, then we are good to merge it. Good job!.

@cvan They have a new API that is called `GetTrackedRemoteCaps()`. We can use it to improve this function. I will work on this next week.

MortimerGoro:
@daoshengmu I updated the PR with the fix. Before merging this have you tried what @bluemarvin mentioned? Is the height ok if your start the app while seated or standing?

daoshengmu:
> @daoshengmu I updated the PR with the fix. Before merging this have you tried what @bluemarvin mentioned? Is the height ok if your start the app while seated or standing?

It looks good. The UX are placed in the global space. When the headset is at sitting or standing mode, the UX's hight is fixed to the global space, we will see the correct result in our view space. It works the same with Oculus Home. When asking recalibrated, it only changes the direction of UX instead of its height.

PR #949, opened 2019-02-14
Add support for arm64-v8a on GoogleVR, Oculus and SVR

MortimerGoro:


bluemarvin:
The title confused me, I didn't think Wave had 64 bit libs. Should be GVR instead of Wave 🙂

app/build.gradle
```diff
@@ -168,10 +175,16 @@ android {
...
                 'oculusvrArmRelease',
+                'oculusvrArm64Debug',
+                'oculusvrArm64Release',
                 'svrArmDebug',
                 'svrArmRelease',
+                'svrArm64Debug',
+                'svrArm64Release',
                 'wavevrArmDebug',
                 'wavevrArmRelease',
                 'noapiArmDebug',
```
bluemarvin: Can you add noapiArm64[Debug,Release] also?

MortimerGoro: Done

MortimerGoro:
I also updated the GeckoView version. Today's 64bit nighlty showed much more performance on https://threejs.org/examples/webvr_ballshooter.html

> The title confused me, I didn't think Wave had 64 bit libs. Should be GVR instead of Wave

Yep, the mistake was only on the commit description :sweat_smile:

Issue #959, opened 2019-02-19
Improve the visual quality on Vive Focus


MortimerGoro:
We have some quality/aliasing issues on Vive:

- [x] Aliasing specially on the edges of the UI
- [x] Blurry/Shimmery/Too small fonts specially noticeable on the Settings panel
- [x] Aliasing on the resize bar lines
- [ ] Some lines in the UI (e.g. dividers on the settings or search suggestions) are hardly visible

Possible solutions that could help:

- Try to blur on the edges of the UI. If you check the Vive Launcher UI they have a big blur/glow color on the edges of their UI widgets
- Add some shadows to the fonts
- Make the divider lines bigger
- Add some fade to transparent on the resizebar edges. We could use a texture or add some more geometry:
![image 1](https://user-images.githubusercontent.com/1070794/53017615-b0d09b00-3450-11e9-9324-adf773f97c55.png)

thenadj:
@MortimerGoro We do not have a Vive Focus to test this on. Is there any work you need from us, modifying specific art assets, or can you fix this programmatically?

cvan:
I have one. I will test and report back here.

MortimerGoro:
@thenadj I pushed some PRs to make things better programmatically, waiting from feedback now from people with a Vive. I'll let you know if we need any more specific thing from art assets.

philip-lamb:
@MortimerGoro Did https://github.com/MozillaReality/FirefoxReality/pull/968 address all the issues documented here? If so, we can close this.

cvan:
@philip-lamb: no biggie, and I don't mean to be too pedantic. I am asking so I can better understand the triaged labels. I interpret the `UX` label as meaning "UX and/or UI guidance needed" - is that an incorrect assumption? Otherwise, nearly all visual bugs would be considered `UX` bugs. Thanks!

MortimerGoro:
@philip-lamb The 3 firstp oint has been addressed. We only need to fix ` Some lines in the UI (e.g. dividers on the settings or search suggestions) are hardly visible` on Vive to close this

PR #960, opened 2019-02-19
Blur the edges of UI background shapes

MortimerGoro:
See https://github.com/MozillaReality/FirefoxReality/issues/959 (tries to improve the first point on navigation bar and rounded corner dialogs)

This is a simple approach to add some blur, just adding a semitransparent stroke between the shape and the transparent border.

We can set a different size in `<dimen name="blur_radius">2dp</dimen>`

We could also set different radius values per platform (e.g this is more needed on Wave than on Oculus right now)

Anyone else that could test this on Vive Focus? (@philip-lamb @thenadj ?)

### DIFF
--- a/app/src/main/res/drawable/dialog_background.xml
+++ b/app/src/main/res/drawable/dialog_background.xml
@@ -2,5 +2,6 @@
 <shape xmlns:android="http://schemas.android.com/apk/res/android"
     android:shape="rectangle">
     <corners android:radius="15dp" />
+    <stroke android:color="@color/asphalt_blur" android:width="@dimen/blur_radius" />
     <solid android:color="@color/asphalt"/>
 </shape>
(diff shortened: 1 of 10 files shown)


thenadj:
@MortimerGoro I don't have a Vive Focus to test this on.

bluemarvin:
I've looked at this and played with the border values. I'm not sure it makes enough of a difference. @cvan have you tried this PR? We should probably make a decision on this.

bluemarvin:
@cvan I think this patch should just added a blur around the URL tool bar and its buttons. I would guess any other layout changes are from other patches. Is that true @MortimerGoro?

MortimerGoro:
It also adds a blur on the background of dialogs and prompts.

The auth promp is a good test to check the difference: https://www.httpwatch.com/httpgallery/authentication/authenticatedimage/default.aspx

e.g.: I think the edges of the prompt have less judder/flickering

PR #965, opened 2019-02-20
Improve resize UI quality

MortimerGoro:
See https://github.com/MozillaReality/FirefoxReality/issues/959

I'm using a texture lookup to gradate the alpha. It's easier to share the code this way (otherwise we'd need to add alpha border vertices to both quad and cylinder geometries in addition to adding support for vertex colors in https://github.com/MozillaReality/vrb or play with fragment shader solution based on distances from center)


bluemarvin:
Not sure why this failed on taskcluster. I retriggered the build.

bluemarvin:
Second try passed, not sure what failed first run.

MortimerGoro:
Closing this, new implementation in https://github.com/MozillaReality/FirefoxReality/pull/968

Issue #216, opened 2018-07-19
Enable Curved Windows (and explore whether to also support flat as option)


avrignaud:
Hardware: All
Steps to Reproduce: N/A
Actual Behavior: flat window
Expected Behavior: option to enable curved window, and presumably control curvature
Recommendation: enable option to toggle between flat and curved windows, and explore whether curvature window control makes sense

From YouTube video review: https://youtu.be/YfF0R2v0kEk?t=250

thenadj:
Trello card created for this issue. We'll explore designs for this as part of the v1.1 work. 
https://trello.com/c/IlzAFoYG/133-curved-browser-window-control

MortimerGoro:
- It might affect multiple window support or window placements designs. Does a curved design work if you don't look at it from the center? how can you move a curved window if you need to be on the center?
- Resizing may be affected too

thenadj:
@avrignaud What's the verdict on this? Seems like window positioning and cylindrical/curved windows don't play well together.

larsbergstrom:
I believe that we will not be doing this for v1.1. We are working on purely flat layers for 1.1, which will enable us to experiment with / investigate for post-1.1.

cvan:
a lot of folks complain about the curvature on [Reddit](https://www.reddit.com/r/OculusGo). personally, I don't find it helps readability, but I'm also not reading comic books or Wikipedia articles for hours, so I'm probably not the best person to weigh in on this. if we do allow the option, I'd recommend it be buried as a non-default preference in the Settings.

cvan:
cc @joshmarinacci: Might be nice to promote here on this Reddit thread this upcoming feature in the next release of FxR: https://www.reddit.com/r/OculusGo/comments/asegfn/oculus_browser_with_a_flat_screen_rather_than/

PR #979, opened 2019-02-28
Implement dynamic switch between flat and curved display

MortimerGoro:
This PR allows to switch between flat and curved display without a restart.

With Oculus Layers it reuses the swapChain so no surfaces are created during the switch.
With GL geometries it reuses the SurfaceTexture.

### DIFF
--- a/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
+++ b/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
@@ -225,14 +225,13 @@ public:
   OculusLayer::BindDelegate bindDelegate;
 
   void Init(JNIEnv * aEnv, vrb::RenderContextPtr& aContext) override {
-    if (this->swapChain) {
-      return;
-    }
-
     this->jniEnv = aEnv;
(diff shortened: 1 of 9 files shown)


app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -225,14 +225,13 @@ class OculusLayerSurface: public OculusLayerBase<T, U> {
   OculusLayer::BindDelegate bindDelegate;
 
   void Init(JNIEnv * aEnv, vrb::RenderContextPtr& aContext) override {
-    if (this->swapChain) {
```
bluemarvin: Does `Init()` get called multiple times? How can the device have a valid swap chain before calling `Init()`?

MortimerGoro: Init can be called from pause/resume life cycle events, and also just after a layer is created when the activity is already active in VR.

philip-lamb:
Resolves #216.

PR #980, opened 2019-02-28
Fix sticking out border in curved resize mode

MortimerGoro:
screen_shot_1902260557

Depends on https://github.com/MozillaReality/vrb/pull/61


MortimerGoro:
Done

Issue #812, opened 2018-11-22
Change `Gamepad#id` from "Gear VR Controller" to "Oculus Go Controller"


cvan:
# Hardware
Oculus Go

Unless there are particular Web Compatibility reasons (cc @bluemarvin), with the Oculus Go we should change the `Gamepad`'d `id` string from `Gear VR Controller` to `Oculus Go Controller`. If there are particular known sites that are (or were) affected, let me know and we can figure something out.

This noticeably would fix some WebVR video players/experiences that sniff for the `id` of the `Gamepad` object to use the respective controller model mesh.

### Firefox Reality

```js
navigator.getGamepads()[0]
```

```
Gamepad
​axes: Array [ 0, 0 ]
​buttons: Array [ GamepadButton, GamepadButton ]
​connected: true
​displayId: 1
​hand: "right"
​hapticActuators: Array []
​id: "Gear VR Controller"
​index: 0
​mapping: ""
​pose: GamepadPose { hasOrientation: true, hasPosition: false, position: null, … }
​timestamp: 191406
<prototype>: GamepadPrototype { id: Getter, index: Getter, mapping: Getter, … }
```

![screenshot](https://user-images.githubusercontent.com/203725/48879622-9ac42780-edda-11e8-8769-b958529ec23b.png)

<hr>

### Chrome

```js
navigator.getGamepads()[0]
```

```
Gamepad {id: "Oculus Go Controller", index: 0, connected: true, timestamp: 2653221748, mapping: "", …}axes: (2) [0, 0]buttons: (2) [GamepadButton, GamepadButton]connected: truedisplayId: 1hand: "right"id: "Oculus Go Controller"index: 0mapping: ""pose: GamepadPoseangularAcceleration: Float32Array(3) [0.09922924637794495, 0.46682077646255493, 0.21701952815055847]angularVelocity: Float32Array(3) [0.0015635552117601037, 0.003237674245610833, 0.0039522601291537285]hasOrientation: truehasPosition: falselinearAcceleration: Float32Array(3) [0.009476134553551674, 0.08094597607851028, -0.009985487908124924]linearVelocity: Float32Array(3) [0, 0, 0]orientation: Float32Array(4) [-0.030928365886211395, 0.06747458130121231, -0.6488243341445923, -0.75730961561203]position: null__proto__: GamepadPosetimestamp: 2654424216vibrationActuator: null__proto__: Gamepad
```

[...]

Issue #401, opened 2018-08-23
FxR controller and display names should match the native browser's names.


bluemarvin:
## Hardware

All

## Current Behavior

We use the same names on all platforms and do not match any existing platforms.

## Possible Solution

We should examine what names are used in each native browser and emulate them. A lot of code is already sniffing for specific platform names and it would be easier to just emulate than try and get existing websites to update their code to look for FxR specific strings.

## Context

The FxR controller doesn't work on sketchfab.com because they are looking for specific controller names in their code.

cvan:
Good catch. Removing `Gamepad#id` (the "display name") has come up quite a bit in spec discussions:

- https://github.com/w3c/gamepad/issues/73
- https://github.com/w3c/gamepad/issues/65
- https://github.com/w3c/gamepad/issues/9
- https://github.com/w3c/gamepad/issues/12

I understand that the main reason for the `Gamepad#id` sniffing is to ensure the `buttons` and `axes` are mapped correctly (or re-mapped for particular browser/OS/gamepad configurations, which I myself had to do quite a bit in making HTML5 games).

If there's not already enough information with the current properties, a dictionary of capabilities (per w3c/gamepad#12) could be added so developers don't need to sniff for the name.

MortimerGoro:
`We use the same names on all platforms and do not match any existing platforms.`

Actually we have already set different names for each display, so it should report different names on each. But yep, it's good to check we are using names that apps are checking correctly.

For the controllers we set the ones used in AFrame (e.g "Gear VR Remote Controller")

avrignaud:
@larsbergstrom and I will figure out webcompat loop.

cvan:
Whichever we choose, we should make sure they are supported by those defined in this popular library used with three.js: https://github.com/stewdio/THREE.VRController/blob/master/VRController.js. (And obviously make PRs when appropriate.) Oculus Go Controller has been recently added, FYI.

philip-lamb:
See also #812

cvan:
I've commented on https://github.com/w3c/gamepad/issues/73#issuecomment-469517729. Will update as I do some research on the actual sites possibly impacted by `Gamepad#id` sniffing.

cvan:
Is there more to be done here? Oculus Go + VIVE Focus `Gamepad#id` strings were changed in issue #812 / PR #981.

philip-lamb:
Closing as resolved.

PR #981, opened 2019-03-04
Replace Gamepad id from "Gear VR Controller" to Oculus Go/HTC HTC Vive Focus controllers

daoshengmu:
Per #812, we should use the same gamepad id, `Oculus Go Controller`, as other browsers when running in Oculus Go display.

### DIFF
--- a/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
+++ b/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
@@ -769,7 +769,7 @@ struct DeviceDelegateOculusVR::State {
           controller->SetButtonCount(count, 6);
         } else {
           // Oculus Go only has one kind of controller model.
-          controller->CreateController(count, 0, "Gear VR Controller");
+          controller->CreateController(count, 0, "Oculus Go Controller");
           controller->SetButtonCount(count, 2);
         }
 
--- a/app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
+++ b/app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
@@ -299,7 +299,7 @@ DeviceDelegateWaveVR::SetControllerDelegate(ControllerDelegatePtr& aController) 
     return;
   }
   for (int32_t index = 0; index < kMaxControllerCount; index++) {
-    m.delegate->CreateController(index, 0, "Gear VR Controller");
+    m.delegate->CreateController(index, 0, "HTC Vive Focus Controller");
   }
 }
 


bluemarvin:
Be aware this was done for compatibility reasons. So make sure changing it doesn't break some of existing immersive web sites.

philip-lamb:
We should also update the string exposed for the HTC Vive Focus controller. Per https://github.com/aframevr/aframe/pull/3876 (https://github.com/aframevr/aframe/pull/3876/files#diff-0cf01d3ef4445863595ddf0ff798ecfbR11) it should be corrected to 'HTC Vive Focus Controller' at https://github.com/MozillaReality/FirefoxReality/blob/master/app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp#L302 .

bluemarvin:
For reference: https://github.com/MozillaReality/FirefoxReality/pull/441

daoshengmu:
I begin to worry about that will affect content that didn't use A-Frame or didn't follow its latest update. I guess we do need more options from @cvan .

philip-lamb:
I'll log an issue to consider a compatibility setting in the UI. If we continuing incorrectly sending Gear VR as the identifier, we'll continue to diverge in webcompat from the Oculus browser. We can potentially sniff older AFrame versions before the Oculus Go controller and HTC Vive Focus controller strings were added.

cvan:
> If we continue incorrectly sending Gear VR as the identifier, we'll continue to diverge in webcompat from the Oculus browser.

I agree that we should report the same `Gamepad#id` as Oculus Browser does.

> We can potentially sniff older AFrame versions before the Oculus Go controller and HTC Vive Focus controller strings were added.

Assuming we can't reach out to the site owners of the more popular A-Frame experiences (e.g., the ones in the Content Feed), ensuring backwards compatibility with older A-Frame versions I don't think is a reasonable expectation.

On a related note, see this comment I made about `Gamepad#id` in general [here](https://github.com/MozillaReality/FirefoxReality/issues/401#issuecomment-469518151):

> I've commented on https://github.com/w3c/gamepad/issues/73#issuecomment-469517729. Will update as I do some research on the actual sites possibly impacted by `Gamepad#id` sniffing.

cvan:
So as to not break [Soundboxing](https://webvr.soundboxing.co/) (and likely other WebVR experiences I'm not aware of), we should probably wait before merging this. I've emailed the [Soundboxing](https://soundboxing.co/) staff. I'll respond here when I hear a response. Thanks, @daoshengmu!

MortimerGoro:
Let's wait to merge this until @cvan confirms that the contents feed demos are ready

MortimerGoro:
@cvan Have you also tested the controller change in HTC Vive focus?

Issue #638, opened 2018-10-17
Design: Update Settings Dialog to allow for better scaling


thenadj:
## Expected Behavior
Settings dialog doesn't scale very well. Right now most features are in Developer Options. 

## Possible Solution
Design a better way to allow better discovery of features and scaling.

thenadj:
@avrignaud Noted. Still adding DE-10 (Settings) for feedback. We're trying to understand how to organize settings and what goes where. Please review and send feedback. 
[DE-10 Settings_002.pdf](https://github.com/MozillaReality/FirefoxReality/files/2485502/DE-10.Settings_002.pdf)

keianhzo:
Which are the options for 1.1?

- Speech Data Collection
- Environment
- Crash Reporting
- Developer Options
- Privacy
- Report An Issue
- Controller Options
- Display

Is that right?

Any particular order?

avrignaud:
I think the Settings options for 1.1 are:
- Language 
- Speech (broken out into permission for collecting voice samples, and perhaps where we allow user to choose preferred voice search language? Latter might be better in Language though?)
- Environment
- Crash Reporting
- Developer Options
- Privacy & Security (think we remove the "Privacy" solo entry?)
- Controller Options
- Report an Issue

I don't think we need Search yet - since we still only have one default search engine, nor Account.

@larsbergstrom @bluemarvin Do we have any opinion on whether we should be asking GV to enable about:preferences support (so design can create a set of pref web pages to use), or whether we should invest time in extending the current settings UI?

thenadj:
Attached is the updated UI Spec for Settings. Ready for review. 
[Uploading UIS-16 Settings_013.pdf…]()

thenadj:
See this issue for additional content to be added to settings: https://github.com/MozillaReality/FirefoxReality/issues/634

keianhzo:
@thenadj I don't seem to find the UIS-16 Settings_013.pdf. The link above redirects to another issue so I'm not sure which one is the final version of the spec.

thenadj:
[UIS-16 Settings_013.pdf](https://github.com/MozillaReality/FirefoxReality/files/2530689/UIS-16.Settings_013.pdf)
Here you go @keianhzo

thenadj:
Please use this spec for updated visual styles and more consistency across all dialogs. 
[...]

Issue #705, opened 2018-11-01
Allow swappable Y-Axis option for touch pad scrolling


avrignaud:
## Hardware

Oculus Go

## Steps to Reproduce
Scrolling on touch pad can feel "reversed" to different users depending on their Y-Axis preference. 

## Expected Behavior

User should have option to swap Y-Axis for touch pad scrolling.

## Possible Solution

User should have option to swap Y-Axis for touch pad scrolling.

```
```

</details>

cvan:
Good catch! Good feature request to address. Though vertical scrolling is most common on web pages, I'd expect the scrolling to apply the same for the X-axis. That's what Operating Systems do (the user prefer to reverse the scrolling direction applies for both).

cvan:
I'm suggesting that we don't make a difference between the X or Y axis. We say simply `scrolling direction`.

![image](https://user-images.githubusercontent.com/203725/47945012-4f45ea00-debc-11e8-9bd9-4c4c7722ea1b.png)

avrignaud:
With the context that I'm an oldschool video game guy (including flight sims), I've always found having the two options to be more helpful for me personally? Apple's example works because they have additional context (animation showing it affects vertical scroll - don't think it affects horizontal?) I suspect users would be confused with just a "scrolling direction" option that doesn't give additional context? @thenadj love your thoughts too (but this isn't high priority for 1.1, to be clear).

caseyyee:
Also should have the ability to disable altogether.  I generally use the click-to-drag and the touchpad interferes with scrolling the page sometimes.

cvan:
> Also should have the ability to disable altogether. I generally use the click-to-drag and the touchpad interferes with scrolling the page sometimes.

Heh, I'm very impressed! I'm curious of what percentage of VR users avoid the touchpad as you do with their 3DoF controllers when browsing webpages in their VR browser. (Perhaps we should ask on https://www.reddit.com/r/OculusGo?)

MortimerGoro:
This have been implemented in this PR: https://github.com/MozillaReality/FirefoxReality/pull/984

Issue #214, opened 2018-07-19
UR Finding: Users would like to choose between different 3D environments


avrignaud:
Hardware: All
Steps to Reproduce: N/A
Actual Behavior: cannot customize/select different 3D environments
Expected Behavior: can select between different 3D environments to personalize browser
Recommendation: make it so

cvan:
I think this requires Design work. Forgive me if it is already filed and being worked on.

cvan:
I see this was originally filed as issue #181. Shall we close one of them as a dupe?

thenadj:
Yes, they seem like dupes. Once we have multiple environment we'll allow the user to switch them. We already have a preliminary design for this (UF-12 Settings_005) but will have to revisit in the future when we will officially work on this feature.

avrignaud:
Moving out of 1.1, should be addressed by overall customization UI in future,

thenadj:
Environment selector is in the latest Settings UI spec for Settings.
[UIS-16 Settings_014.pdf](https://github.com/MozillaReality/FirefoxReality/files/2539887/UIS-16.Settings_014.pdf)

cvan:
looking good, thanks.

![image](https://user-images.githubusercontent.com/203725/47882875-2e13c980-dde8-11e8-99c1-2041936b25b6.png)

1. could we have each row follow the list style (with the _Iron_ 1pt line and the padding)?
2. could we change `Environment Override` (with the `On`/`Off` toggle) to be consistent with the other environment rows (with a radio field)? as is, enabling the `Environment Override` will invalidate the selected custom environment anyway, so it'd make sense to treat them the same.
3. could we put a heading/title at the top of the modal to say `Environment`? I think it'd be helpful for all the other settings to have a contextual title too.

great work on these Settings changes! 👍

thenadj:
@jenniferhoang See comments above.

jenniferhoang:
Thank you for the recommendations @cvan, Please see attached for updated changes. 
[UIS-16 Settings_015.pdf](https://github.com/MozillaReality/FirefoxReality/files/2555353/UIS-16.Settings_015.pdf)

1 and 2. Row dividers with list style & Environment override UI change
![screen shot 2018-11-06 at 1 50 17 pm](https://user-images.githubusercontent.com/41343760/48099586-ee2c5800-e1d4-11e8-96fe-6130aa817114.png)

3. Dialog Menu title spec
[...]

PR #984, opened 2019-03-05
Update Settings Dialog to allow for better scaling

MortimerGoro:
Fixes https://github.com/MozillaReality/FirefoxReality/issues/638 and https://github.com/MozillaReality/FirefoxReality/issues/214 and #705


bluemarvin:
Now that the pref dialog is larger, it feels too close.  We might want to consider moving it back a bit.
The external storage preference seems to be broken:
screen_shot_1903052600
The switch shows `Allow` but it is still red. Also there is this dialog behind the preference dialog which can be seen if you close the preference dialog:
screen_shot_1903052626
If this dialog is accepted, then you get the system prompt. This might be a bug in GeckoView.

The environment dialog has the wrong text for the reset button:
screen_shot_1903051508

The privacy dialog has the wrong text for the reset button as well.
screen_shot_1903051601
I also really dislike using toggle buttons for this functionality. My expectation as a user is that clicking on it will change the preference. I find this dialog confusing and unexpected:
screen_shot_1903051726

bluemarvin:
Looks like the external storage is a GeckoView issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1527074

bluemarvin:
One other thing that might be useful is to add a `Use Current Page` button to the Homepage option.

bluemarvin:
Got this crash while testing noapi  https://crash-stats.mozilla.com/report/index/bp-bb896c7a-043d-4243-940e-1cb560190305

Unfortunately stack trace isn't helpful. I was in the `Display` prefs dialog and had pressed the back button. Haven't been abler to reproduce the issue yet.

MortimerGoro:
PR updated:

- Use button and label instead of the switch
- Fix settings distance due to larger size
- Fix wrong text in reset buttons
- Fix external storage read permission
- Improve text on honeybomb buttons (`sans-serif-light` doesn't look good in VR)

@philip-lamb regarding "Allow environment override" I agree that it's confusing. It's used to load enviroments from a hardcoded folder in sdcard. We need to find a better text, and maybe add a tooltip because there is not enough space to explain it. We can address that as a follow-up issue

bluemarvin:
Wiki page explaining external environments:
https://github.com/MozillaReality/FirefoxReality/wiki/Environments

MortimerGoro:
I can add a `more info` button below the env override switch that opens that URL, what do you think?

bluemarvin:
I got this crash when I hit the `reset` button in `Display` options dialog:

https://crash-stats.mozilla.com/report/index/d7b7d268-2f35-42da-8008-1dd570190306
https://crash-stats.mozilla.com/report/index/bp-9854595e-bf55-4655-b780-e42e40190306

It was reproducible.

bluemarvin:
It looks like the tracking protection toggle doesn't actually work.
If you go here: https://mozilla.github.io/tracking-test/
If TP is disabled you should see:
screen shot 2019-03-06 at 10 22 55 am
But I always see regardless of the setting:
screen shot 2019-03-06 at 10 23 07 am

app/src/common/shared/org/mozilla/vrbrowser/browser/SessionStore.java
```diff
@@ -906,6 +904,30 @@ public void setMultiprocess(final boolean enabled) {
...
+        if (state != null && state.mSettings.multiprocess != aEnabled) {
+            state.mSettings.multiprocess = aEnabled;
+            recreateSession(state.mSettings);
+        }
+    }
+
+    public void setTrackingEnabled(final boolean aEnabled) {
+        State state = getCurrentState();
+        if (state != null && state.mSettings.trackingProtection != aEnabled) {
+            state.mSettings.multiprocess = aEnabled;
```
bluemarvin: This seems to be the problem, looks like a CCP error. `multiprocess` should be `trackingProtection`

MortimerGoro:
@bluemarvin Pushed a new commit with fixes for tracking and the crash. The crash also happened on master, was not related to this PR.

MortimerGoro:
I have added a help icon in the environment override setting that opens the wiki page

philip-lamb:
Looks good, except for one minor UI issue... if you click "deny" on any permissions dialog, and later trigger the dialog again, the "don't ask again" box is really cramped. Is this our dialog or a system one? 
device-2019-03-08-112826

bluemarvin:
> Looks good, except for one minor UI issue... if you click "deny" on any permissions dialog, and later trigger the dialog again, the "don't ask again" box is really cramped. Is this our dialog or a system one?

That's the system's UI

philip-lamb:
Tested one additional item; scroll direction. The "natural" vs "reversed" should apply to both X and Y axes... this is basically the difference between touch-surface scrolling and joystick scrolling, so it should be consistent in both axes. In #705 there was a request to separate X and Y scrolling axes which I think we should consider in another issue (including design).

MortimerGoro:
@philip-lamb good catch. I updated the PR to fix the horizontal reversed scrolling

app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/dialogs/SettingsWidget.java
```diff
@@ -322,6 +346,19 @@ private void showDeveloperOptionsDialog() {
...
+        if (widget == null) {
+            widget = createChild(ControllerOptionsWidget.class, false);
+            mControllerOptionsDialogHandle = widget.getHandle();
+            widget.setDelegate(() -> onOptionsDialogDismissed());
+        }
+
+        widget.show();
+    }
+
     private void showLanguageOptionsDialog() {
```
cvan: UI: I notice in the [UIS-16 Settings design spec](https://github.com/MozillaReality/FirefoxReality/files/2909833/UIS-16_Settings_016.pdf), there is a subtle horizontal gradient that appears when there is further content on an end of the scrollable container.

image

currently, (even with the scrollbar) I sometimes get confused as to whether there is content to scroll and whether  the text is truncated. a gradient would help that. admittedly, it's probably more engineering effort to implement than the benefit would be.

image

app/src/main/res/layout/honeycomb_button.xml
```diff
@@ -19,11 +19,13 @@
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:layout_gravity="center"
+            android:paddingStart="3dp"
+            android:paddingEnd="3dp"
             android:layout_marginStart="15dp"
             android:layout_marginEnd="15dp"
-            android:fontFamily="sans-serif-light"
+            android:fontFamily="sans-serif"
```
cvan: minor: Q: the drop-shadows around the text – were those added by changing the `android:fontFamily`? or are they defined somewhere else?

image

image

app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/options/PrivacyOptionsWidget.java
```diff
@@ -0,0 +1,199 @@
...
+    @Override
+    public void hide(@HideFlags int aHideFlags) {
+        super.hide(aHideFlags);
+
+        mWidgetManager.removeWorldClickListener(this);
+    }
+
+    private void togglePermission(ButtonSetting aButton, String aPermission) {
+        if (mWidgetManager.isPermissionGranted(aPermission)) {
+            showAlert(aButton.getDescription(), getContext().getString(R.string.security_options_permissions_reject_message));
```
cvan: UI nit: can we alter the padding to be ~34px (or whatever that is in Android `dp` units) from each corner?

image

vs.

image

app/src/main/res/layout/options_controller.xml
```diff
@@ -0,0 +1,84 @@
...
+        <org.mozilla.vrbrowser.ui.views.UIButton
+            android:id="@+id/backButton"
+            style="?attr/navigationBarButtonStyle"
+            android:layout_width="40dp"
+            android:layout_height="40dp"
+            android:layout_alignParentStart="true"
+            android:layout_alignParentTop="true"
+            android:layout_centerVertical="true"
+            android:layout_marginBottom="10dp"
+            android:src="@drawable/ic_icon_back"
```
cvan: UX request: to better provide the user with context when reading the particular options for the Settings dialog menu, can we add the **title** at the top to each?

image

app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/dialogs/EnvironmentOptionsWidget.java
```diff
@@ -0,0 +1,202 @@
...
+        setEnv(mEnvironmentsRadio.getIdForValue(env), false);
+
+        mEnvOverrideSwitch = findViewById(R.id.envOverrideSwitch);
+        mEnvOverrideSwitch.setOnCheckedChangeListener(mEnvOverrideListener);
+        setEnvOverride(SettingsStore.getInstance(getContext()).isEnvironmentOverrideEnabled());
+        mEnvOverrideSwitch.setHelpDelegate(() -> {
+            if (mAudio != null) {
+                mAudio.playSound(AudioEngine.Sound.CLICK);
+            }
+            SessionStore.get().loadUri(getContext().getString(R.string.environment_override_help_url));
```
cvan: this is helpful. thanks for adding :+1:

<hr>

P.S. UX nit request: to make it easier for developers who will likely be reading documentation from their desktop PC screens, it might be nice to have @philip-lamb submit a bit.ly request for a short URL (à la https://mzl.la/fxr or https://mzl.la/reality), and the tiny URL could possibly be shown in a tooltip upon hover.

image

app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/dialogs/EnvironmentOptionsWidget.java
```diff
@@ -0,0 +1,202 @@
...
+            if (mAudio != null) {
+                mAudio.playSound(AudioEngine.Sound.CLICK);
+            }
+
+            onDismiss();
+        });
+
+        String env = SettingsStore.getInstance(getContext()).getEnvironment();
+        mEnvironmentsRadio = findViewById(R.id.environmentRadio);
+        mEnvironmentsRadio.setOnCheckedChangeListener(mEnvsListener);
```
cvan: UX request: would it be possible to increase the radius of the click target for the radio such that clicking anywhere on the row (the thumbnail, the radio button, or anywhere around but not precisely on the text line) would trigger a checked-change event?

I found myself thinking the UI was broken, until I was more precise with my clicks.

image

app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/dialogs/EnvironmentOptionsWidget.java
```diff
@@ -0,0 +1,202 @@
...
+
+        widget.show();
+    }
+
+    private void onRestartDialogDismissed() {
+        show();
+    }
+
+    // WindowManagerDelegate.FocusChangeListener
+
```
cvan: nit: can remove newline

app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/dialogs/EnvironmentOptionsWidget.java
```diff
@@ -0,0 +1,202 @@
...
+    @Override
+    public void onGlobalFocusChanged(View oldFocus, View newFocus) {
+        if (oldFocus == this && isVisible() && findViewById(newFocus.getId()) == null) {
+            onDismiss();
+        }
+    }
+
+    // WorldClickListener
+
```
cvan: nit: can remove newline

app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/options/ControllerOptionsWidget.java
```diff
@@ -0,0 +1,161 @@
...
+
+    private RadioGroupSetting.OnCheckedChangeListener mPointerColorListener = (radioGroup, checkedId, doApply) -> {
+        setPointerColor(checkedId, doApply);
+    };
+
+    private RadioGroupSetting.OnCheckedChangeListener mScrollDirectionListener = (radioGroup, checkedId, doApply) -> {
+        setScrollDirection(checkedId, doApply);
+    };
+
+
```
cvan: nit: can remove newline

MortimerGoro:
Good catch @cvan on the radio button click area. I refactored the environment layout to make it work clicking on any area of the row.

PR #986, opened 2019-03-07
Fix editing only the first field on a DoubleEditSetting

### LINKED ISSUE #955: Editing only the first field of screen size and max screen do not get saved.
## Version

FxR 1.1.2
## Steps to Reproduce

1. Open settings widget
2. Click on display option.
3. Edit only the width of either window or max window option.
4. Click save.

## Current Behavior

The options retain their previous value.

## Expected Behavior

The values should be updated to the new entered value.

MortimerGoro:
Fixes #955


philip-lamb:
Saving the entered value now works, but there are some other weird issues here with resizing. I'll log them in another issue.

bluemarvin:
The screen size and max screen size settings pre-date dynamic window resizing. I'm not sure it even makes sense to have these in the visual prefs anymore.

MortimerGoro:
@cvan I'm going to merge this fix. We can create another issue for that problem if it isn't already filled

Issue #987, opened 2019-03-07
Use `substitute-local-geckoview.gradle` from Bug 1533465 for local GeckoView substitution.


ncalexan:
This is a development ergonomics change only.  Over in [Bug 1533465](https://bugzilla.mozilla.org/show_bug.cgi?id=1533465) I'm trying to unify how downstream GeckoView consumers (at least, Mozilla consumers) substitute local GeckoView versions.  (This is all leading up to Gradle composite builds, someday, but that day is not this day.)

This is an example of consuming that ticket in a downstream consumer.

philip-lamb:
Will be fixed by #1579

PR #988, opened 2019-03-07
Use `substitute-local-geckoview.gradle` from Bug 1533465 for local GeckoView substitutions. (#987)

ncalexan:
This needs to wait:

1. Until the Pre: part is handled, hopefully by bumping mainline's GV version forward.
2. Until [Bug 1533465](https://bugzilla.mozilla.org/show_bug.cgi?id=1533465) lands.


bluemarvin:
@ncalexan Will this patch make https://github.com/MozillaReality/FirefoxReality/pull/1082 unnecessary?

ncalexan:
> @ncalexan Will this patch make #1082 unnecessary?

No, they're unconnected.  I'm going to close this in favor of someday landing https://github.com/mozilla-mobile/android-automation-tools/pull/46, which will do this in a better way than this approach.

Issue #977, opened 2019-02-28
One of two controllers will disappear when using Oculus 6Dof devices.


daoshengmu:
## Hardware

Oculus 6Dof device

## Steps to Reproduce

1. Open FxR, it will disappear one of controllers intermittently.

## Current Behavior

One of controllers will disappear intermittently.

## Expected Behavior

We should see both controllers, unless there is only one controller be turned on.

## Possible Solution

I would guess this is an asyn problem of the model or image loader when using multiple controllers.

## Error Logs and Stack Traces

<details open>

```
```

</details>

daoshengmu:
I notice the controller models are loaded properly because the callback function at `LoadFinishedCallback() in ModelLoaderAndroid.cpp` is called. When this failure happens, there is only one controller geometry be added to the drawlList (`DrawableList::AddDrawable() in DrawableList.cpp`). I am trying to find out if it is removed by the culling mechanism. 

 @bluemarvin do you have any thought?


(Sorry for my ADB sometimes will interrupt the connection with the debugger, it makes the debugging work be more complicated to me.)

bluemarvin:
I would look to see if the Device is some how disabling the model.

PR #989, opened 2019-03-08
Fixed the second controller doesn't show up problem

daoshengmu:
Per #977. We find the second controller does not show up intermittently, and I notice it is because the `model` vector of `ControllerContainer.cpp` is not resized properly.

### DIFF
--- a/app/src/main/cpp/ControllerContainer.cpp
+++ b/app/src/main/cpp/ControllerContainer.cpp
@@ -44,7 +44,7 @@ struct ControllerContainer::State {
   }
 
   void SetUpModelsGroup(const int32_t aModelIndex) {
-    if (models.size() >= aModelIndex) {
+    if (aModelIndex >= models.size()) {
       models.resize((size_t)(aModelIndex + 1));
     }
     if (!models[aModelIndex]) {
--- a/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
+++ b/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
@@ -580,7 +580,13 @@ struct DeviceDelegateOculusVR::State {
     vrb::Matrix controllerTransform = vrb::Matrix::Identity();
     ovrInputStateTrackedRemote controllerState = {};
     ElbowModel::HandEnum hand = ElbowModel::HandEnum::Right;
+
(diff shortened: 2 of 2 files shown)


app/src/main/cpp/ControllerContainer.cpp
```diff
@@ -44,7 +44,7 @@ struct ControllerContainer::State {
   }
 
   void SetUpModelsGroup(const int32_t aModelIndex) {
-    if (models.size() >= aModelIndex) {
+    if (aModelIndex >= models.size()) {
```
bluemarvin: Good catch.

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -709,10 +715,16 @@ struct DeviceDelegateOculusVR::State {
...
+  bool IsOculusQuest() const {
     return deviceType >= VRAPI_DEVICE_TYPE_OCULUSQUEST_START && deviceType <= VRAPI_DEVICE_TYPE_OCULUSQUEST_END;
   }
 
+  bool Is6DOF() const {
+    // ovrInputHeadsetCapabilities is unavailable in Oculus mobile SDK,
+    // the current workaround is checking if it is Quest.
+    return IsOculusQuest();
+  }
+
```
bluemarvin: I find having two different `Is6DOF()` functions confusing. Also is it possible to use a Go 3DoF controller with the Quest?

daoshengmu: One of `Is6DOF()` is for controllers, and the other one is for headsets. Currently, we only use `ControllerState.Is6DOF()` for getting the capability from controllers. We didn't have `Is6DOF()` requirement for headsets yet, I just prepare it for the future. Ideally, `ovrInputHeadsetCapabilities` should tell us the capability  flag headsets, but it doesn't work right now, and I have checked it is the same with OVR mobile samples.

daoshengmu:
> I think this code would benefit from some clean up. I wonder if we could make it more flexible by keying off the reported caps instead of the type? And only use the type for the controller model file mapping?

yep. That's what I was doing. Probably, I am not doing good enough :)

PR #993, opened 2019-03-08
Fix setting window size oddities

### LINKED ISSUE #990: Window resize oddities
## Hardware
Tested on Oculus Go, from commit https://github.com/MozillaReality/FirefoxReality/commit/646c2f00ffe04e4b61947fed47c612749bb108ae

## Steps to Reproduce
See https://drive.google.com/file/d/1-eiA7GHaWG4kXLUOpq2URHBce1uqd9L_/view?usp=sharing
From clean install, open settings->Display. Observe window size. Click edit, then click save. Observe change in window size. Click reset. Now try with window widths of 801, and observe. Similarly 200, 400, changing only width.

Finally, edit window size using window drag handle mode. Then open window size setting in Settings->Display.

## Current Behavior
1. Window width of exactly 800 produces incorrect window height.
2. Editing only window width appears to result in window width being kept constant, and window height being adjusted to keep set aspect ratio.
3. Editing window sizes using drag controls has no effect on window sizes in settings pane.

## Expected Behavior
1. Window width of 800 should not result in strange window height.
2. Window width and height should correctly change to match setting.
3. Editing window size with the grab handles should change the values in the settings pane.

MortimerGoro:
Fixes #990

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/options/DisplayOptionsWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/options/DisplayOptionsWidget.java
@@ -115,7 +115,7 @@ public class DisplayOptionsWidget extends UIWidget implements
         mWindowSizeEdit.setHint1(String.valueOf(SettingsStore.WINDOW_WIDTH_DEFAULT));
         mWindowSizeEdit.setHint2(String.valueOf(SettingsStore.WINDOW_HEIGHT_DEFAULT));
         mWindowSizeEdit.setDefaultFirstValue(String.valueOf(SettingsStore.WINDOW_WIDTH_DEFAULT));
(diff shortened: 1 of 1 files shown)


bluemarvin:
I think there is some confusion over what the window size field represent. This is actually the default window size. So when the user clicks the `1x` button on the resize button bar, this is the size the window should return to.

MortimerGoro:
I agree that it's confusing. I understand it as the window pixel size for 1x world size: similar to a zooming feature, show more or less pixels in the same world meters size.

Should we just remove these settings? What's the use case to have them?

bluemarvin:
I think it should at least be renamed. I don't see anything wrong with allowing users to set the default size of a window.

philip-lamb:
OK, let's get this change in first and we can consider labelling change separately.

PR #996, opened 2019-03-11
Add optional audio gesture intent paramenter

MortimerGoro:


bluemarvin:
We are going to loose the ability to set prefs this way soon.

fernandojsg:
@bluemarvin does it mean that another way to implement that behaviour should be implemented? or that it will be impossible to set these prefs? :\

bluemarvin:
> @bluemarvin does it mean that another way to implement that behaviour should be implemented? or that it will be impossible to set these prefs? :\

The plan is to remove setting prefs in GV. I'll make sure it doesn't happen until we have a way to support what you need. One thing that we do in GV test is use the remote debugging protocol to set prefs. Would it be possible for the performance tests to use it as well?

fernandojsg:
> > @bluemarvin does it mean that another way to implement that behaviour should be implemented? or that it will be impossible to set these prefs? :\
> 
> The plan is to remove setting prefs in GV. I'll make sure it doesn't happen until we have a way to support what you need. One thing that we do in GV test is use the remote debugging protocol to set prefs. Would it be possible for the performance tests to use it as well?

Not sure how that works, could you provide me some example o link to doc to understand what should I do in order to set the prefs that way?

bluemarvin:
> Not sure how that works, could you provide me some example o link to doc to understand what should I do in order to set the prefs that way?

Looks like it might be a little involved unfortunately. This is the documentation for the [Remote Debugging Protocol](https://docs.firefox-dev.tools/backend/protocol.html). And the [Java code](https://searchfox.org/mozilla-central/source/mobile/android/geckoview/src/androidTest/java/org/mozilla/geckoview/test/rdp) that uses the protocol for GeckoView tests. Snorp found this: [Foxdriver](https://saucelabs.com/blog/foxdriver-a-nodejs-remote-debugging-client) but looks like it might be abandoned. I might be able to help write some infrastructure to help make this easier. Just need to figure out what the best strategy is. Maybe we should create an issue some where and continue the conversation there.

fernandojsg:
@MortimerGoro is this anyhow different to the same option on firefox desktop? I tried to set it there
image
and I got the same error on the demo https://fernandojsg.github.io/cruciform/?play&quality=low
```
NotAllowedError: The play method is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.
```

fernandojsg:
> > Not sure how that works, could you provide me some example o link to doc to understand what should I do in order to set the prefs that way?
> 
> Looks like it might be a little involved unfortunately. This is the documentation for the [Remote Debugging Protocol](https://docs.firefox-dev.tools/backend/protocol.html). And the [Java code](https://searchfox.org/mozilla-central/source/mobile/android/geckoview/src/androidTest/java/org/mozilla/geckoview/test/rdp) that uses the protocol for GeckoView tests. Snorp found this: [Foxdriver](https://saucelabs.com/blog/foxdriver-a-nodejs-remote-debugging-client) but looks like it might be abandoned. I might be able to help write some infrastructure to help make this easier. Just need to figure out what the best strategy is. Maybe we should create an issue some where and continue the conversation there.

Yep, I'll create an issue and we could move the discussion there :+1:

fernandojsg:
fwiw in chrome is working correctly using this flag
image

MortimerGoro:
@fernandojsg I handled more preferences so you can test this again.

The new param should be called `media.autoplay.enabled`

fernandojsg:
I tried it and work correctly with the latest changes

PR #1002, opened 2019-03-12
Discard back button presses to exit the app in WaveVR and Daydream

### LINKED ISSUE #568: Prompt users if they want to quit on back button press
## Hardware

WaveVR and Daydream
## Steps to Reproduce

1. Launch FxR
2. Press back button on controller

## Current Behavior

FxR exits to device home environment.

## Expected Behavior

FxR should prompt user if they want to exit FxR

## Possible Solution

On the Go, when the user pushed the back button with no more history, it will trigger a system UI asking if the user wants to exit or continue. On the Daydream and WaveVR the user is dumped back into the home env with out prompting which can be annoying when mashing the back button one too many times.

MortimerGoro:
Fixes #568


bluemarvin:
Not sure why the build failed. Re-triggered and passed: https://tools.taskcluster.net/groups/Jiv0bonlTvOVUzRzopmOSA/tasks/cirCzX-5S96foglaaIWARw/details

PR #1003, opened 2019-03-12
Fix unexpected forward page actions due to SwipeRight gesture on WaveVR

MortimerGoro:


app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
```diff
@@ -426,6 +426,7 @@ DeviceDelegateWaveVR::ProcessEvents() {
         break;
       case WVR_EventType_TouchUntapped: {
         VRB_DEBUG("WVR_EventType_TouchUntapped");
+        break;
```
bluemarvin: Good catch, I was noticing this problem just yesterday. It was super annoying.

cvan: we should probably fix the braces and the indenting of the `case`s [above](https://github.com/MozillaReality/FirefoxReality/pull/1003/files#diff-d6e5cad7b6d966752af1293c96fc7155R421), just in case

Issue #1027, opened 2019-03-16
Add a pref to enable developer options


fernandojsg:
The same way we can set a pref to enable `e10s` it could be great to have a preference to enable remote debugging and redirect logcat as everytime I run a test I restart the browser and these preferences get lost.

cvan:
If adb props get persisted on headset reboot (forget, but I think they do not), can we support those too?

This would be a helpful addition. Thanks for filing.

philip-lamb:
See also #874.

PR #1029, opened 2019-03-18
If Developer mode is enabled on device, default to 'Remote debugging'…

### LINKED ISSUE #874: Automatically enable `Remote Debugging` and `Redirect Console to Logcat` when in Developer Mode (Oculus' Developer Mode, adb, or Android Studio)
## Hardware
Oculus Go
ab56151

## Steps to Reproduce

1. Ensure Developer Mode is enabled for your device (from the Oculus mobile companion app, open `Settings`, select the headset in the list, then enable from `More Options > Developer Mode`)
2. Run in the Terminal: `adb uninstall org.mozilla.vrbrowser`
3. Sideload an APK from `master` via `adb` or from Android Studio
4. The first two developer options, `Remote Debugging` and `Redirect Console to Logcat`, in `Settings > Developer Options` are default to `Off`.

## Current Behavior
The first two developer options, `Remote Debugging` and `Redirect Console to Logcat`, in `Settings > Developer Options` are default to `Off`.

## Expected Behavior
If we know FxR is in Developer Mode, it'd be nice for the first two developer options, `Remote Debugging` and `Redirect Console to Logcat`, in `Settings > Developer Options` to default to `On`.

philip-lamb:
… and 'Console logs' both enabled. Fixes #874.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/SettingsStore.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/SettingsStore.java
@@ -36,8 +36,6 @@ public class SettingsStore {
     private SharedPreferences mPrefs;
 
     // Developer options default values
-    public final static boolean REMOTE_DEBUGGING_DEFAULT = false;
-    public final static boolean CONSOLE_LOGS_DEFAULT = false;
     public final static boolean ENV_OVERRIDE_DEFAULT = false;
     public final static boolean MULTIPROCESS_DEFAULT = false;
     public final static boolean TRACKING_DEFAULT = true;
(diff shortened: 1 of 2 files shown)


MortimerGoro:
It works for me in the Vive Focus, but not in Oculus Go.

bluemarvin:
Remote debugging is disabled by default for both performance and security reasons. The tradeoffs should probably be understood before proceeding with this patch.

philip-lamb:
> Remote debugging is disabled by default for both performance and security reasons. The tradeoffs should probably be understood before proceeding with this patch.

I got an authoritative response from the Developer Tools team, which notes that the performance effect is limited to the loading of the debugger server and dependencies (a load time of 300ms was mentioned) and that if this was undesirable, loading of the debugger could be delayed until after the UI was up. Otherwise, there is no performance effect unless a session is connected.

> It works for me in Vive Focus, but not in Oculus Go.

I got a response from the Oculus engineers and have added the switch used in their implementation.

MortimerGoro:
Is there any security implication of enabling it by default?

philip-lamb:
> Is there any security implication of enabling it by default?

No, but considering that having these options enabled by default imposes a small time penalty on app launch and that many developers who have their device in developer mode won't also need these on by default,  I think the method proposed in https://github.com/MozillaReality/FirefoxReality/issues/1027 is better.

PR #1032, opened 2019-03-18
Set correct Oculus WebVR device name

MortimerGoro:
We are using "Oculus" instead of "Oculus Go" for the WebVR device name.

I found that this affects Sketchfab performance (with "Oculus Go" it uses a lower quality model/shader). Could this affect other engines @fernandojsg, @cvan?

@daoshengmu Please review the correct name for Quest

### DIFF
--- a/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
+++ b/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
@@ -988,7 +988,11 @@ DeviceDelegateOculusVR::RegisterImmersiveDisplay(ImmersiveDisplayPtr aDisplay) {
     return;
   }
 
-  m.immersiveDisplay->SetDeviceName("Oculus");
+  if (m.IsOculusQuest()) {
+    m.immersiveDisplay->SetDeviceName("Oculus Quest");
+  } else {
+    m.immersiveDisplay->SetDeviceName("Oculus Go");
+  }
   uint32_t width, height;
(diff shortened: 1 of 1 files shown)


app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -988,7 +988,11 @@ DeviceDelegateOculusVR::RegisterImmersiveDisplay(ImmersiveDisplayPtr aDisplay) {
     return;
   }
 
-  m.immersiveDisplay->SetDeviceName("Oculus");
+  if (m.IsOculusQuest()) {
+    m.immersiveDisplay->SetDeviceName("Oculus Quest");
```
cvan: to prepare for the [WebXR Device API](https://immersive-web.github.io/webxr/#applicationflow) (which intentionally does not expose identifier strings for devices), shouldn't we begin deprecating [`VRDisplay#displayName`](https://immersive-web.github.io/webvr/spec/1.1/#dom-vrdisplay-displayname)?

daoshengmu: It looks correct. Getting `displayName: "Oculus Quest"` from Oculus browser.

cvan: :+1: good to know that we have parity with the Oculus Browser. in that case, we can defer deprecating the device string until WebXR is implemented (in both browsers) and web-compatibility isn't affected.

PR #1039, opened 2019-03-22
Do not dim scene brightness when selecting environments

MortimerGoro:


MortimerGoro:
@cvan I'm merging this to tag the release build for today, follow-up if you find something

Issue #620, opened 2018-10-12
English keyboard redesign


caseyyee:
## Hardware
Go

## Steps to Reproduce
1. Click URL bar
2. Enter a IP address

## Current Behavior
It takes way more key presses than should be required. 
We need to solve for the constant shifting and cursoring needed to entering common url's and strings.
`# _ - = ! ? / + @ % :` are all very commonly used and should be available. Some of these require shifting twice to get to.

## Possible Solution
Add a numeric keys to the top row so you don't have to shift the keyboard between letters and numbers to enter addresses and strings. Also add commonly used symbols, at most 1 shift away, but ideally none. We are not limited by screen real-estate in VR. We should use this to our benefit.

keianhzo:
Another alternative is to add specific keyboard layouts for different input types. We currently have assigned symbols1 for numeric and the standard qwerty for the rest. We could add more keyboard variants for other input types like URL, Phone, Email, etc.

larsbergstrom:
@keianhzo Do we need to look at this for 1.1?

keianhzo:
@larsbergstrom This would require a new keyboard spec, not sure we have design & dev room for this in 1.1, I'd just keep it backlogged for 1.2.

caseyyee:
@thenadj Thanks!  Also another one is the "-", which we come across a lot in URL's.  eg: 
"eff.com/spot-the-surveillance" is a recent one that frustrated me recently.

Given that we have so much "screen" real estate (unlike mobile), I don't see why we need to _shift_ out the keyboard when we have the space to work with.

thenadj:
@caseyyee Yep, '-' is already on our radar for top level symbols. And we agree on number layout in three rows rather than top level row. Thx for your feedback!! We are in the final steps of putting recommendations together and I'll share them with the team soon.

philip-lamb:
https://trello.com/c/VLsCI8iC/35-uis-6-english-keyboard

Issue #84, opened 2018-06-29
Could we please have a shift key on the right in the keyboard?


rtanglao:
a shift key on the right enables me to type my passwords more quickly!

PR #1050, opened 2019-04-01
English keyboard redesign

MortimerGoro:
Fixes https://github.com/MozillaReality/FirefoxReality/issues/620

This PR implements the english keyboard redesign. Other parts (e.g. autocompletion UI, placement improvements) will be implemented in different PRs.

This PR mimics the spec but some things I'd consider to change after trying the keyboard:

- It may be worth it to move the backspace and enter button to the top rows, and the extra symbols to the 3rd and 4rd rows
- The spec doesn't include the second shift key that some users requested and it's currently on master: https://github.com/MozillaReality/FirefoxReality/issues/84

Thoughts?


MortimerGoro:
cc @thenadj and @jvonitter

MortimerGoro:
Thanks for the detailed review @cvan 

PR updated:
- Added microphone and voice search implementation to the keyboard. Now you can do a voice search on any input in a web
- Updated keyboard theme with the colors defined in the final spec
- Added background for mode change button `ABC/%&=`
- Fixed clipped buttons

> I find myself accidentally pressing the wrong keys for Backspace and Enter keys. they feel they should be much placed higher, towards the top of the keyboard. I don't think it's just muscle memory, but because it's inconsistent with other system keyboards I use (non-VR and VR one

I agree, I had a similar feeling. I mentioned it in the issue. I bet for moving them to the top and adding the shift key to the right again. We should have enough space.

> it's odd to me that the numeric keyboard floats to the right of the keyboard. visually, it'd be nice if they appeared more connected. that way the (×) close icon would make more sense. I think the (×) close icon looks out of place even more so with the numeric pad now.

The `x` button is also going to conflict with the autocompletion UI, even if both views are merged. Is there any other place to put the X button @thenadj? e.g. Oculus browser has it inside the keyboard

> if you press on the keyboard or numeric pad, but not on a key, the UI sounds still get triggered, which is confusing IMO

I didn't get any sounds, don't we have sounds disabled?

MortimerGoro:
Reg the new color theme I like how it feels in `cave`, `underwater` and `snow` environments, but I feel  contrast issues with the `offworld` env color. Anyone feeling this too?

bluemarvin:
When using the microphone button to fill in text fields, the UI needs to be different. Right now is says "What would you like search on the web" which is confusing.

bluemarvin:
One thing I just noticed that may be existing functionality is that if you press a key and then move the pointer to a different key and then release, it will press the button the pointer is currently over, Is that what we want? It doesn't feel right to me. I wonder if a button shouldn't capture the pointer until the press is released?

thenadj:
We are working on design explorations to address the issues stated above. With more soon...
@MortimerGoro @avrignaud @jvonitter

cvan:
@MortimerGoro: no rush, but wanted to check if you were blocked on reviews, if you wanted to address the UX recommendations highlighted above in a follow-up enhancement issues, or if we are waiting for revised UIS specs from @thenadj. Thanks, everyone. This PR is looking real good. :+1:

MortimerGoro:
I saw there is a new PDF in the specs:
[DE-12 Keyboard_012.pdf](https://github.com/MozillaReality/FirefoxReality/files/3059620/DE-12.Keyboard_012.pdf)

There isn't exit button yet in the keyboard spec. Users can close the keyboard clicking by clicking in the world but I think discoverability can be a problem cc @thenadj 

> 
On pressing down a key (e.g., ABC/%&=) on the last row (but not yet releasing), the text of the appears lowered than it does in the letter keys. Can you tweak that value?

I'll fix that

>  When using the microphone button to fill in text fields, the UI needs to be different. Right now it says "What would you like search on the web" which is confusing.

I can fix the message when triggering the voice input from the keyboard.

> When the keyboard is being used for a text field on a web page / basically any place that's not directly entering in the URL bar, could we remove the Microphone button in the keyboard (or entirely?)? (This is consistent with how mobile browsers handle keyboard input, which are the closest analogues we have, aside from Oculus Browser, to battle-tested prior art for keyboard input for VR browsers.)

Are you proposing to entirely remove the voice input from keyboard? can you give more details? I don't fully understand the point of the mobile keyboard analogue

thenadj:
@MortimerGoro We decided for the time being to remove the close button all together and to let the user exit the keyboard by clicking anywhere outside of the keyboard. Ideally we'd user test this to get feedback and make an informed decision. From our own testing with friends, this was not an issue so far. If we see that users are struggling with this interaction, then we can add a close button to the left or right of the keyboard for which we will supply a design at that time. Placing the button at the top of the keyboard no longer works due to alignment with text boxes and auto-complete UI.

cc @jvo @philip-lamb @avrignaud

cvan:
> Are you proposing to entirely remove the voice input from keyboard? can you give more details? I don't fully understand the point of the mobile keyboard analogy

@MortimerGoro:

(See my recommendation below. I'd love to hear others' thoughts on items I've possibly overlooked.)

To illustrate the complexity of having the `Microphone` key always appear on the keyboard, here are the two user scenarios I can think of:

1. **have `Microphone` key use Speech-to-Text Voice Search after pressing the `Enter` key:** keyboard open from pressing the URL bar
2. **have `Microphone` key use Speech-to-Text API to fill the contents of the focused element on a web page:** keyboard open from pressing an editable text field on a web page (e.g., `<input>`, [`<input inputmode>`](https://css-tricks.com/finger-friendly-numerical-inputs-with-inputmode/), `<textarea>`, [`<div contenteditable>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable), [`<datalist>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist#Examples))

Because FxR's Speech-to-Text Voice functionality is used for redirecting users to a search results page, I'd recommend deferring scenario 2 these to later. In the meantime, I'd consider disabling/dimming the `Microphone` in scenario 2.

Given today's constraints of STT in FxR, what are others' thoughts on the UX, engineering, and product? Thanks, everyone!

CC: @thenadj @jvonitter @philip-lamb @bluemarvin @avrignaud

MortimerGoro:
PR updated with:
- Remove keyboard exit key
- Move backspace and enter to top
- Fix text lowered offset while hover/click.
- Fix text layout for special keys (e.g ENTER, ABC)
- Fix visual issues with special characters (longpress popover characters)
- Fix first key misaligmnet (in a different PR: https://github.com/MozillaReality/FirefoxReality/pull/1069)
- Cancel keyboard presses if keydown was produced on a different key (in a different PR: https://github.com/MozillaReality/FirefoxReality/pull/1070)

bluemarvin:
Looks good. But why is the numeric keyboard inverted? On a physical keyboard it goes:
```
789
456
123
0
```
While ours is:
```
123
456
789
 0
```

bluemarvin:
On Focus when the key pop-up is shown, there is a gray border across the bottom of the keyboard:
Screen_Shot_1904105541

cvan:
for `<input type="tel">` and `<input type="num">` on mobile browsers (and ATMs in IRL), numeric pads are shown in this order:

```
123
456
789
 0
```

image

the consistency w/ how users enter numerals in other browsers probably helps. I know I expect that format when I register for sites/shop online (more than I'm afraid to admit).

bluemarvin:
I guess for me, since we are simulating a physical keyboard in space, I expect the numeric keypad layout to match the physical layout.

thenadj:
@bluemarvin We switched the order of the numbers around to match mobile numeric keyboard which originated with the old touchtone phone. While the overall appearance of the keyboard is more like the physical hardware version, the numbers pad on the right feels more like the mobile one with the # and * at the bottom. We're trying to get a feel for things in XR and be open to mixing metaphors. Let's roll this version out with the mobile order and add it to our next usability study to see how users respond. For us it felt more natural to reverse the order. 
cc @jvonitter @avrignaud @philip-lamb @cvan

thenadj:
@philip-lamb Yes, we were made aware of that by Janice last week. We have an updated keyboard design in this Trello card: https://trello.com/c/ZCy2A0qs/294-pro-3-keyboard-refinements with the removed microphone and the added functionality of the auto complete bar and the close and move button. Some functionality still needs to be discussed before we can add final spec to this issue here. We'll discuss this with you shortly in our upcoming meeting.

MortimerGoro:
PR updated replacing the microphone input with the globe icon. 

The globe button implementation will land in the Chinese Pinyin keyboard PR

thenadj:
@MortimerGoro Cool! Can you add a video so we can see what the interaction looks like?

MortimerGoro:
@thenadj I posted a chinese keyboard video in #fxr channel in slack (Github didn't allow to put it here). Can you install an APK if I send you one? That way you can try both the English and Chinese keyboards. (Note: English keyboard doesn't have auto-completion implemented yet)

thenadj:
Ok, just saw it. Great!! Yes please email me an APK.

Issue #1051, opened 2019-04-02
Improve blurry YouTube videos by enforcing default minimum of 1440p (QHD)


cvan:
## Hardware
All

## Steps to Reproduce
1. Load a YouTube video - especially with the 360 ones it's noticeable.
2. Notice the blurry quality.
3. The user has to know to press the Settings button, then the Quality button, then select a reasonable quality, and then the quality is good enough.

## Current Behavior
The default playback quality of YouTube's in FxR are unreasonably poor.

## Expected Behavior
The default playback quality of YouTube's in FxR are of a fair quality (i.e., 1440p).

## Context
Discussed in @MortimerGoro's PR #1047: https://github.com/MozillaReality/FirefoxReality/pull/1047#pullrequestreview-220604653

cvan:
done in PR #1052

Issue #837, opened 2018-11-26
"360 video playback is not supported" message on YouTube


caseyyee:
## Hardware
Go
a9da13

## Steps to Reproduce
1. Navigate to any 360 content on Youtube. (eg https://www.youtube.com/watch?v=LdUxW8_59bA)

## Current Behavior
YouTube displays "360 video playback is not supported on this browser"

![360](https://user-images.githubusercontent.com/1560884/49038978-54bcea00-f173-11e8-8b26-f2b4892e65b9.png)

cvan:
I also notice a similar warning appears for [VR180 videos](https://www.youtube.com/watch?v=PZtpnL8x21g):

> VR180 video playback is not supported on this browser.

![image](https://user-images.githubusercontent.com/203725/49058800-9aa99a80-f1d3-11e8-8f67-a5cd36b1e63d.png)

cvan:
this has been fixed since PR #1052, but Firefox chokes on the syntax error in the style selector here https://github.com/MozillaReality/FirefoxReality/pull/1052/files#r283601515, which causes the `Full screen is unavailable` message to still be visible (see issue #1056).

PR #832, opened 2018-11-26
Add extra parameters to Youtube override

MortimerGoro:
Oculus Browser adds some extra parameters to Youtube URL. Not sure what they do.

Issue #793, opened 2018-11-20
Support VR180 playback on YouTube


cvan:
## Hardware
All

## Steps to Reproduce

1. Load a VR180 video (not a 360 video) on YouTube https://youtube.com/playlist?list=PLU8wpH_Lfhmu_kk955BFwvV0yT0PEcenA
2. Notice the unwrapped video is not shown (despite the new UA Override).

## Current Behavior

Notice the unwrapped video is not shown (despite the new UA Override).

With the new UA Override, properly unwrapped YouTube 360 videos are shown with the new media controls (#330).

VR180 is a kind-of proprietary format created by YouTube/Google.

If you dig into the source code of a video detail page, you can find the original unwrapped video track. The browser *could* do some trickery to extract that info. from a YouTube VR180 video page, but we probably should not start that precedent. That is a job for WebExtensions' content scripts.

## Expected Behavior
For the user to be able to use FxR's media controls for setting the Projection Mode, YouTube should serve unwrapped 180 videos for the same UA we are using to get unwrapped 360 videos.

## Possible Solution

Ideally, we reach out to YouTube to ask them to change the VR180 video pages to serve a similar video player as the 360 ones.

Or the browser can introduce functionality to parse VR180 metadata and streamlined video playback.

## Context

Oculus Browser also does not support VR180.

The native Oculus Go app for YouTube plays VR 180 extremely well. The VR180 catalogue is relatively small today but seems to be growing rapidly.

For any solution we chose, we should move quickly so FxR is ready when VR180 becomes commonplace (even if for only cinephiles and today's VR enthusiasts).

## Error Logs and Stack Traces

<details open>

```
```

</details>

bluemarvin:
Is this still an issue? @daron-walters or @MortimerGoro is this something we can address with the Web Extension? We should close this if no longer applicable.

MortimerGoro:
The projection is broken. The problem is that we don't get the raw equirectangular video from Youtube. So we have to research a way to get it, as we did with 360 videos (Faking the UA). It will require some research. It might not be easy to guess, last time I checked Oculus Browser have the same problem. So Youtube hasn't checked the "Oculus UA" with 180VR format.

It might be a intended behavior from Youtube, so they can force users to use the Youtube App instead of browsers

PR #1052, opened 2019-04-02
improve default playback quality of YouTube videos to 1440p (or `vq=`, `quality=` query-string parameter) (fixes issue #1051)

cvan:
for issue #1051

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/SessionStore.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/SessionStore.java
@@ -59,7 +59,7 @@ import static org.mozilla.vrbrowser.utils.ServoUtils.isInstanceOfServoSession;
 import static org.mozilla.vrbrowser.utils.ServoUtils.isServoAvailable;
 
 public class SessionStore implements ContentBlocking.Delegate, GeckoSession.NavigationDelegate,
-        GeckoSession.ProgressDelegate, GeckoSession.ContentDelegate, GeckoSession.TextInputDelegate, 
+        GeckoSession.ProgressDelegate, GeckoSession.ContentDelegate, GeckoSession.TextInputDelegate,
         GeckoSession.PromptDelegate, GeckoSession.MediaDelegate, SharedPreferences.OnSharedPreferenceChangeListener {
 
     private static SessionStore mInstance;
@@ -1057,7 +1057,9 @@ public class SessionStore implements ContentBlocking.Delegate, GeckoSession.Navi
     private String checkYoutubeOverride(String aUri) {
         try {
             Uri uri = Uri.parse(aUri);
-            if (!uri.getHost().toLowerCase().contains("www.youtube.")) {
+            String hostLower = uri.getHost().toLowerCase();
+            if (!hostLower.endsWith(".youtube.com") &&
+                !hostLower.endsWith(".youtube-nocookie.com")) {
                 return null;
             }
             String query = uri.getQueryParameter("disable_polymer");
--- a/app/src/main/assets/web_extensions/youtube_webcompat/main.css
+++ b/app/src/main/assets/web_extensions/youtube_webcompat/main.css
@@ -1,5 +1,6 @@
-/* To hide harmless warning about 360° video playback. */
-div.alert-with-button-renderer,
-ytm-alert-with-button-renderer {
+/* To hide harmless warnings. */
+.alert-with-button-renderer, .ytm-alert-with-button-renderer, ytm-alert-with-button-renderer,
+.ytp-ad-module, ytp-ad-module
+.ytp-generic-popup, ytp-generic-popup {
   display: none;
 }
--- a/app/src/main/assets/web_extensions/youtube_webcompat/main.css
+++ b/app/src/main/assets/web_extensions/youtube_webcompat/main.css
@@ -1,3 +1,5 @@
-div.alert-with-button-renderer {
-	display:none;
+/* To hide harmless warning about 360° video playback. */
+div.alert-with-button-renderer,
+ytm-alert-with-button-renderer {
+  display: none;
 }
--- a/app/src/main/assets/web_extensions/youtube_webcompat/main.js
+++ b/app/src/main/assets/web_extensions/youtube_webcompat/main.js
@@ -1,19 +1,174 @@
 (function () {
   // If missing, inject a `<meta name="viewport">` tag to trigger YouTube's mobile layout.
-  const viewport = document.head.querySelector('meta[name="viewport"]');
-  if (!viewport) {
-    viewport = document.createElement('meta');
-    viewport.name = 'viewport';
-    viewport.content = 'width=user-width, initial-scale=1';
-    document.head.appendChild(viewport);
+  window.addEventListener('load', () => {
+    let viewport = document.head.querySelector('meta[name="viewport"]');
+    if (!viewport) {
+      viewport = document.createElement('meta');
+      viewport.name = 'viewport';
+      viewport.content = 'width=device-width, initial-scale=1';
+      document.head.appendChild(viewport);
+    }
+  });
+
+  const LOGTAG = '[firefoxreality:webcompat]'
+  const qs = new URLSearchParams(window.location.search);
+  let retryTimeout = null;
+
+  function getTruthyQS (key) {
+    if (!qs || !qs.has(key)) {
+      return false;
+    }
+    const valueLower = (qs.get('key') || '').trim().toLowerCase();
+    return valueLower === '' || valueLower === '1' || valueLower === 'true' || valueLower === 'yes' || valueLower === 'on';
   }
 
-  // Open the `Settings` menu.
-  document.querySelector('ytp-settings-button, .ytp-settings-button').click();
+  const prefs = {
+    hd: false,
+    quality: 1440,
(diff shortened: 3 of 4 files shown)


MortimerGoro:
I tried with this video but it still selects 380 or 480p by default: https://www.youtube.com/watch?v=sPyAQQklc1s

Can you check it?

cvan:
@MortimerGoro I fiddled with it a bit. fixing now, should have it in a good state by tomorrow. will ping you when it's ready to test again. thanks!

cvan:
> This is great, the hard part will be to maintain it when YouTube breaks it.

quite possibly, yes. still tinkering here and there. I think I can cut the code down considerably.

another thought: possibly we can remotely host the CSS + JS so if (I mean when) things break, we don't have to roll out new releases. thoughts?

bluemarvin:
Since we don’t sign the extensions remote hosting would be an exploit vector.

cvan:
> Since we don’t sign the extensions remote hosting would be an exploit vector.

yeah, I agree. CSP isn't enforced on these content scripts (yet; GV bug?), so we're already pushing it heh.

app/src/common/shared/org/mozilla/vrbrowser/browser/SessionStore.java
```diff
@@ -1042,7 +1042,9 @@ public void onCanGoForward(@NonNull GeckoSession aSession, boolean aCanGoForward
     private String checkYoutubeOverride(String aUri) {
         try {
             Uri uri = Uri.parse(aUri);
-            if (!uri.getHost().toLowerCase().contains("www.youtube.")) {
+            String hostLower = !uri.getHost().toLowerCase();
```
bluemarvin: Typo here ~!~

cvan: thanks (been running this WebExt on desktop for rapider development)

bluemarvin:
Web extensions are only intended for internal use right now in GV so there aren’t security measures by design I think.

philip-lamb:
Can we also comment/test whether this works against issues #832 #837 #793.

cvan:
> Can we also comment/test whether this works against issues 
Thanks for checking!

> #832
Nope, unrelated - we can probably close that. I haven't found it necessary.

> #837
Yep, will be handled. I will reference the issue in the commit message when I push it to this PR branch.

> #793 
No, but we could eventually in that issue. VR180 support will require a lot more investigation and likely coordination w/ other browser vendors.

cvan:
@MortimerGoro ready for another review, if you have a min

bluemarvin:
I just tested this patch with https://www.youtube.com/watch?v=hFjzFSidX3s&mozVideoProjection=360_auto and it default to 360p resolution.

MortimerGoro:
I see this error in the console

`[JavaScript Error: "SyntaxError: missing = in const declaration" {file: "moz-extension://a9fe3216-9463-45c6-bf52-4eb51c350ea4/main.js" line: 69 column: 10 source: "    const prefs.qualities = ["}]`

MortimerGoro:
If fixed the SyntaxError but it still doesn't work.

@cvan I see that you have changed the code to user player.setPlaybackQuality(). I already tried that the first time we added the webextension and added some comments there. I found that `setPlaybackQuality` is  a no-op in the current Youtube API. It's mentioned in the docs:

```
setPlaybackQuality is now considered a "no-op"; calling this function will not change the player behavior. The player will use a variety of signals to determine the optimal playback quality.
Users are able to manually request a specific playback quality via the quality selector in the player controls.
```

So I think the only way would be to try the fake touches again

cvan:
@MortimerGoro:

> @cvan I see that you have changed the code to use player.setPlaybackQuality(). I already tried that the first time we added the webextension and added some comments there. I found that `setPlaybackQuality` is  a no-op in the current Youtube API. It's mentioned in the docs:

Yeah, I can actually remove the [`player.setPlaybackQuality(newBestQuality)` line](https://github.com/MozillaReality/FirefoxReality/pull/1052/files#diff-44d6081a68b4885b80582ec64fc90a4dR144).

But what I discovered is that you can use [`player.setPlaybackQualityRange(newBestQuality, newBestQuality)`](https://github.com/MozillaReality/FirefoxReality/pull/1052/files#diff-44d6081a68b4885b80582ec64fc90a4dR150), and that actually seems to work.

I've been able to run my branch successfully on YouTube videos on desktop Firefox + Chrome, and the default minimum quality is changed or if specified by the `vq=` or `quality=` query-string parameter, that quality gets enforced.

Where I'm stuck on:

- What's made this branch difficult is the loading of the content script via GV's Web Extension API seems to load differently than desktop Firefox's. I might want to change `document_start` to `document_end` in the `manifest.json`.
- The firing of the `onStateChange` event from YouTube's video player isn't fired on FxR but is on Firefox desktop w/ the same UA (w/ `disable_polymer=1` correctly set in the URL) and Chrome too.
- For some reason (likely user-activation requirement or YouTube's policies for progressively limiting the experience based on screen size/bandwidth/performance/etc.), YouTube doesn't begin playing videos, and _most_ times [`player.getAvailableQualityLevels()` isn't available](https://github.com/MozillaReality/FirefoxReality/pull/1052/files#diff-44d6081a68b4885b80582ec64fc90a4dR60). I use `player.getAvailableQualityLevels()` to cross-reference with all the supported YouTube quality levels to identify which index in the array to select to change the quality. There's also code to find [the next best quality level](https://github.com/MozillaReality/FirefoxReality/pull/1052/files#diff-44d6081a68b4885b80582ec64fc90a4dR120) if `1440p` is not available (or the forced value in the query-string parameter of the URL).

To overcome these issues, I think I'm going to have to stop relying on these proprietary YouTube events to know when we are ready to set the quality.

I'll likely have to use a `setInterval` loop every couple milliseconds to keep checking until the player is ready and the `player.getAvailableQualityLevels()` actually returns the video's quality (which seems to be only after the video starts playing, which requires a user-activation click to play the video).

If anyone has any better approaches here, let me know.

As you can imagine, the development process of making changes with Android Studio -> Oculus Go every time there's a JS change is slow and difficult to remotely debug via desktop Firefox's `about:debugging` and the WebIDE.

Earlier, I did some experimenting with injecting a script hosted elsewhere so we'd be able to update the content script for YouTube whenever without having to ship a new release with the hardcoded content script.

@bluemarvin expressed concerns that because GeckoView's WebExtension API is not intended for public/end-developer consumption yet, it's likely unreasonable or irresponsible to inject a third-party script (even if it's ours and hosted on a Mozilla-owned domain). (CSP policies can be enforced, but still that's a possible MITM attack vector.)

The precedent I've seen in Gecko, specifically for Android WebCompat issues, is a hardcoded [extension that's stored in `mozilla-central`](https://searchfox.org/mozilla-central/source/mobile/android/extensions/webcompat) and bootstrapped into Firefox. Every new adjustment requires a new Firefox release, but it seems to not been an issue from the bugs and commit activity I read. @miketaylr likely has some recommendations of best practices here for a scalable, realistic solution for maintaining WebCompat content scripts in FxR.

To speed up development, I've been writing the code and running it from `about:addons` on desktop Firefox, and that's helped me iterate faster.

Writing the hacks has been fun, and I still have the old code for manually pressing the buttons in the DOM. So we could go back to that if the [`player.setPlaybackQualityRange(newBestQuality, newBestQuality)`](https://github.com/MozillaReality/FirefoxReality/pull/1052/files#diff-44d6081a68b4885b80582ec64fc90a4dR150) code doesn't run properly in FxR. (I got it all working in FxR several times but accidentally broke it. Fragile stuff, but I'll battle-test this and make sure it always works.)

This patch is in a broken state right now because I've rewritten it so many times. In the meantime, in between coordinating the v1.1.4 release and other work tasks, I'm going to prioritise working on this to get it consistently stable and ready to review + land. If anyone has recommendations on the above items, I'd appreciate the help and will adjust the patch accordingly. Thanks, everyone.

MortimerGoro:
I think `setInterval` is the only realiable way right now to wait for quality availabilities. I'm ok with using that for now

MortimerGoro:
just tried `player.setPlaybackQualityRange(newBestQuality, newBestQuality)` using remote debugger in FxR and desktop and it worked in both. Good catch! So if we make the load detection more reliable with setInterval it should be ready to land

bluemarvin:
I've updated the PR to use the methods discussed. Please give it a try.

MortimerGoro:
It works. I'm going to merge this for the release

cvan:
thanks for the assists! much appreciated :heart:

app/src/main/assets/web_extensions/youtube_webcompat/main.css
```diff
@@ -1,3 +1,6 @@
-div.alert-with-button-renderer {
-	display:none;
+/* To hide harmless warnings. */
+.alert-with-button-renderer, .ytm-alert-with-button-renderer, ytm-alert-with-button-renderer,
+.ytp-ad-module, ytp-ad-module
```
cvan: oops, missing comma (regressed issue #837 and #1056)

cvan: fixed in PR #1217: https://github.com/MozillaReality/FirefoxReality/pull/1217/files#diff-ebdf9c97fadbb79c463a9b4302a68491R3

PR #1053, opened 2019-04-02
Update Servo

paulrouget:
This fixes the bleeding issue with Servo.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/utils/ServoUtils.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/utils/ServoUtils.java
@@ -4,6 +4,8 @@ import android.content.Context;
 import android.util.Log;
 
 import org.mozilla.geckoview.GeckoSession;
+import org.mozilla.vrbrowser.browser.SettingsStore;
+import org.mozilla.vrbrowser.VRBrowserActivity;
 
 import java.lang.reflect.Constructor;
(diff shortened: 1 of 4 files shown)


app/src/common/shared/org/mozilla/vrbrowser/utils/ServoUtils.java
```diff
@@ -32,10 +39,11 @@ public static boolean isInstanceOfServoSession(Object obj) {
     }
 
     public static GeckoSession createServoSession(Context context) {
+        boolean isOculus = BuildConfig.FLAVOR_platform.equalsIgnoreCase("oculusvr");
```
jdm: Unused?

paulrouget: Erf…

MortimerGoro: It's better to use `SettingsStore.getInstance(context).getLayersEnabled()`

MortimerGoro: So it will add the border if we implement layers in other platforms, or don't add the border if we disable layers in Oculus

paulrouget: Addressed.

PR #1061, opened 2019-04-04
Pin to the two (big) cores on Oculus Go

### LINKED ISSUE #934: Evaluate pinning the browser to the big cores
@MortimerGoro What's the best place / way to evaluate perf right now? I did a quick spike on pinning the FxR threads to the first two / big cores:
https://github.com/larsbergstrom/FirefoxReality/commit/91fbd0d4cf844b78407259375374e11e95234362

Hubs and other WebVR seem slightly more responsive, but without more concrete evaluation steps, it's hard for me to say something definite.

MortimerGoro:
Fixes #934.

Do not merge yet, we need another perf profile

### DIFF
--- a/app/src/main/cpp/native-lib.cpp
+++ b/app/src/main/cpp/native-lib.cpp
@@ -20,6 +20,8 @@
 #include "DeviceDelegateSVR.h"
 #endif
 
+#include <sched.h>
+
 #include <android/looper.h>
 #include <unistd.h>
 #include "VRBrowser.h"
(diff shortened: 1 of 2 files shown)


bluemarvin:
Looks like gradle crashed on the taskcluster build, I re-triggered.

bluemarvin:
Is this something we are going to be able to check at runtime? What about CPUs with four big cores?

philip-lamb:
> Is this something we are going to be able to check at runtime? What about CPUs with four big cores?

We are likely to be able to optimise for ARM big.LITTLE on every device based on QC820/821/835/845 SoC (the QC855 replaces big.LITTLE with DynamIQ). I will look further into what we can do to dynamically identify and adapt.

philip-lamb:
Based on poor performance profile, we will set aside this style of optimisation for now.

PR #1064, opened 2019-04-08
Oculus Fixed Foveated Rendering

MortimerGoro:
Adds user selectable values for foveated rendering in the chrome and WebVR scenes


MortimerGoro:
PR updated to hide the WebVR foveated setting (until we have real foveating support in WebGL FBO/shaders)

Issue #1041, opened 2019-03-23
Git sha doesn't show up upon double-clicking the Version in Settings


cvan:
## Steps to Reproduce
1. Load the Settings.
2. Double-click on the Version.
3. Notice the git sha doesn't show up.

## Current Behavior
The git sha used to show up when you double clicked on the Version line.

Something must've regressed in [`SettingsWidget`](https://github.com/MozillaReality/FirefoxReality/blob/675218e/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/dialogs/SettingsWidget.java#L58-L76).

## Expected Behavior
The git sha used to shows up when you click on the Version line. We can remove the need for a double click and display it on a single click instead.

cvan:
handled in PR #1073

PR #1073, opened 2019-04-11
fix git-sha version in Settings (fixes issue #1041); improve click target

cvan:
fixes regression #1041

also improved the clickable region and removed some unneeded attributes. and ensured no pixels were harmed in the layout change.


MortimerGoro:
Works great! Thanks

PR #1082, opened 2019-04-15
Add snakeyaml to deps

snorp:
GeckoView uses snakeyaml now, and it is included as a dependency in
the Maven POM, but that doesn't help when we're using a local
GeckoView build.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -289,6 +289,7 @@ dependencies {
     implementation deps.support.app_compat
     implementation deps.support.annotations
     implementation deps.constraint_layout
+    implementation deps.snakeyaml
 
     // Android Components
     implementation deps.mozilla_speech
--- a/versions.gradle
+++ b/versions.gradle
@@ -38,6 +38,7 @@ versions.atsl_rules = "1.1.0-alpha4"
 versions.espresso = "3.1.0-alpha4"
 versions.android_gradle_plugin = "3.2.0"
 versions.kotlin = "1.2.71"
+versions.snakeyaml = "1.24"
 def deps = [:]
 
 def gecko_view = [:]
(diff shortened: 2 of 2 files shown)


snorp:
It...actually doesn't use snakeyaml yet cause it got backed out. But it will. Probably.

ncalexan:
You shouldn't need this, because [Bug 1533385](https://bugzilla.mozilla.org/show_bug.cgi?id=1533385) was backed out shortly after the issue FxR sees was raised.  We won't reland without confidence that FxR's debugging will be functional.

bluemarvin:
As discussed, this will be needed until a better method can be found to enable builds using a local build of GeckoView.

philip-lamb:
@bluemarvin Up to you whether this needs to be landed.

ncalexan:
> As discussed, this will be needed until a better method can be found to enable builds using a local build of GeckoView.

The underlying issue here is that FxR is taking/substituting a "bare .AAR" dependency.  No .POM means no dependency information; no transitive dependencies means that _any_ GV dependency has to be plumbed into FxR directly.  That's why we saw SnakeYAML failures: they weren't to do with debugging, they were to do with this substitution pattern.

The way to do this and to capture GV's transitive dependencies is to use the Maven repository produced by `./mach gradle geckoview:publishWithGeckoBinariesDebugPublicationToMavenRepository`.  That's what my (still unlanded, sorry!) [PR](https://github.com/mozilla-mobile/android-automation-tools/pull/46) does.  I will specifically test this scenario because I didn't appreciate this subtlety myself until now.

bluemarvin:
I going to land this until we can get a better solution as I just ran into this again today.

PR #1085, opened 2019-04-16
Remove  RECEIVE_BOOT_COMPLETED permission using manifest merger

MortimerGoro:


cvan:
@MortimerGoro: thanks! this fixes it. the `RECEIVE_BOOT_COMPLETED` permission is now no longer used.

Issue #1034, opened 2019-03-20
Add l10n toml


Delphine:
This is only a configuration change, to be used by the upcoming l10n ecosystem. Thus not tests or changelogs or QA tags on this one.
CC @Pike

Pike:
We have one, but I need to make a slight twist to it.

philip-lamb:
@Pike Is there anything I can do to move this along? We have a l10n.toml, but does this issue relate to us getting an entry into https://github.com/mozilla-l10n/android-l10n/blob/master/config.toml ?

philip-lamb:
Fixed by #1086

PR #1086, opened 2019-04-16
fixes 1034, minor adjustments to l10n.toml for android-l10n

Pike:
Sorry for the delay here, I did a small reformat on the locale list, and added the branch data we need for android-l10n.

Philip, are you a good person to rubber-stamp this?

CC @Delphine


cvan:
fixes issue #1034

PR #1088, opened 2019-04-17
Also trigger back with the right 6DOF controller

MortimerGoro:


cvan:
this makes me so happy. thank you!

PR #1093, opened 2019-04-19
Fix position of 6DoF controllers and beam position

bluemarvin:


app/src/main/cpp/ControllerContainer.cpp
```diff
@@ -3,7 +3,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-#include <assert.h>
+#include <cassert>
```
daoshengmu: do we need this change?

daoshengmu: Do we need this change?

bluemarvin: It's not needed at all, I removed the `#include` altogether.

philip-lamb:
Fixes https://github.com/MozillaReality/FirefoxReality-SecureBugs/issues/4

PR #1094, opened 2019-04-19
Fix voice search glitch

MortimerGoro:
After watching the video I think it happens just due to a missing background


bluemarvin:
I disabled the spinning animation and the glitch goes away. I may just do that if we can't figure out a better solution in the short term.

cvan:
@MortimerGoro @bluemarvin fixed the PNG

cvan:
patched in PR #1098. let's investigate + revisit a more permanent fix later. thanks, all!

PR #1097, opened 2019-04-19
Add Java device type API

bluemarvin:


app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
```diff
@@ -734,6 +735,13 @@ public int getPointerColor() {
         return SettingsStore.getInstance(this).getPointerColor();
     }
```
cvan: can we add a method for [`getDeviceType()`](https://github.com/MozillaReality/FirefoxReality/pull/1097/files#diff-94535c76764a3a4f9a3d7cbf496c0d7dR21) than returns a string for each (e.g., `"unknown"`, `"oculus-go"`, etc.)? or should I just import `DeviceType` directly?

app/src/common/shared/org/mozilla/vrbrowser/input/DeviceType.java
```diff
@@ -0,0 +1,24 @@
+package org.mozilla.vrbrowser.input;
+
+import android.util.Log;
+
+public class DeviceType {
+    // These values need to match those in Device.h
+    public static final int Unknown = 0;
+    public static final int OculusGo = 1;
+    public static final int OculusQuest = 2;
```
cvan: we should probably remember to add the others, but this is fine for now

bluemarvin: I only wanted to add what we needed.

app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
bluemarvin: You should just be using `DeviceType.getDeviceType()` directly.

PR #1099, opened 2019-04-19
Ensure browser UI appears at correct level for 6DoF head tracking

bluemarvin:


bluemarvin:
@cvan Give this a try and see if that fixes the UI offset for 6DoF. Probably should hold off landing until we have time to test thoroughly.

bluemarvin:
> this is a fantastic improvement. the Tray is no longer obscured and the dialogues appear at eye level.
> 
> could we tweak the height a bit so the Media Controls are closer to average height level?
> 

I think the media controls are on the high side? My PR should put the eyes at the same level as the Go.

bluemarvin:
I'm going to merge. The video UX should probably be a different issue?

cvan:
> I'm going to merge. The video UX should probably be a different issue?

yeah, it is - I filed it separately privately.

PR #1110, opened 2019-04-22
Prevent options reset button from making a sound even when muted

bluemarvin:


MortimerGoro:
See https://github.com/MozillaReality/FirefoxReality/pull/1114

bluemarvin:
`setSoundEffectsEnabled(false);` didn’t seem to work for the reset button when I tried it.

bluemarvin:
Fixed in #1114

PR #1111, opened 2019-04-23
Fix the controller back button again for Oculus.

bluemarvin:


bluemarvin:
The back  button seems to be working again on my Go. I'm not sure we need this now.

MortimerGoro:
It also works ok for me without the PR

bluemarvin:
Closing this. We know which functions to twiddle if it happens again.

PR #1121, opened 2019-04-24
Fix Voice dialog animation glitch

MortimerGoro:


bluemarvin:
I double checked and tried this PR and the glitch still happens. Maybe it's the blending method being used?

MortimerGoro:
Ok, we can close this one

PR #1122, opened 2019-04-24
Show error message for offline WiFi/Internet connection

### LINKED ISSUE #1075: Consider showing Offline toast message (or reloading the active page) when network connectivity is lost (or blocked by a captive portal) upon resuming FxR on Oculus Go
## Hardware
Oculus Go

## Steps to Reproduce
1. Load FxR to the Content Feed homepage.
2. Exit to the Oculus system.
3. Open Settings > Wifi.
4. Disconnect the connection.
5. Press the Back buttons to return to the paused FxR task window.
6. Click the Resume button to reopen FxR.
7. Notice some elements on the page are now missing, but the user is shown no Network Error page.

## Current Behavior
Resuming FxR affects visibility of elements on the page, instead of showing FxR's Network Error page.

## Expected Behavior
I'd expect, if upon resuming FxR the network connection is detected to now be down, that the either an Offline toast notification message is shown to the user _or_ the current page/tab would be reloaded and the user would see FxR's Network Error page. (I realise that's not necessarily always the user's expected. Could be problematic if the user was in the middle of a session, such as for login/registration/payment, or inputting data, such posting a comment on Reddit or writing an email.)

Nothing discernible in the JavaScript Console logs nor Network activity tab in Firefox's `about:debugging`/WebIDE.

MortimerGoro:
Fixes #1075

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
@@ -297,6 +297,8 @@ public class VRBrowserActivity extends PlatformActivity implements WidgetManager
             widget.onPause();
         }
         mConnectivityReceiver.unregister(this);
+        // Reset so the dialog will show again on resume.
+        mConnectionAvailable = true;
         super.onPause();
     }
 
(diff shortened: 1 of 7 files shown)


bluemarvin:
Showing in `onResume()` might be good so if they leave and come back the dialog is shown again?

PR #1136, opened 2019-04-30
Prevent disabled controllers from getting focus

### LINKED ISSUE #1135: Controller pointer fails to show until clicked.
## Hardware

WaveVR
## Steps to Reproduce

1. Launch FxR

## Current Behavior

No pointer is visible until the controller is clicked.

## Expected Behavior

The pointer should be visible upon application launch.

## Context

Was probably broken during dual controller refactor.

bluemarvin:
Fixes #1135

### DIFF
--- a/app/src/main/cpp/BrowserWorld.cpp
+++ b/app/src/main/cpp/BrowserWorld.cpp
@@ -269,11 +269,11 @@ BrowserWorld::State::EnsureControllerFocused() {
       break;
     }
 
-    if (!controller.leftHanded) {
+    if (!controller.leftHanded && controller.enabled) {
       right = &controller;
     }
 
-    if (!first) {
+    if (!first && controller.enabled) {
       first = &controller;
     }
   }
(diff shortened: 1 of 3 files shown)


bluemarvin:
Break 6DoF controllers. Investigating.

bluemarvin:
> Breaks 6DoF controllers. Investigating.

Fixed.

daoshengmu:
It works on Focus+. but I notice the light beam model is still shown up at the unfocused controller. We should hide the light beam once it is unfocused. Besides, the ray direction is a little too high, we need to do some adjustments when we replace with new controller models.

app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -269,11 +269,11 @@ BrowserWorld::State::EnsureControllerFocused() {
       break;
     }
 
-    if (!controller.leftHanded) {
+    if (!controller.leftHanded && controller.enabled) {
       right = &controller;
     }
 
-    if (!first) {
+    if (!first && controller.enabled) {
```
daoshengmu: I am a little confused with this `first`'s meaning. Why don't we replace it with `left`?

bluemarvin: Because it is the first enabled controller discovered, which might be the right hand controller. It can probably be cleaned up. Originally, I just gave the first controller discovered focus, but on the Quest it seems the order is random. So I added the code to default to the right hand until I found a way to default to the controller than has focus in the home environment.

bluemarvin:
@MortimerGoro I'm going to land, if you have any additional feed back, please file as a follow up issue.

PR #1143, opened 2019-05-02
Add Focus+ controller

MortimerGoro:


bluemarvin:
These models look like they are for the dev kit. I was able to find models for the Focus Plus on the device so will close this.

Issue #636, opened 2018-10-16
Keyboard should be more comfortable to use.


thenadj:
## Hardware
Oculus Go

## Current Behavior
Keyboard position is not optimal. 

##Recommend
Users should be able to reposition keyboard. Also look into smart positioning for text fields. 

See this issue for reference:
https://github.com/MozillaReality/FirefoxReality/issues/378

gfodor:
nice! worth checking out altspace or windows mixed reality's keyboard -- they avoid the "user can't see what they are typing" problem due to visual separate between keyboard + text field by including a text field above the keyboard that shows the value of the currently selected field. this approach has the added benefit of normalizing the visual and UX for text entry across all pages, so if a web page (for example) has a very small, hard to read text field or a poorly rendered font the user will still have an accessible way to see the current value as they edit it.

gfodor:
oh, I think steam's keyboard does this too in their steam VR web browser.

gfodor:
(so, you can still move the keyboard to somewhere comfortable for the user, but it reduces the need to have the text field be in the near visual field of the keyboard at all times)

cvan:
this was was filed as #415 and #378. we should forward-dupe one/both of them or close this issue as a dupe.

cvan:
also, see the keyboard in the Oculus Browser:

![image](https://user-images.githubusercontent.com/203725/47052400-eee54780-d15c-11e8-9c2e-abf78b38df62.png)

thenadj:
@gfodor Thanks for the feedback. We originally had a text field as part of the keyboard but were asked to remove it. Honestly can't remember the reason, maybe just scope to get to v1.0. We should definitely think about adding that back, it's a great idea for all the reasons you mentioned.

avrignaud:
@cvan I asked Nadja to open this Design issue so we'd have a copy in the Design Columns of the new GH project board. I think the issues you refer to above are already "In Progress" from the engineering perspective? We should chew on whether we want to be crisp on moving issues out of engineering -> design (and only have one), or have a design and engineering issue, allowing them to be tracked as separate issues in parallel. @larsbergstrom have an opinion?

Also Cvan, can you give an example of how one would forward-dupe an issue? I'm not familiar with the term, like to understand better. Thanks!

keianhzo:
I like the way Oculus does it, it center aligns the keyboard to the middle of the text field.

[...]

PR #1153, opened 2019-05-03
Hide tray when Keyboard is visible

MortimerGoro:
Fixes #1152

PR #1151, opened 2019-05-03
Implement keyboard language selection popup

MortimerGoro:
Fixes #1134. Fixes #1150.

PR #1155, opened 2019-05-03
Keyboard refinements

MortimerGoro:
Includes some keyboard refinements from the latest spec published in https://github.com/MozillaReality/FirefoxReality/issues/636

- Improved margins
- Added close button
- Autocompletion view only visible when there are candidates
- Keyboard top margin is the same on all languages whether there is auto completion or not

Note: This PR still doesn't change the keyboard world size, position or angle. I prefer to move that to other PR


MortimerGoro:
Better to test this together with #1151 and #1153

MortimerGoro:
Thanks for the review, I've increased the margins

PR #1158, opened 2019-05-06
Implement Spanish keyboard

MortimerGoro:
Note: This branch includes https://github.com/MozillaReality/FirefoxReality/pull/1151 to make review easier. It will be removed when #1151 lands


bluemarvin:
We are going to start landing the keyboard PRs and then fixing in follow up issues since the current number of PRs makes testing challenging.

cvan:
@bluemarvin @MortimerGoro: thanks, filed issues #1179 and #1180 to track the two items I identified below that will be useful to fix globally for the keyboards.

PR #1159, opened 2019-05-06
Fix missed keys on the keyboard

### LINKED ISSUE #1154: Missed keys on the keyboard
It's easy to miss a keyboard click if you move the controller fast after pressing the key (before releasing it inside).

Probably is a regression from: https://github.com/MozillaReality/FirefoxReality/pull/1070

We can add a timer check to fix this issue or back-out that commit.

MortimerGoro:
Fixes #1154


bluemarvin:
I'm going to land this, we can file follow ups if necessary. After typing with this patch, going back to the old keyboard behavior is too painful.

Issue #647, opened 2018-10-19
Creating localized test builds


Delphine:
Once localization work is completed for v1.1, testing phase will roll-out and will require localized builds to test on.
I'm thinking the best option would be that once localizations are back in your repo, that these builds get automatically updated on a daily basis (so, "Nightly builds") - so that every time we make a correction, we can verify it the next day (fwiw this is what we have for other mobile projects as well).
The agency needs to test localizations on device, so providing builds via APK will be the easiest, I believe.
We started talking about this with @bluemarvin and @nancyhang via email thread, but probably easier to track and prioritize this via GitHub issue :)
EDIT: something I forgot to mention is that testers will no have headsets, so will have to be testing in 2D mode

(also CC @pike and @gueroJeff)

philip-lamb:
Given that we currently have binaries from each build of master, this might be just a matter of making them easier to find.

Pike:
Yeah, I agree.

philip-lamb:
Fixed in #1172

PR #1172, opened 2019-05-07
add direct links to bleeding-edge developer APKs for each platform generated from Taskcluster builds

cvan:
for https://github.com/MozillaReality/FirefoxReality/issues/647#issuecomment-489023874

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ You can find us on Twitter [@MozillaReality](https://twitter.com/mozillareality)
 **[Download](https://fxr.glitch.me/)** [bleeding-edge](https://github.com/MozillaReality/FirefoxReality/commits/master) developer (_debug_) APKs for Firefox Reality:
 
 - [Oculus Go (`oculusvr`)](https://fxr.glitch.me/oculusvr)
-- [HTC VIVE Focus (`wavevr`)](https://fxr.glitch.me/wavevr)
+- [HTC VIVE Focus & Pico G2 (`wavevr`)](https://fxr.glitch.me/wavevr)
 - [Google Daydream (`googlevr`)](https://fxr.glitch.me/googlevr)
 - [Snapdragon VR (`svr`)](https://fxr.glitch.me/svr)
 - [No API / Android Emulator for PC (`noapi`)](https://fxr.glitch.me/noapi)
@@ -24,7 +24,7 @@ You can find us on Twitter [@MozillaReality](https://twitter.com/mozillareality)
 
 ## L10n
 
-Whenever a new string is added to a localizable strings file (strings.xml ,localpages.xml, ...) a string description must be provided as a comment above the new string. Also the project STRs wiki page must be updated with a key description, steps to reproduce and the expected results.
+Whenever a new string is added to a localizable strings file (strings.xml, localpages.xml, …) a string description must be provided as a comment above the new string. Also the project STRs wiki page must be updated with a key description, steps to reproduce and the expected results.
 The L10n wiki page can be found [here](https://github.com/MozillaReality/FirefoxReality/wiki/L10n).
 
 ## Setup instructions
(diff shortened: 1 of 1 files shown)


cvan:
@philip-lamb updated, thanks!

also, exposed all the various builds on the [landing page](https://fxr.glitch.me/):

image

README.md
```diff
@@ -6,15 +6,25 @@ You can find us on Twitter [@MozillaReality](https://twitter.com/mozillareality)
...
-## Bleeding-edge developer APKs
+## Download developer APKs (debug builds)
+
+**[Download](https://fxr.glitch.me/)** [bleeding-edge](https://github.com/MozillaReality/FirefoxReality/commits/master) developer (_debug_) APKs for Firefox Reality:
+
+- [Oculus Go (`oculusvr`)](https://fxr.glitch.me/oculusvr)
+- [HTC VIVE Focus & Pico G2 (`wavevr`)](https://fxr.glitch.me/wavevr)
+- [Google Daydream (`googlevr`)](https://fxr.glitch.me/googlevr)
+- [Snapdragon VR (`svr`)](https://fxr.glitch.me/svr)
+- [No API / Android Emulator for PC (`noapi`)](https://fxr.glitch.me/noapi)
```
cvan: if y'all can think of a clearer description of `noapi`, I'll update this

README.md
```diff
@@ -6,15 +6,25 @@ You can find us on Twitter [@MozillaReality](https://twitter.com/mozillareality)
 
 [![Task Status](https://github.taskcluster.net/v1/repository/MozillaReality/FirefoxReality/master/badge.svg)](https://github.taskcluster.net/v1/repository/MozillaReality/FirefoxReality/master/latest) [Build results](https://github.taskcluster.net/v1/repository/MozillaReality/FirefoxReality/master/latest)
 
-## Bleeding-edge developer APKs
+## Download developer APKs (debug builds)
```
cvan: this heading reads awkwardly IMO. any suggestions?

cvan:
P.S. I updated the script to serve the APK URLs from the last successfully completed build if the [latest Taskcluster build](https://github.taskcluster.net/v1/repository/MozillaReality/FirefoxReality/master/latest) is still *running* and not yet finished.

bluemarvin:
This is cool. My only hesitation is that we will now have dependencies on glitch. I think taskcluster has something called indexing that allows you to make linking easier? Maybe using indexing would help remove one layer (i.e. glitch)?

cvan:
> This is cool. My only hesitation is that we will now have dependencies on glitch. I think taskcluster has something called indexing that allows you to make linking easier? Maybe using indexing would help remove one layer (i.e. glitch)?

yeah, I don't want to use Glitch. built it there for ease of prototyping. I can easily move this to our paid Heroku instance today. does Taskcluster plan on improving its API to more easily link to this stuff?

README.md
philip-lamb: It's OK. The reason I think it's worth mentioning it explicitly is performance may be different for debug binaries, based on profiling and log code, optimization disabled etc.

README.md
philip-lamb: Maybe "No VR device API"? Up to you.

README.md
bluemarvin: From what I can tell these links are pointing to staging builds which are release builds, the only difference is they are signed with the staging key instead of the the release key and thus can not be used in App Stores. But they are fully optimized builds. We do also build debug version of each APK but these link do not seem to be pointing to them.
For example:
```
public/FirefoxReality-1.1.4-31281846-googlevr-arm-debug.apk
public/FirefoxReality-1.1.4-31281846-googlevr-arm-staging-signed-aligned.apk
```
The first apk is the debug build for GoogleVR and the second APK is the full release build signed with the staging key.

bluemarvin:
> yeah, I don't want to use Glitch. built it there for ease of prototyping. I can easily move this to our paid Heroku instance today. does Taskcluster plan on improving its API to more easily link to this stuff?

I'm pretty sure the indexing allows us to link directly to the apks. We just need to add the indexing rules to our taskcluster file. I just haven't had time to read up on how it works.

cvan:
> Please file the followup issue to move the links from glitch. Thanks.

thanks, I've filed issue #1200 and taskcluster/taskcluster#724.

PR #1173, opened 2019-05-07
Adding a new line in auto completion view when the width of candidates is too big.

daoshengmu:


daoshengmu:
Screen Shot 2019-05-06 at 11 19 22 AM

We will see this issue when using Zhuyin input (#1161)

We are trying to fix this problem, and this can be reproduced when typing `ㄉㄚˋ` or `ㄏㄠˇ` with Zhuyin keyboard.

app/src/common/shared/org/mozilla/vrbrowser/ui/views/AutoCompletionView.java
```diff
@@ -60,6 +62,8 @@ public AutoCompletionView(Context aContext, AttributeSet aAttrs, int aDefStyle)
     private void initialize(Context aContext) {
         inflate(aContext, R.layout.autocompletion_bar, this);
         mFirstLine = findViewById(R.id.autoCompletionFirstLine);
+        mLineWidth = WidgetPlacement.dpDimension(getContext(), R.dimen.keyboard_width);
+        mPerKeyWidth = WidgetPlacement.dpDimension(getContext(), R.dimen.keyboard_key_width);
```
daoshengmu: I am trying to get the correct line width from `mFirstLine `, but it always returns zero from `getWidth()` or `getmesakuredwidth()`. `R.dimen.keyboard_width ` is not exactly correct because our real width should be the same with `world_width`.

daoshengmu: @MortimerGoro I am trying to use post runnable to make sure this UI has been loaded. but it still shows zero.
```
       mFirstLine.post(() -> {
            mLineWidth = mFirstLine.getWidth(); // the width is still not ready...
        });
```
Alternatively, `R.dimen.world_width` is more close to the correct result.

app/src/common/shared/org/mozilla/vrbrowser/ui/views/AutoCompletionView.java
```diff
@@ -60,6 +62,8 @@ public AutoCompletionView(Context aContext, AttributeSet aAttrs, int aDefStyle)
     private void initialize(Context aContext) {
         inflate(aContext, R.layout.autocompletion_bar, this);
         mFirstLine = findViewById(R.id.autoCompletionFirstLine);
+        mLineWidth = WidgetPlacement.pixelDimension(getContext(), R.dimen.world_width);
```
MortimerGoro: It's not correct to use the world width here. Ideally the View should get the current view width and detect layout changes. You can override `onSizeChanged` in order to detect the size and layout of the whole view: https://developer.android.com/reference/android/view/View.html#onSizeChanged%28int,%20int,%20int,%20int%29

bluemarvin:
I'll wait for the review comment to be addressed before landing this one.

app/src/common/shared/org/mozilla/vrbrowser/ui/views/AutoCompletionView.java
daoshengmu: @MortimerGoro I think it is good enough after adding
```
@Override
    public void onSizeChanged(int w, int h, int oldw, int oldh) {
        if (mLineWidth == 0) {
            View scrollView = findViewById(R.id.firstLineScroll);
            scrollView.measure(MeasureSpec.UNSPECIFIED, MeasureSpec.UNSPECIFIED);
            mLineWidth = scrollView.getWidth();
        }
    }
```

The `onSizeChanged` will be delay dispatched after the first time of `setItems()`. Therefore, the mLineWidth is still zero, we need to use kMaxItemsPerLine to workaround this.

daoshengmu:
I decide to add `<dimen name="autocompletion_widget_line_width">530dp</dimen>`, it will make things simple.

PR #1175, opened 2019-05-08
Show correct enter key label based on input IME options

MortimerGoro:
- Most of the implementation already landed in #1116 but we were missing a update call when the keyboard is changed (in addition to autocompletion candidate changes)
- I changed the URL bar action to `go`.
- Test page to try `search` action: https://developer.mozilla.org/es/docs/Web/HTML/Elemento/input/search

It seems that GV reports action `go` for Amazon and Google searches. iPhone browser reports `search`for both pages. May that be another GV bug @bluemarvin @cvan ?

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/keyboards/BaseKeyboard.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/keyboards/BaseKeyboard.java
@@ -34,6 +34,11 @@ public abstract class BaseKeyboard implements KeyboardInterface {
     }
 
     @Override
+    public String getSpaceKeyText(String aComposingText) {
+        return StringUtils.getStringByLocale(mContext, R.string.keyboard_space_label, getLocale()).toUpperCase();
+    }
+
+    @Override
     public String getComposingText(String aComposing, String aCode) {
         return aComposing.replaceFirst(Pattern.quote(aCode), "");
     }
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/keyboards/KeyboardInterface.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/keyboards/KeyboardInterface.java
@@ -37,7 +37,7 @@ public interface KeyboardInterface {
     String getComposingText(String aComposing, String aCode);
     String getKeyboardTitle();
     Locale getLocale();
-    default String getSpaceKeyText(String aComposingText) { return ""; }
+    String getSpaceKeyText(String aComposingText);
     String getEnterKeyText(int aIMEOptions, String aComposingText);
     String getModeChangeKeyText();
 }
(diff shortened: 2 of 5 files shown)


philip-lamb:
If there are missing localisations in geckoview, we should probably back this change out for v1.2, since it's preferable to have at least the enter key localised correctly rather than having enter/search/go incorrectly localised. Comments? @MortimerGoro @bluemarvin

MortimerGoro:
@philip-lamb the localizations are in Firefox Reality. Geckoview is localization agnostic for that, they just use a enum for the type of action.

MortimerGoro:
I'm ok with delaying this if we don't have the localizations ready. BTW we need to add the a entry for the `space`key translations. I'll go ahead and add it so translators have everything ready

cvan:
> I'm ok with delaying this if we don't have the localizations ready. BTW we need to add a entry for the `space` key translations. I'll go ahead and add it so translators have everything ready

we should probably still land this so the strings are in `strings.xml` ready for localisers. but I'll defer to @philip-lamb, @jvonitter, et. al.

philip-lamb:
> we should probably still land this so the strings are in `strings.xml` ready for localisers. but I'll defer to @philip-lamb, @jvonitter, et. al.

@MortimerGoro read your mind 😄... https://github.com/MozillaReality/FirefoxReality/pull/1184 is landed and we can link the labels once the translations are in.

MortimerGoro:
Ready for review again: The new texts are landed and I fixed the unable to load URLs problem

philip-lamb:
Unblocked as l10ns are in for key locales.

daoshengmu:
Interesting. I can't reproduce this problem from `release-1.2`. One thing I noticed is we are showing a wrong simplified Chinese string in URL bar when typing in Traditional Chinese from @cvan's screenshots. However, on my Oculus Go, it is showing `Search the Web or enter address` that is also wrong when switching to Chinese keyboards.

philip-lamb:
> Interesting. I can't reproduce this problem from `release-1.2`. One thing I noticed is we are showing a wrong simplified Chinese string in URL bar when typing in Traditional Chinese from @cvan's screenshots.

Would you have a look at https://pontoon.mozilla.org/zh-CN/android-l10n/MozillaReality/FirefoxReality/app/src/main/res/values/strings.xml/?string=198958 and see if what is there makes sense.

> However, on my Oculus Go, it is showing `Search the Web or enter address` that is also wrong when switching to Chinese keyboards.

That string is determined by the UI locale, which is not changed by changing the keyboard, so that is not unexpected.

MortimerGoro:
Yes, that string is expected to be fixed to the device locale. It's not specific for the current keyboard. I see the same behavior on other platforms such as iOS and Android.

I'm going to merge this for today's release candidate. @cvan follow-up if you find something else

PR #1184, opened 2019-05-09
Add space key label translation entry

MortimerGoro:


MortimerGoro:
I'm merging this so the it's available for translation work

PR #1189, opened 2019-05-09
Clean compositing text when switching keyboard.

daoshengmu:
I notice composting text are still existing after switching keyboards, we should clean them up to avoid it cause unexpected situations.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/KeyboardWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/KeyboardWidget.java
@@ -252,15 +252,7 @@ public class KeyboardWidget extends UIWidget implements CustomKeyboardView.OnKey
                 ((TextView)mFocusedView).addTextChangedListener(this);
             }
         } else {
-            if (mComposingText.length() > 0 && mInputConnection != null) {
-                mComposingText = "";
-                // Clear composited text when the keyboard is dismissed
-                final InputConnection input = mInputConnection;
-                postInputCommand(() -> {
-                    displayComposingText("");
-                    input.finishComposingText();
-                });
-            }
+            cleanComposingText();
             hideOverlays();
             mInputConnection = null;
         }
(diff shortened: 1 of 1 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/KeyboardWidget.java
```diff
@@ -577,6 +577,9 @@ private void handleGlobeClick() {
     }
 
     private void handleLanguageChange(KeyboardInterface aKeyboard) {
+        mComposingDisplayText = "";
+        mComposingText = "";
+        mInputConnection.finishComposingText();
```
MortimerGoro: mInputConnection method should be called on a different thread. Check the next snippet, can you move it to a function so we don't repeat this code?

```
if (mComposingText.length() > 0 && mInputConnection != null) {
            mComposingText = "";
            // Clear composited text when the keyboard is dismissed
            final InputConnection input = mInputConnection;
            postInputCommand(() -> {
                displayComposingText("");
                input.finishComposingText();
            });
        }
```

daoshengmu: Fixed and also move the one under `updateFoxusedView()` to the new function `cleanCompositingText()`.

app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/KeyboardWidget.java
```diff
@@ -252,15 +252,7 @@ public void updateFocusedView(View aFocusedView) {
...
-            if (mComposingText.length() > 0 && mInputConnection != null) {
-                mComposingText = "";
-                // Clear composited text when the keyboard is dismissed
-                final InputConnection input = mInputConnection;
-                postInputCommand(() -> {
-                    displayComposingText("");
-                    input.finishComposingText();
-                });
-            }
+            cleanCompositingText();;
```
MortimerGoro: nit: double semicolon

app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/KeyboardWidget.java
```diff
@@ -504,6 +496,17 @@ private KeyboardInterface getKeyboardForLocale(@Nullable Locale aLocale) {
         return null;
     }
 
+    private void cleanCompositingText() {
```
MortimerGoro: nit: cleanComposingText

daoshengmu:
> Right now when we change keyboard focus or close the keyboard we discard the composing text instead of finishing it. This PR is committing the partial composing text instead. I think we should use the same behavior in all situations. Which one do you think is better?

I saw both of this two behaviors before, we can use the same behavior and see how users react it.

MortimerGoro:
@cvan I'm merging this, create follow-up issues if you find anything else

PR #1192, opened 2019-05-10
Fix shift icons

### LINKED ISSUE #1188: Shift keys not visible in master
Keyboard shift keys are not visible in master. They can still be clicked and they partially appear.

MortimerGoro:
Fixes #1188


MortimerGoro:
@cvan I'm merging this to unblock the keyboard work, feel free to update with the fixed new assets if there is time do it for the release

Issue #1107, opened 2019-04-22
Update icons with new assets


jvonitter:
## Current Behavior
Some icons have outdated art. 
New assets are available for the following icons:
- [ ] icon-secure
- [ ] URL-text-cursor-carrot
- [ ] URL-text-delete-all
- [ ] icon-window-resize
- [ ] icons-private browsing
- [ ] menu-settings-honeycomb-fill
- [ ] icon-360mode-transportcontrols-back
- [ ] icon-360mode-transportcontrols-rewind10
- [ ] icon-360mode-transportcontrols-pause
- [ ] icon-360mode-transportcontrols-fastforward30
- [ ] icon-360mode-transportcontrols-volume
- [ ] icon-360mode-transportcontrols-volume-mute
- [ ] icon-immersivemode-back
- [ ] icon-immersivemode-resize
- [ ] icon-immersivemode-brightness
- [ ] icon-dialog-microphone
- [ ] icon-keyboard-shift-off
- [ ] icon-dialog-accept

Files are located here: https://drive.google.com/open?id=1xvvz_WbvThSKAW6__JYdRLU5-AvM0c8a

## Expected Behavior
See this document for icons and context: 
[UIS-65_Icon_Placement_&_Names_002.pdf](https://github.com/MozillaReality/FirefoxReality/files/3105118/UIS-65_Icon_Placement_._Names_002.pdf)

```
```

</details>

cvan:
@thenadj:

I noticed the Globe icon (`icon-keyboard-globe.svg`) is different from the [icon currently in use](https://github.com/MozillaReality/FirefoxReality/blob/master/app/src/main/res/drawable/ic_icon_keyboard_globe.xml) ([added](https://github.com/MozillaReality/FirefoxReality/pull/1050/files#diff-368fa8c523d5373507c35f8d84a8a629) in PR #1050).

which should we use, A or B?

![image](https://user-images.githubusercontent.com/203725/57255730-99458f00-7009-11e9-9437-6183b2746122.png)

<hr>

P.S. apologies in advance for not asking all questions at once. if it's easier for to ask + answer all at once, let me know. otherwise, feel free to reply whenever, and I'll update my branch accordingly.

cvan:
done in PR #1168. thanks!

PR #1199, opened 2019-05-10
fix shift key (fixes issue #1188)

cvan:
- fixes issue #1188, including modified shape per new icons provided in issue #1107 and fixed the Shift On and Caps Lock sizes so they don't get smaller and larger, respectively (as it did previously w/ the old icon)
- this properly fixes the keys so we can revert the revert in PR #1192 (thanks, @MortimerGoro)
- verified there are no visual regressions by checking side by side against this branch vs. f0c5f45a (the commit before the regression in c9d06ba, PR #1107)


MortimerGoro:
LGTM, thanks!

PR #1202, opened 2019-05-10
Avoid Infinite restart loop in the CrashReporterService

MortimerGoro:
See #651

@bluemarvin I can still reproduce a case of infinite restart but the restart is not originated from our CrashReporterService. Has some Gecko behavior changed or does it attempt to restart?

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
@@ -159,8 +159,16 @@ public class VRBrowserActivity extends PlatformActivity implements WidgetManager
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
+        // Fix for infinite restart on startup crashes.
+        long lastRestartTime = SettingsStore.getInstance(getBaseContext()).getLatestCrashRestartTime();
+        boolean cancelRestart = lastRestartTime > 0 && (System.currentTimeMillis() - lastRestartTime) < CrashReporterService.MIN_RESTART_INTERVAL_MS;
+        if (cancelRestart) {
+            finish();
+            return;
+        }
+
         // Set a global exception handler as soon as possible
-        GlobalExceptionHandler.register();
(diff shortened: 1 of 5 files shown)


bluemarvin:
Gecko  does not try to restart. If it isn't FxR code it would have to be the system.

bluemarvin:
I pushed a few fixes but need to wait for a response from HTC:
https://community.viveport.com/t5/Vive-Wave-SDK/Proper-way-to-exit-native-application-from-Java/td-p/30790/jump-to/first-unread-message

Over correct way to exit application from Java.

bluemarvin:
I'm stumped. I've played around with this some more and for some reason, even though the crash reporter does not try and restart us. The ActivityMangager does:
`ActivityManager: Start proc 22402:org.mozilla.vrbrowser/u0a222 for activity org.mozilla.vrbrowser/.VRBrowserActivity`

There might be something in our manifest but I don't see anything obvious.

bluemarvin:
Adding:
```

**Comment by MortimerGoro** [comment]
Thanks for the patch, I updated the PR. I tried it and it works ok. 

`Not sure if this is the correct solution.`

I think it's ok, I don't see another solution.

**Comment by bluemarvin** [comment]
I wish we could figure out why FxR gets instantly restarted after a crash by the ActivityMangager. It seems to happen on every platform too.

File: app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java [hunk 1]
```diff
@@ -105,6 +105,7 @@ public void run() {
     static final int GestureSwipeLeft = 0;
     static final int GestureSwipeRight = 1;
     static final int SwipeDelay = 1000; // milliseconds
+    private static final long MIN_RESTART_INTERVAL_MS = 3000;
```
bluemarvin: I wonder if this shouldn't be public somewhere so it is only define in one place?

MortimerGoro: I reused the variable already defined in CrashReporterService

PR #1205, opened 2019-05-10
Adjust Focus touchpad scroll speed.

daoshengmu:
It is still too fast. We should adjust it to `0.7`.

### DIFF
--- a/app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
+++ b/app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
@@ -282,7 +282,8 @@ struct DeviceDelegateWaveVR::State {
 
       if (touchpadTouched) {
         WVR_Axis_t axis = WVR_GetInputAnalogAxis(controller.type, WVR_InputId_Alias1_Touchpad);
-        delegate->SetTouchPosition(controller.index, axis.x, -axis.y);
+        // We are matching touch pad range from {-1, 1} to the Oculus {0, 1}.
+        delegate->SetTouchPosition(controller.index, (axis.x + 1) * 0.5, (-axis.y + 1) * 0.5);
         immersiveAxes[0] = axis.x;
         immersiveAxes[1] = -axis.y;
         controller.touched = true;


app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
```diff
@@ -281,10 +281,11 @@ struct DeviceDelegateWaveVR::State {
...
 
       if (touchpadTouched) {
+        const float speed = 0.7f;
         WVR_Axis_t axis = WVR_GetInputAnalogAxis(controller.type, WVR_InputId_Alias1_Touchpad);
-        delegate->SetTouchPosition(controller.index, axis.x, -axis.y);
-        immersiveAxes[0] = axis.x;
-        immersiveAxes[1] = -axis.y;
+        delegate->SetTouchPosition(controller.index, axis.x * speed, -axis.y * speed);
+        immersiveAxes[0] = axis.x * speed;
+        immersiveAxes[1] = -axis.y * speed;
```
bluemarvin: We should not be adjusting the scale of the immersiveAxes. The raw data should be passed to the page in immersive mode. Also since there is no time component, the `speed` variable is really a just a scale.

daoshengmu:
I have compared the touchpad data range between Oculus VR and WaveVR.

In Oculus VR, the range is `X: {0, +1}, Y: {0, +1}`.
In Wave VR, the range is `X: {-1, +1}, Y:  {+1, -1}`.

We already inverted the y-axis in WaveVR, the remaining thing we need to do is mapping `-1~+1` to `0~+1` range.

app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
```diff
@@ -282,7 +282,7 @@ struct DeviceDelegateWaveVR::State {
 
       if (touchpadTouched) {
         WVR_Axis_t axis = WVR_GetInputAnalogAxis(controller.type, WVR_InputId_Alias1_Touchpad);
-        delegate->SetTouchPosition(controller.index, axis.x, -axis.y);
+        delegate->SetTouchPosition(controller.index, (axis.x + 1) * 0.5, (-axis.y + 1) * 0.5);
```
bluemarvin: Please add a comment on why this adjustment is being made (i.e., we are matching the Oculus touch pad range).

Issue #1212, opened 2019-05-13
Delight-VR player gets stuck on "START" screen because of Delight-VR's User-Agent sniffing


cvan:
## Hardware
**Oculus Go**
master, 1.1.4, 1.1.3
(1.1.2 appears fine: https://github.com/webcompat/web-bugs/issues/29995#issuecomment-491268051)

## Steps to Reproduce

1. From Firefox Reality (tested on the Oculus Go), load a page with [Delight-VR's player](https://delight-vr.com/documentation/getting-started/#embed), such as https://delight-vr.com/documentation/getting-started/#embed.
2. Try to play the video in VR mode.
3. Try to press the START button, but notice it doesn't seem to do anything.

Tested the [same page](https://delight-vr.com/documentation/getting-started/#embed) in Oculus Browser, and it works as expected (because of the different values in  `navigator.userAgent`).

## Current Behavior
From Firefox Reality, Delight-VR incorrectly detects the browser (as a tablet device instead of a mobile-VR browser), which prevents users from being able to click the **START** button to play a video.

## Expected Behavior
Delight-VR videos cannot be played, causing videos to correctly play in VR when pressing the **START** button.

## Possible Solution
Ideal solution is to get Delight-VR to fix [the code in production](https://cdn.delight-vr.com/1.5.2/0.dl8-3e3af913f9d53da7a03b113754466c5ffdddb3cd.js), but until then there's a way we can patch this with a Web Extension (while at the same time allowing us to purge [FxR's User-Agent Override list of origins](https://github.com/MozillaReality/FirefoxReality/blob/master/app/src/main/assets/userAgentOverride.json)).

## Context
Reported WebCompat issue here: https://github.com/webcompat/web-bugs/issues/31025

I did some investigation, and it appears that Delight-VR's JavaScript parses `navigator.userAgent` and when using Firefox Reality (vs. a Chrome-based browser) Delight-VR thinks it's using an "Unknown Tablet" device, when it should be detected as instead a mobile browser with WebVR supported (which it otherwise does).

There are active email threads with Delight-VR. There is already code in place for `Util.isFirefoxReality` and a few other occurrences, but Delight-VR's mobile/tablet detection code needs to be adjusted to support a "mobile" phone.

See https://cdn.delight-vr.com/1.5.2/0.dl8-3e3af913f9d53da7a03b113754466c5ffdddb3cd.js:

<details closed>
<summary>View JS code: mobile detection rules</summary>

```js
[...]

PR #1213, opened 2019-05-13
add content script for Delight-VR's User-Agent sniffing bug (fixes issue #1212)

cvan:
fixes issue #1212

<hr>

to verify, try w/ this branch vs. `master`:

1. launch this URL: https://delight-vr.com/documentation/getting-started/#embed
2. press the play icon and then the **ENTER** button to play the Delight-VR player


app/src/main/assets/web_extensions/webcompat_delightvr/main.js
```diff
@@ -0,0 +1,104 @@
...
+    evt.stopPropagation();
+    target.remove();
+
+    // Inject an inline <script> onto the page to override the value of
+    // `navigator.userAgent` when acccessed by future scripts.
+    const scriptPatch = document.createElement('script');
+    scriptPatch.setAttribute('data-fxr-patch-for', 'ua');
+    scriptPatch.setAttribute('data-fxr-patched', 'ua');
+    scriptPatch.textContent = `
+  // *** Injected by FxR (v1.2) ***
```
bluemarvin: Include the app version might be confusing since we hopefully won't stay on 1.2 as long as we stayed on 1.1.2?

bluemarvin:
It looks like this WebExt scans the file name of every JS file any page tries to load? Are performance implications from doing this scan for every page? How does it affect page load time?

cvan:
> It looks like this WebExt scans the file name of every JS file any page tries to load? Are there performance implications from doing this scan for every page? How does it affect page load time?

glad it went noticed. I will benchmark and get back to you. I wonder if there is any perf mitigation we can do from using `<all_urls>`.

I will look how slow it is vs. the hardcoded `userAgentOverride.json` method vs. this and also the latency added by using `beforescriptexecute`.

app/src/main/assets/web_extensions/webcompat_delightvr/main.js
cvan: will remove version. hopefully this code can be removed soon anyway. will push on Delight-VR to fix.

cvan:
@bluemarvin: until I find a few moments to benchmark this solution, could you take a look at this simpler approach in https://github.com/MozillaReality/FirefoxReality/pull/1216/files? it's obviously less robust, as it won't work for new hosts using Delight-VR JS player we don't know yet about.

cvan:
will close in favour of #1216

PR #1216, opened 2019-05-14
fix User-Agent Override list for Delight-VR (fixes issue #1212; related to PR #1213)

cvan:
- fixes issue #1212
- related to PR #1213


cvan:
note for posterity's sake: unfortunately, even with https://bugzilla.mozilla.org/show_bug.cgi?id=1511281#c14 fixed, the script calls `navigator.userAgent` in the context of the origin using it, so the origin must be in the UA Override list to work.

PR #1218, opened 2019-05-14
Use the same widget texture for all settings screens

### LINKED ISSUE #1197: Go to second layer in Settings, the transition between two screens looks like flashing
## Hardware
HTC Vive Focus Plus
FxR f0c5f45a8d970e0087f08dce6e3d35c381f583c4
From testing partner report.

## Steps to Reproduce
Go to second layer in Settings

## Current Behavior
The transition between two screens looks like flashing

## Expected Behavior
Should not flash.

## Notes

MortimerGoro:
Fixes #1197


MortimerGoro:
@cvan I can't reproduce this in my Vive and Oculus devices

MortimerGoro:
@cvan it seems something related to accessibility or focus indicator. @bluemarvin has seen that outline in NoAPI a lot with different UIs. We can use that to reproduce it in a easier way.

PR #1219, opened 2019-05-14
Fix spanish text for display settings button

MortimerGoro:


app/src/main/res/values-es-rES/strings.xml
```diff
@@ -138,7 +138,7 @@
 
     <!-- This string is used to label a button in the 'Settings' dialog window that, when pressed,
         opens a dialog box that contains display-related settings: window size, pixel density, etc. -->
-    <string name="settings_display">Mostrar</string>
+    <string name="settings_display">Opciones de visualización</string>
```
cvan: aren't these strings auto-generated from Pontoon/the latest L10n system?

@Pike: if so, do we need to update this string in the central source of truth as well?

Pike:
Yeah, we need this in Pontoon. Can you leave that suggestion on https://pontoon.mozilla.org/es-ES/android-l10n/MozillaReality/FirefoxReality/app/src/main/res/values/strings.xml/?search=settings_di&string=197914 ? Simple Spanish just uses "Visualización", which might be better, as shorter?

MortimerGoro:
Ok, I'll add that suggestion to Pontoon.

> Simple Spanish just uses "Visualización", which might be better, as shorter?

That's ok too.

> can't really review either, though I think I found out that it's currently a verb instead of a noun

Yes, it's a verb, the equivalent of `show` in english.

MortimerGoro:
Fixed via https://github.com/MozillaReality/FirefoxReality/pull/1234

PR #1224, opened 2019-05-14
Re-enable Java crash handler

bluemarvin:


bluemarvin:
Now that the CrashHanlder works again, we should remove this.

PR #1239, opened 2019-05-20
Enable 72hz display rate in Oculus Go

### LINKED ISSUE #555: Oculus Go: Please increase framerate to 72Hz
Oculus Go:
By default, the Oculus Go renders at 60fps, which leads to visible flicker.
The frame rate can be increased to 72Hz, to avoid this.
Oculus provides a detailed description here: https://developer.oculus.com/documentation/mobilesdk/latest/concepts/mobile-72hz/
Please increase the default frame rate to 72Hz, or, if this causes performance issues, provide an option to enable and disable 72Hz.

MortimerGoro:
Fixes #555

Oculus has already fixed the SDK problem we had some time ago

### DIFF
--- a/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
+++ b/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
@@ -733,6 +733,17 @@ struct DeviceDelegateOculusVR::State {
     }
   }
 
+  void UpdateDisplayRefreshRate() {
+    if (!ovr || !IsOculusGo()) {
+      return;
+    }
+    if (renderMode == device::RenderMode::StandAlone) {
+      vrapi_SetDisplayRefreshRate(ovr, 72.0f);
+    } else {
+      vrapi_SetDisplayRefreshRate(ovr, 60.0f);
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -733,6 +733,17 @@ struct DeviceDelegateOculusVR::State {
     }
   }
 
+  void UpdateDisplayRefreshRate() {
+    if (!ovr || !IsOculusGo()) {
```
philip-lamb: Do we not need to set this for Quest?

MortimerGoro: We don't need it for Quest because it is always 75hz there

daoshengmu: Why we still make the immersive mode to be 60hz? It seem likes in the immersive mode, we would expect the FPS be higher to kick rAFs as fast as possible.

drewwebster:
I am confused. is 72hz enabled now in firefox reality? I still see flickering in webvr samples

PR #1240, opened 2019-05-20
Update Android components to v0.52

MortimerGoro:


daoshengmu:
@pocmo If we wanna upgrade Android components from v0.31 to v0.52, is there any change in telemetry we need to be careful?

pocmo:
The only change I'm aware of are the switch from `HttpURLConnectionTelemetryClient` to a generic client that uses any `concept-fetch` implementation. And your change for that looks fine. Other than that there's no active development happening, since the telemetry team has been working on a replacement:
https://github.com/mozilla-mobile/android-components/blob/master/components/service/glean/README.md

daoshengmu:
I have verified telemetry works. In the possible future, we need to move to `Glean`.

pocmo:
FYI @Dexterp37: We talked about potentially migrating apps away from the old telemetry library. But back then we could only focus on Fenix. Are there any plans for the existing apps using the old library?

Dexterp37:
> 
> 
> FYI @Dexterp37: We talked about potentially migrating apps away from the old telemetry library. But back then we could only focus on Fenix. Are there any plans for the existing apps using the old library?

Hey! In addition to Fenix, we have integrated Glean in firefox-tv and are waiting for the data to come back for [validation purpose](https://bugzilla.mozilla.org/show_bug.cgi?id=1548819).

Aside for that, I don't know of any plan, on our end, to transition applications from using legacy telemetry to Glean.

@georgf , thoughts?

georgf:
Currently Q2 is still about supporting the Fenix MVP launch with the new Glean library.

Transitioning existing applications is something we can start in Q3, once the Fenix MVP is launched.
By then we'd have the basic Glean features set up & tested end-to-end, so we should be able to start moving existing applications over.

PR #1255, opened 2019-05-23
Enable Wave VR foveated rendering in App. (Fixes #1228)

### LINKED ISSUE #1228: Enable Foveated Rendering in WaveVR
Oculus VR is already available to choose the foveated rendering level in the menu, WaveVR also supports [Foveated-Render API](https://hub.vive.com/storage/app/doc/en-us/WVR_RenderFoveation.html?highlight=foveation) We should consider to enable it on WaveVR too.

Follow the same configuration as we did in OculusVR, although in our current rendering pipeline, we will not get too much help on FVR, but it should be nice to have, and I already have a patch in local.

daoshengmu:
In Wave VR [1], doing foveated rendering by using FBO index from two eye buffers. They seem to have trouble in setting `WVR_RenderFoveation(false);`, so I have to restart the app to make it work although this API is under beta.

@MortimerGoro The changing of foveated rendering level in App is not so obvious in Oculus Go, is my feeling wrong?

Regarding to WebVR foveated rendering, we need to do it at the Gecko side, we can do a follow up research at [Bug 1553956](https://bugzilla.mozilla.org/show_bug.cgi?id=1553956) to see if it is possible to make.

[1] https://hub.vive.com/storage/app/doc/en-us/WVR_RenderFoveation.html

### DIFF
--- a/app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
+++ b/app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
@@ -63,6 +63,7 @@ struct DeviceDelegateWaveVR::State {
   vrb::Color clearColor;
   float near;
   float far;
+  float foveatedFov;
   void* leftTextureQueue;
   void* rightTextureQueue;
   device::RenderMode renderMode;
@@ -74,6 +75,8 @@ struct DeviceDelegateWaveVR::State {
   vrb::CameraEyePtr cameras[2];
   uint32_t renderWidth;
   uint32_t renderHeight;
+  uint32_t standaloneFoveatedLevel;
+
   WVR_DevicePosePair_t devicePairs[WVR_DEVICE_COUNT_LEVEL_1];
   ElbowModelPtr elbow;
   ControllerDelegatePtr delegate;
@@ -90,6 +93,7 @@ struct DeviceDelegateWaveVR::State {
       : isRunning(true)
       , near(0.1f)
       , far(100.f)
+      , foveatedFov(0.0f)
       , renderMode(device::RenderMode::StandAlone)
       , leftFBOIndex(0)
       , rightFBOIndex(0)
@@ -97,6 +101,7 @@ struct DeviceDelegateWaveVR::State {
       , rightTextureQueue(nullptr)
       , renderWidth(0)
       , renderHeight(0)
+      , standaloneFoveatedLevel(0)
       , devicePairs {}
       , controllers {}
       , lastSubmitDiscarded(false)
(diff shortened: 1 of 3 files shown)


app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
```diff
@@ -373,13 +422,20 @@ void
...
+
 void
 DeviceDelegateWaveVR::SetClipPlanes(const float aNear, const float aFar) {
   m.near = aNear;
   m.far = aFar;
   m.InitializeCameras();
 }
 
+void
+DeviceDelegateWaveVR::SetFoveatedLevel(const int32_t aAppLevel, const int32_t aWebVRLevel) {
```
bluemarvin: It doesn't make any sense to pass the WebVR level into the DeviceDelegate. We should probably just remove it from the function call. When it is implemented it will either have to to gecko via the external VR interface.

app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
```diff
@@ -191,6 +202,44 @@ struct DeviceDelegateWaveVR::State {
...
+              static_cast<WVR_PeripheralQuality>(WVR_PeripheralQuality_High-(quality - 1));
+
+      WVR_RenderFoveationParams_t foveated;
+      foveated.focalX = foveated.focalY = 0.0f;
+      foveated.fovealFov = foveatedFov;
+      foveated.periQuality = peripheralQuality;
+      WVR_TextureParams_t eyeTexture = WVR_GetTexture(leftTextureQueue, leftFBOIndex);
+      WVR_PreRenderEye(WVR_Eye_Left, &eyeTexture, &foveated);
+      eyeTexture = WVR_GetTexture(rightTextureQueue, rightFBOIndex);
+      WVR_PreRenderEye(WVR_Eye_Right, &eyeTexture, &foveated);
```
bluemarvin: This doesn't look right. From the documentation, `WVR_PreRenderEye()` needs to be called each time before we render to the texture. This looks like it just gets the current texture and does it once when the level changes. This should probably go in either `DeviceDelegateWaveVR::StartFrame()` or in `DeviceDelegateWaveVR::BindEye()`.

app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
daoshengmu: I just removed the variable for setting WebVR level. I think we would to send the WebVR level from FxR settings to GV via the external VR interface to make WebGL framebuffer to change its foveated level.

app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
daoshengmu: Removed the if-checking of level changes. As you said, we should just set the foveated level when rendering scene. I have already put it at `DeviceDelegateWaveVR::StartFrame()` -> `DeviceDelegateWaveVR::UpdateFoveatedLevel()` at my previous patch, so it should be good.

MortimerGoro:
@bluemarvin Why does taskcluster show a error? I see `BUILD SUCCESSFUL in 1h 15m 21s` in the logs

bluemarvin:
> @bluemarvin Why does taskcluster show a error? I see `BUILD SUCCESSFUL in 1h 15m 21s` in the logs

I think it is because I reran it instead of retriggering it. If you look, the task ran twice and failed on a gradle error the first time. Rerun and retrigger are right next to each other in the menu.

PR #1256, opened 2019-05-23
Add virtual controller to NoAPI immersive mode

bluemarvin:


app/src/noapi/cpp/DeviceDelegateNoAPI.h
```diff
@@ -39,13 +41,16 @@ class DeviceDelegateNoAPI : public DeviceDelegate {
...
+  void InitializeJava(JNIEnv* aEnv, jobject aActivity);
+  void ShutdownJava();
   void SetViewport(const int aWidth, const int aHeight);
   void Pause();
   void Resume();
   void MoveAxis(const float aX, const float aY, const float aZ);
   void RotateHeading(const float aHeading);
   void RotatePitch(const float aPitch);
   void TouchEvent(const bool aDown, const float aX, const float aY);
+  void controllerButtonPressed(const bool aDown);
```
MortimerGoro: nit: Uppercase c in controller

app/src/noapi/cpp/DeviceDelegateNoAPI.cpp
```diff
@@ -154,8 +180,10 @@ DeviceDelegateNoAPI::SetClipPlanes(const float aNear, const float aFar) {
 void
 DeviceDelegateNoAPI::SetControllerDelegate(ControllerDelegatePtr& aController) {
   m.controller = aController;
-  m.controller->CreateController(0, -1, "");
+  m.controller->CreateController(0, -1, "FxR Virtual Controller");
```
MortimerGoro: Should we use a common controller name here for the engines which filter by controller name?

bluemarvin: I was using the Daydream controller name string initially, but was curious what would happen if I used a unique name most  everything I tried worked. But it doesn't work with hubs. so I guess I could use a known string. I tried using "Oculus Touch (Right)" and Hubs actually won't start in FxR NoAPI. I've tried all the different known controller names and none of them really work so I'm inclined to just leave it.

PR #1261, opened 2019-05-24
Ensure UA override works for redirects

bluemarvin:


bluemarvin:
The code removed for this PR was originally for a fix to GeckoView where we would get `onLoadRequest()` for elements in the page such as iframes that would cause UA override to get disabled. That issue has now been fixed in GeckoView. However, that old fix depended on `onPageStop()` to reset the UA override flag. Unfortunately redirects do not call `onPageStop()` so the flag was never getting reset and the UA override was not being checked for redirected pages.

PR #1264, opened 2019-05-24
Bump GV version. Enable MultiView draft extension.

### LINKED ISSUE #1208: Implement webgl2 draft ext OVR_multiview2
## Hardware
All

## Context
Tracking bug for https://bugzilla.mozilla.org/show_bug.cgi?id=1536672

<details open>

```
```

</details>

MortimerGoro:
Fixes #1208


fernandojsg:
Do we really want to have that flag enabled by default? I mean I want multiview for sure, but `webgl-draft-extensions` could lead to some issues on other extensions that are behind that flag and are still experimental isn't it? Could it be possible to just enable `multiview` instead?

PR #1265, opened 2019-05-24
Bump GV version. Fix WebVR texture sharpness issues

### LINKED ISSUE #937: Poor rendering of detail in WebVR scenes
## Hardware
Oculus Go

## Steps to Reproduce
1. Load http://pano.yfway.com/30801042 or a demo from the [krapno's WebVR example](https://krpano.com/webvr/) in Firefox Reality.
2. Navigate to other panos by the gazing on a hotspot, and move your head around to notice the tracking is a bit slow.
4. Open the same site in Oculus Browser.
5. Upon navigating to other panos, notice the tracking is more stable than it is in Firefox Reality.

Originally reported in the [krpano forums](https://krpano.com/forum/wbb/index.php?page=Thread&postID=77699).

## Current Behavior
The tracking is laggy in [krpano-powered WebVR galleries](https://krpano.com/webvr/).

## Expected Behavior
The tracking is not laggy in [krpano-powered WebVR galleries](https://krpano.com/webvr/).

## Context
Originally reported in the [krpano forums](https://krpano.com/forum/wbb/index.php?page=Thread&postID=77699).

MortimerGoro:
Fixes #937

Do not merge yet, I created this to get some initial feedback of this solution. The problem is that the testcase creates a 2x canvas surface but we are still using the default size in our WebVR swapChain. This PR resizes the swapChain based on the received surface size (taking some thresholds into account).

We need a way to get the WebVR surface size received from Gecko in order to call `SetImmersiveSize` with the correct values. OpenGL ES doesn't have an API to get the texture size until 3.1. Anyway I don't like the OpenGL query solution (getters aren't good for performance). Ideally we can include the surface size in the external_vr struct or also in GeckoSurfaceTexture.java

### DIFF
--- a/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
+++ b/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
@@ -1110,6 +1110,30 @@ DeviceDelegateOculusVR::RegisterImmersiveDisplay(ImmersiveDisplayPtr aDisplay) {
   m.UpdatePerspective();
 }
 
+void
+DeviceDelegateOculusVR::SetImmersiveSize(const uint32_t aEyeWidth, const uint32_t aEyeHeight) {
+  uint32_t recommendedWidth, recommendedHeight;
+  m.GetImmersiveRenderSize(recommendedWidth, recommendedHeight);
+
+  const uint32_t maxWidth = recommendedWidth * 2;
+  const uint32_t maxHeight = recommendedHeight * 2;
+  const uint32_t minWidth = recommendedWidth / 2;
+  const uint32_t minHeight = recommendedHeight / 2;
+
+  const uint32_t targetWidth = (uint32_t) fmaxf(fminf(aEyeWidth, maxWidth), minWidth);
+  const uint32_t targetHeight = (uint32_t) fmaxf(fminf(aEyeHeight, maxHeight), minHeight);
+
+  if (targetWidth != m.renderWidth || targetHeight != m.renderHeight) {
+    m.renderWidth = targetWidth;
+    m.renderHeight = targetHeight;
+    vrb::RenderContextPtr render = m.context.lock();
+    for (int i = 0; i < VRAPI_EYE_COUNT; ++i) {
+      m.eyeSwapChains[i]->Init(render, m.renderMode, m.renderWidth, m.renderHeight);
+    }
+    VRB_LOG("Resize immersive mode swapChain: %dx%d", targetWidth, targetHeight);
+  }
+}
+
 vrb::CameraPtr
 DeviceDelegateOculusVR::GetCamera(const device::Eye aWhich) {
   const int32_t index = device::EyeIndex(aWhich);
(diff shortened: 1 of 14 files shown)


daoshengmu:
I tried to print the WebGL frambuffer size of this example [1] from Gecko. At the first creation, it tried to use `300x150`, but it failed, then finally using `1385x779`. The code tells me 
```
attemptSize.width /= 2;
attemptSize.height /= 2;
```

But, it got bigger numbers, I felt surprised. I need to trace the logic later.

At, `DeviceDelegateOculusVR::EnterVR()`, we use `renderWidth(1024) x renderHeight(1024)`, it uses the same buffer size that is shown at [2]. I would like to use the framebuffer size from WebGL to DeviceDelegateOculusVR::EnterVR() via ExternalVR, that should be a right fix.

[1] https://webvr.info/samples/04-simple-mirroring.html
[2] https://webvr.info/samples/00-hello-webvr.html

MortimerGoro:
@daoshengmu 

I get 2048x1024 in the `04-simple-mirroring.html` webvr sample (1024x1024 per eye).

MortimerGoro:
Related GV patch: https://bugzilla.mozilla.org/show_bug.cgi?id=1554626

MortimerGoro:
GV patch already merged and updated in the PR. 

This is ready for the final review

bluemarvin:
@MortimerGoro Wouldn't this be an issue for WaveVR and GoogleVR too?

MortimerGoro:
@bluemarvin true, I forgot about that, I'll update the wave and google backends

daoshengmu:
It looks good to me on OculusVR, I will review it again once a new update comes for other platforms.

MortimerGoro:
Wave and Daydream platforms are ready

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -1110,6 +1110,30 @@ DeviceDelegateOculusVR::RegisterImmersiveDisplay(ImmersiveDisplayPtr aDisplay) {
...
+  m.GetImmersiveRenderSize(recommendedWidth, recommendedHeight);
+
+  const uint32_t maxWidth = recommendedWidth * 2;
+  const uint32_t maxHeight = recommendedHeight * 2;
+  const uint32_t minWidth = recommendedWidth / 2;
+  const uint32_t minHeight = recommendedHeight / 2;
+
+  const uint32_t targetWidth = (uint32_t) fmaxf(fminf(aEyeWidth, maxWidth), minWidth);
+  const uint32_t targetHeight = (uint32_t) fmaxf(fminf(aEyeHeight, maxHeight), minHeight);
+
```
daoshengmu: nit: I am thinking if we can move 
```
   const uint32_t maxWidth = recommendedWidth * 2;
   const uint32_t maxHeight = recommendedHeight * 2;
   const uint32_t minWidth = recommendedWidth / 2;
   const uint32_t minHeight = recommendedHeight / 2;
   const uint32_t targetWidth = (uint32_t) fmaxf(fminf(aEyeWidth, maxWidth), minWidth);
   const uint32_t targetHeight = (uint32_t) fmaxf(fminf(aEyeHeight, maxHeight), minHeight);
```
to class DeviceDelegate as `GetTargetWidth()` and `GetTargetHeight()`, And getting them via `uint32 w = GetTargetWidth(recommendedWidth);`, `uint32 h = GetTargetHeight(recommendedHeight);`  how do you think about it ?

MortimerGoro: Done

bluemarvin:
@MortimerGoro What pages did you use to test this?

bluemarvin:
Specifically, using this site: https://www.tourdeforce360.com/grandtheater/vtour.html

It is 100% reproducible.

MortimerGoro:
@bluemarvin PR updated with the fix for daydream

PR #1272, opened 2019-05-28
Pinyin keyboard fixes

### LINKED ISSUE #1253: minor changes to Pinyin keyboard
Minor updates based on feedback. Latest spec is here:
https://trello.com/c/GeLVSIig/119-uis-27-chinese-keyboard-pinyin

Waiting for one more piece of feedback and then we'll be at final design.

MortimerGoro:
- Add correct pinyin punctuation autocompletion (pinyin comma and dot)
- Improve autocompletion view layouts 

Fixes #1253

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/keyboards/ChinesePinyinKeyboard.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/keyboards/ChinesePinyinKeyboard.java
@@ -12,6 +12,7 @@ import org.mozilla.vrbrowser.input.CustomKeyboard;
 import org.mozilla.vrbrowser.utils.StringUtils;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -27,6 +28,9 @@ public class ChinesePinyinKeyboard extends BaseKeyboard {
     private DBHelper mDB;
     private HashMap<String, KeyMap> mKeymaps = new HashMap<>();
     private HashMap<String, KeyMap> mExtraKeymaps = new HashMap<>();
+    private List<Character> mAutocompleteEndings = Arrays.asList(
+            ' ', '，', '。','!','?','ー'
+    );
 
     public ChinesePinyinKeyboard(Context aContext) {
         super(aContext);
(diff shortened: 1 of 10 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/keyboards/ChinesePinyinKeyboard.java
```diff
@@ -50,7 +50,9 @@ public CandidatesResult getCandidates(String aComposingText) {
         }
 
         // Autocomplete when space is clicked
```
daoshengmu: nit: We should change this comment because we also do the same things on comma and period.

MortimerGoro:
@daoshengmu It's based on the latest final design in #1253. The changes are:

- Add the Chinese comma
- Add the Chinese period
- Add the Chinese long sound symbol
- Updated hover styles for autocompletion items
- Improve autocomplete items size math

I'm missing any extra change? Emojies are not ready yet

app/src/common/shared/org/mozilla/vrbrowser/ui/keyboards/ChinesePinyinKeyboard.java
MortimerGoro: Fixed

app/src/common/shared/org/mozilla/vrbrowser/ui/keyboards/ChinesePinyinKeyboard.java
```diff
@@ -49,7 +49,7 @@ public CandidatesResult getCandidates(String aComposingText) {
             return null;
         }
 
-        // Autocomplete when space is clicked
+        // Autocomplete when special characters are clicked
         boolean autocomponse = aComposingText.endsWith(" ")
                                || aComposingText.endsWith("，")
                                || aComposingText.endsWith("。");
```
daoshengmu: It seems like `!`, `?`, and long sound symbol also need to be added to this rule.

daoshengmu:
org mozilla vrbrowser-20190530-154203
I was trying to type `ko`, and the autocomplete item layout seem to go wrong as above.

MortimerGoro:
I've fixed the extra autocompletion characters and the layout issue.

> The hover style is not the same with the design document. It supposes to be Tangerine color.

I see the tangerine color is not used for all the hovers but only for the  `selected option by spacebar`. Right know the spacebar always selects the first autocompletion option. Is the design assuming we can select another one with spacebar? how is it selected? cc @thenadj @jvonitter could you clarify this?

thenadj:
@MortimerGoro There is no autocomplete navigation for the Pinyin keyboard. The spacebar ONLY selects the first option in the autocomplete. To choose other options, the user has to click characters with their controller. The specs at the bottom of the document are to show different interaction states.

Ping me in Slack if it doesn't make sense.

MortimerGoro:
@thenadj @daoshengmu since there is no autocomplete navigation, should we use the Tangerine color always on the fist autocomplete option? If we don't want the color the PR is ready to merge. If we want the color I'll update the PR

daoshengmu:
It looks good to me now. We just need to know if we want to use the Tangerine color on the fist autocomplete option

thenadj:
Yes, we want to use the tangerine color on the first autocomplete item. It also helps with consistency of other keyboards like Japanese that have this color too.

MortimerGoro:
I added the tangerine color

PR #1273, opened 2019-05-28
Fix unneeded static keyword inside empty namespace warnings

MortimerGoro:


bluemarvin:
I have a large clean up patch that had this as part of it. I hope it is easy to rebase 😅

PR #1278, opened 2019-05-29
Update upload_symbols.sh to use new script for finding native lib

bluemarvin:


tools/taskcluster/upload_symbols.sh
```diff
@@ -1,11 +1,6 @@
 #/bin/sh
-git clone git@github.com:MozillaReality/symbolgenerator.git tools/taskcluster/symbols && cd tools/taskcluster/symbols && git rebase origin/master && cd ../../..
+#git clone git@github.com:MozillaReality/symbolgenerator.git tools/taskcluster/symbols && cd tools/taskcluster/symbols && git rebase origin/master && cd ../../..
```
daoshengmu: It looks like you will not clone the script when doing a new run? Why? and How you get the script at a new image?

bluemarvin: Good catch, I comment it out for testing.

bluemarvin:
As a follow up, I would like to move the cloning into the .taskcluster.yml file. I think we could then move the upload_symbols.sh into the symbolgenerator repo.

PR #1280, opened 2019-05-31
Fix Java crash reporter

bluemarvin:
Java crashes did not have the correct version or build id set and were thus not showing up on crash stats.
Additionally, the infinite loop crash prevention code was bocking the app from restarting after a normal
crash. So now in addition to keeping a time stamp, a count is kept so it is know if FxR has restarted
multiple times in a short period.


bluemarvin:
Fixes #1281 and #1282

PR #1289, opened 2019-05-31
Re-enable R8 in compat mode

bluemarvin:


bluemarvin:
I was able to build with full mode so let's try that for now. We can fall back to compat mode if full mode causes any issues.

PR #1290, opened 2019-05-31
Enable R8 full mode (fixes #1190)

### LINKED ISSUE #1190: Transition from Proguard to R8
R8 is replacing Proguard in Android tools.

We have a number of dependencies on proguard in our build system to manipulate library dependencies.

We should transition to R8, first in compatibility mode, and eventually in native mode.

https://android-developers.googleblog.com/2018/11/r8-new-code-shrinker-from-google-is.html

bluemarvin:


MortimerGoro:
I checked the generatic bytecode and it doesn't include the problematic android.speech dependencies. Is it doing that automatically?  I think that in fullMode it doesn't use the proguard rules

keianhzo:
It's crashing for me in regular and full R8 modes throwing some **ClassNotFoundException** exceptions. Some more Proguard rules tweaking will progrably be necessary to make it work.

Also I get a gradle warning when syncing:

> WARNING: The option setting 'android.enableR8.fullMode=true' is experimental and unsupported.
> The current default is 'false'.

OculusVR arm7 release using
- Proguard: 66,2MB
- R8: 64MB
- R8 Full: 64MB

Making this work might require some more work to debug which new rules we need to add to make it work with R8.

bluemarvin:
Closing since this looks like it will take more work.

PR #1300, opened 2019-06-05
Fixed the Telemetry issues

keianhzo:
Telemetry events were not being sent since last update. Removed setPersisted flag from the Telemetry scheduler jobs as it required the RECEIVE_BOOT_COMPLETED permission

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/telemetry/FxRTelemetryScheduler.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/telemetry/FxRTelemetryScheduler.java
@@ -1,0 +1,46 @@
+package org.mozilla.vrbrowser.telemetry;
+
+import android.app.job.JobInfo;
+import android.app.job.JobScheduler;
+import android.content.ComponentName;
+import android.content.Context;
+
+import org.mozilla.telemetry.config.TelemetryConfiguration;
+import org.mozilla.telemetry.schedule.TelemetryScheduler;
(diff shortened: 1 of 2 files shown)


bluemarvin:
Regressed by #1085

daoshengmu:
@pocmo Could you help us review this patch again? Our FxR telemetry seems to be broken after removing `RECEIVE_BOOT_COMPLETED` permission (we have to because the requirement of the app store ) at #1085. We gave a workaround to make our own `TelemetryScheduler` but it seems still not totally right regarding to the telemetry result. The only difference between this and the current Mozilla AC [1] is we don’t set `setPersisted` .

[1] https://github.com/mozilla-mobile/android-components/blob/4e56e3a1375c4b6e7bcd218a632fa5551add9b34/components/service/telemetry/src/main/java/org/mozilla/telemetry/schedule/jobscheduler/JobSchedulerTelemetryScheduler.java#L37

pocmo:
> Our FxR telemetry seems to be broken after removing RECEIVE_BOOT_COMPLETED permission

What exactly does "broken" mean in this context? Does it never upload anything?

daoshengmu:
> > Our FxR telemetry seems to be broken after removing RECEIVE_BOOT_COMPLETED permission
> 
> What exactly does "broken" mean in this context? Does it never upload anything?

It still uploads by some people but just a few, it looks like only 10% users are able to upload these data.

PR #1304, opened 2019-06-06
Clean up unused WebVR Foveated-Rendering code. (fixes #1275)

### LINKED ISSUE #1275: Stop passing WebVR Foveation level into DeviceDelegate
## Current Behavior

This is a refactor bug and follow up for #1255. We pass WebVR foveated level to DeviceDelegate but we have no solutions to do this in FxR side, we will need to do it in WebGL, so it is unnecessary to pass it to DeviceDelegate.

## Expected Behavior

Remove WebVR foveated level parameter and do it outside of FxR.

```
```

</details>

daoshengmu:
We understand WebVR Foveated-Rendering will be impossible to do in FxR, it has to be in GV. Let's clean up them.

### DIFF
--- a/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
+++ b/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
@@ -616,7 +616,6 @@ struct DeviceDelegateOculusVR::State {
   uint32_t renderWidth = 0;
   uint32_t renderHeight = 0;
   int32_t standaloneFoveatedLevel = 0;
-  int32_t immersiveFoveatedLevel = 0;
   vrb::Color clearColor;
   float near = 0.1f;
   float far = 100.f;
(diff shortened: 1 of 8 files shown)


MortimerGoro:
If we implement straight WebVR render to Oculus swapChain, isn't the WebVR foveated has to bet set up in FxR?

daoshengmu:
We will use the setting in FxR, then asking WebGL to bind a foveated-rendering format color texture to WebGL's framebuffer.

bluemarvin:
@MortimerGoro It isn't clear how we will handle WebVR since it looks like we can use the OGL extension to do it our selves in WebGL. I would rather not have this in the API if it doesn't actually do anything right now. Once we have a better understanding of how we will enable it for WebVR we can then create an API that makes sense then.

MortimerGoro:
Ok, sounds good

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -716,8 +715,6 @@ struct DeviceDelegateOculusVR::State {
     }
     if (renderMode == device::RenderMode::StandAlone) {
       vrapi_SetPropertyInt(&java, VRAPI_FOVEATION_LEVEL, standaloneFoveatedLevel);
-    } else {
-      vrapi_SetPropertyInt(&java, VRAPI_FOVEATION_LEVEL, immersiveFoveatedLevel);
```
MortimerGoro: We should disable foveated here, otherwise both Gecko and FxR foveated can be applied together here (if foveated is enabled for standalone)

daoshengmu: Updated. I would like to make both of immersive and standalone mode use the same `standaloneFoveatedLevel` in OVR for now. I guess when we are ready doing it in GV::WebGL, we would disable foveated here and ask WebGL's framebuffer generate a foveated texture to us according to this `standaloneFoveatedLevel`.

MortimerGoro: Ok, it makes sense. You have removed `vrapi_SetPropertyInt(&java, VRAPI_FOVEATION_LEVEL, standaloneFoveatedLevel);`

PR #1307, opened 2019-06-10
Add separate Oculus 6DOF build target

MortimerGoro:


bluemarvin:
@MortimerGoro My expectation is that 3DoF will eventually cease to be supported. Would it make more sense to name the build targets `oculusvr3Dof` for the Go and `oculusvr` for the Quest and all future 6Dof devices?

MortimerGoro:
@bluemarvin it makes sense. PR already updated

bluemarvin:
Hm, for some reason oculusvrArm64 isn't building. I'll look into the script.

PR #1314, opened 2019-06-12
Automatically select better Vimeo quality and make fullscreen visible

MortimerGoro:
- Automatically select better Vimeo quality by default
- Automatically scroll viewport to make full screen icon visible by default.


MortimerGoro:
Example testing URL: https://vimeo.com/307317753

PR #1337, opened 2019-06-18
Voice search permission fix

### LINKED ISSUE #1336: Voice microphone permission cannot be rejected
## Hardware
All

## Steps to Reproduce
1. Fresh FxR install
2. Click on the microphone icon
3. System microphone permission should show
4. Click on "Deny"

## Current Behavior
If you deny the permission the permission dialog keeps on popping up.

## Expected Behavior
The permission should be denied and the voice search should show an error message instead of the voice search.

## Context
This is a regression, we already have a specific message for handling the permission denied case.

<details open>

```
```

</details>

keianhzo:
Fixes #1336 Microphone permission loop when the permission is rejected.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/dialogs/VoiceSearchWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/dialogs/VoiceSearchWidget.java
@@ -261,7 +261,7 @@ public class VoiceSearchWidget extends UIDialog implements WidgetManagerDelegate
                 show();
 
             } else {
-                super.show();
+                super.show(true);
                 setPermissionNotGranted();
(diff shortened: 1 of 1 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/dialogs/VoiceSearchWidget.java
```diff
@@ -261,7 +261,7 @@ public void onRequestPermissionsResult(int requestCode, String[] permissions, in
                 show();
 
             } else {
-                super.show();
+                super.show(true);
```
MortimerGoro: We may want to change the true to a enum in the future to improve readability (as a follow-up)

PR #1339, opened 2019-06-18
Fixes radio button hover state

### LINKED ISSUE #1196: Radio button should hover state when raycast pointed on
## Hardware
HTC Vive Focus Plus
FxR f0c5f45a8d970e0087f08dce6e3d35c381f583c4
From testing partner report.

## Steps to Reproduce
Hover pointer over radio button (in settings UI? or in webpage? Not clear from report).

## Current Behavior
Radio button does not hover state when raycast pointed on

## Expected Behavior
Radio button should hover state when raycast pointed on

## Notes
If this report refers to web content, what is the expected behaviour? Same question for Android UI. We can close as `invalid` if what we do is expected behaviour.

keianhzo:
Fixes #1196 Fixes radio button hover state styles


keianhzo:
I found the new spec a bit confusing though, when you check an option as the pointer is on top of the radio button it shows the grey hovered state instead of the checked state which might confuse the user.

jvonitter:
Thanks Manuel! Can we pass an apk to Nadja and team for them to see what you mean about the button state?

keianhzo:
@thenadj You can test the spec implementation in this APK: https://send.firefox.com/download/2f0aedb434403033/#G-whCzmp4Yt7h14j6tWvZg

Please see my comments above. Thanks!

thenadj:
@keianhzo I agree, the states were not specd correctly. Attached is the updated version. I'll also add it to [GH-1196](https://github.com/MozillaReality/FirefoxReality/issues/1196)

[UIS-66 Toggle and Radio Button States_003.pdf](https://github.com/MozillaReality/FirefoxReality/files/3312109/UIS-66.Toggle.and.Radio.Button.States_003.pdf)

keianhzo:
@thenadj Updated APK: https://send.firefox.com/download/7ba3bad5e72b8948/#Z4U2dAGtd1_BaGgRpEBHYQ

thenadj:
@keianhzo Thanks for updating the radio buttons. They look great now.

MortimerGoro:
I find the hover color for checked radio button confusing. It feels like the button is not checked when hovered

thenadj:
Ok, let's add 'hover-select' state. See updates in attached UIS-66.
[UIS-66 Toggle and Radio Button States_004.pdf](https://github.com/MozillaReality/FirefoxReality/files/3326621/UIS-66.Toggle.and.Radio.Button.States_004.pdf)

@jvonitter  
UIS-66 also includes hover for the toggle controls. Is this work tracked somewhere?

keianhzo:
@thenadj @MortimerGoro Updated according to the latest spec, also added the Switch spec as I couldn't find any open issue for that. 

https://send.firefox.com/download/ae9485dfaf47b9b2/#C7W1IKwQQ2wlNPAw3u3OUw

bluemarvin:
I don't see the color change on hover for off state of the switch.

keianhzo:
@bluemarvin I think the hover color difference between checked-hover and unchecked-hover is too subtle, do you mean that or you just don't see any change when you hover on an unchecked option and it just remains unchecked?

bluemarvin:
> @bluemarvin I think the hover color difference between checked-hover and unchecked-hover is too subtle, do you mean that or you just don't see any change when you hover on an unchecked option and it just remains unchecked?

I just don't see a difference between the off hover color red and the off selected state. They look exactly the same to me. So either it isn't working or they are too close to the same shade of red.

bluemarvin:
The off switch color should be done in follow up.

PR #1348, opened 2019-06-27
Remove camera permission in Oculus Go and Quest builds

### LINKED ISSUE #1345: Don't request camera permissions (either in manifest or in-app) on Oculus VR.
## Hardware
Oculus Quest
Oculus Go

## Current Behavior
* Current `app/src/main/AndroidManifest.xml` includes: `<uses-permission android:name="android.permission.CAMERA" />` and this shows on launch on Quest:
OFR-16
* Security and privacy pane includes prompts to allow camera permissions.
* Web pages requesting camera access are not currently automatically denied access on Oculus Go and Quest platforms.

## Expected Behavior
From Quest guidelines:
> "[VRC.Quest.Security.2] The app must request the minimum number of permissions required to function." 

## Possible Solution
On Oculus builds, remove camera permission from manifest, from security prefs, and deny webpages access to camera automatically.

## Context
https://developer.oculus.com/distribute/latest/concepts/vrc-quest-security-2/

keianhzo:
Fixes #1345 Added a tag in the Oculus manifest to remove inherited camera permission from the parent manifest. Also removed it from the Privacy options just for oculus builds.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/settings/PrivacyOptionsView.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/settings/PrivacyOptionsView.java
@@ -8,10 +8,12 @@ package org.mozilla.vrbrowser.ui.widgets.settings;
 import android.Manifest;
 import android.content.Context;
 import android.util.Pair;
+import android.view.View;
 import android.widget.ScrollView;
 import android.widget.TextView;
 
 import org.mozilla.geckoview.GeckoSession;
+import org.mozilla.vrbrowser.BuildConfig;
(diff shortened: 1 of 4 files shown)


bluemarvin:
I see the black screen on the Go. Asking if we can do anything in GV to prevent it.

bluemarvin:
Wasn't a GeckoView issue. But it turns out we already had to fix this for Wave. I added the fix to oculusvr so we won't get a flash.

keianhzo:
@bluemarvin Works fine for the camera permission but still flickers for any other permissions. As it seems Android doesn't offer an API to check if a permission has been blocked before requesting it, I'll land this and file a bug to address this in the future.

PR #1349, opened 2019-06-27
Perform Oculus Platform entitlement checks

MortimerGoro:


app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -686,16 +692,35 @@ struct DeviceDelegateOculusVR::State {
...
       VRB_DEBUG("Detected Oculus Quest");
       deviceType = device::OculusQuest;
+    } else if ((type >= VRAPI_DEVICE_TYPE_GEARVR_START) && (type <= VRAPI_DEVICE_TYPE_GEARVR_END)) {
+      VRB_DEBUG("Detected Gear VR");
+      appId = OCULUS_3DOF_APP_ID;
     } else {
       VRB_DEBUG("Detected Unknown Oculus device");
     }
+
+    ovrRequest result = ovr_PlatformInitializeAndroidAsynchronous(appId, java.ActivityObject, java.Env);
```
bluemarvin: I wonder if it wouldn't be better to only include the OVR lib and API calls in the store builds?

MortimerGoro: I thought about that but my concern was that if there is a related bug to that Asynchronous load we can detect it earlier on dev builds

bluemarvin: But I thought you said it was failing already in local builds? I must be confused.

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -1217,7 +1243,40 @@ DeviceDelegateOculusVR::SetCPULevel(const device::CPULevel aLevel) {
...
+  while ((message = ovr_PopMessage()) != nullptr) {
+    switch (ovr_Message_GetType(message)) {
+      case ovrMessage_PlatformInitializeAndroidAsynchronous: {
+        ovrPlatformInitializeHandle handle = ovr_Message_GetPlatformInitialize(message);
+        ovrPlatformInitializeResult result = ovr_PlatformInitialize_GetResult(handle);
+        if (result == ovrPlatformInitialize_Success) {
+          VRB_DEBUG("OVR Platform Initialized.");
+        } else {
+          VRB_ERROR("OVR Platform Initialize failed: %s", ovrPlatformInitializeResult_ToString(result));
+          VRBrowser::HaltActivity(0);
```
MortimerGoro: missing `#if STORE_BUILD == 1` here

MortimerGoro: I'll update the Pr with the fix

app/src/main/res/values/strings.xml
```diff
@@ -685,5 +685,9 @@
     <string name="no_internet_title">Internet Connection Lost</string>
     <!-- This string is displayed in the message of the notification message shown when the device loses internet connection. -->
     <string name="no_internet_message">Check your device configuration to fix the problem.</string>
-
+    <!-- This string is displayed in the title of the alert dialog shown when the user does not have permission to run the application on the device. -->
+    <string name="not_entitled_title">Unable To Run</string>
+    <!-- This string is displaye in the message body of the alert dialog shown when the user does not have permission to run the application on the device.
+         '%1$s' will be replace at runtime with the app's name. -->
+    <string name="not_entitled_message">%1$s does not have permission to run on this device.</string>
```
jvonitter: Update the explanation string for Unable to Run to say:
"%1$s does not have permission to run on this device and will now exit."

PR #1352, opened 2019-07-01
User enum in widget show

### LINKED ISSUE #1346: Use enums in widget show method to improve readability
We have two show methods in UIWidget and one of them uses a boolean parameter in case the view needs to request the focus. It is confusing when reading code what the view is doing in each one of those cases. 

Consolidate those methods into a single show method accepting an enum that describes if the view needs to request the view focus.

An enum is good enough for the moment.

**Additional context**

keianhzo:
Fixes #1346 Refactored the show methods from UIWidget into a single method that uses a enum for readability.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/SuggestionsWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/SuggestionsWidget.java
@@ -113,8 +113,8 @@ public class SuggestionsWidget extends UIWidget implements WidgetManagerDelegate
     }
 
     @Override
-    public void show() {
-        super.show(false);
+    public void show(@ShowFlags int aShowFlags) {
+        super.show(CLEAR_FOCUS);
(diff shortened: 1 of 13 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/SuggestionsWidget.java
```diff
@@ -113,8 +113,8 @@ protected void initializeWidgetPlacement(WidgetPlacement aPlacement) {
     }
 
     @Override
-    public void show() {
-        super.show(false);
+    public void show(@ShowFlags int aShowFlags) {
+        super.show(CLEAR_FOCUS);
```
bluemarvin: Shouldn't this be `super.show(aShowFlags);`?

keianhzo: Should be ok. show(false) focus parameter should be equivalent to show(CLEAR_FOCUS)

PR #1359, opened 2019-07-03
Re-enable  RECEIVE_BOOT_COMPLETED permission on all platforms but Quest.

MortimerGoro:


MortimerGoro:
I didn't find a better way to remove the permission for different flavors without duplicating the manifest. Unfortunately the manifest merger doesn't support variables in `tools:node=""` or empty name in permission remove node. So it includes a duplicate `<uses-permission android:name="android.permission.CAMERA" tools:node="remove"/>`line but the manifest merges processes that correctly

app/build.gradle
```diff
@@ -155,7 +155,7 @@ android {
                     arguments "-DVR_SDK_LIB=oculusvr-lib", "-DVR_SDK_EXTRA_LIB=ovrplatform-lib", "-DOCULUSVR=ON"
                 }
             }
-            manifestPlaceholders = [ headtrackingRequired:"false" ]
+            manifestPlaceholders = [ headtrackingRequired:"false", permissionToRemove:"android.permission.CAMERA" ]
```
bluemarvin: Isn't this already done in `app/src/oculusvrArmDebug/AndroidManifest.xml`?

MortimerGoro:
I added a new commit because we should use a different scheduler for RECEIVE_BOOT_COMPLETED permission scenario

Issue #1354, opened 2019-07-02
Focus Plus controller orientation is not the same with Vive


daoshengmu:
## Hardware

Vive Focus Plus with v.1.63.623.1

## Steps to Reproduce

1. Open FxR and click the system button on controllers, comparing the orientation between in launcher and FxR.

## Current Behavior

There is about 30 degree difference in pitch.

## Expected Behavior

We should let FxR controller orientation to be the same with Vive's.

## Possible Solution

Adjust the transform matrix, replace controller model with a new one, or take a look at if Wave SDK provides new controller APIs to help us get controller model's information.

## Error Logs and Stack Traces

<details open>

```
```

</details>

PR #1363, opened 2019-07-03
Transfrom controller pitch angle to match Focus controllers in system.

daoshengmu:
Per #1354, we want to make Focus Plus controller's position match the controllers in system.


daoshengmu:
I can't find a way to get their system version to decide if we need to make this change on their previous system. I might need to get assistance from the partner.

bluemarvin:
> I can't find a way to get their system version to decide if we need to make this change on their previous system. I might need to get assistance from the partner.

That's probably fine. I doubt there will be very many people running the old system.

PR #1379, opened 2019-07-09
Do not resize fullscreen window size if the window size is actually larger than the threshold.

### LINKED ISSUE #1362: Full screen transitions in Hubs do not work as expected
## Hardware
Oculus Quest

## Steps to Reproduce
- Set your screen size in FxR to 3x
- Go to hubs.mozilla.com and create an enter a room
- When in VR, hit the "magic wand" button in the HUD to open the media browser
- Go through the flow to show the media browser full screen by clicking on Continue
- Notice the content is clipped and the viewport is reset to 1x

## Current Behavior
- When an app calls the fullscreen API, it appears the CSS layout is happening at the user's viewport size (1x-3x) but the actual screen quad is reset to 1x. This results in clipping of the content.

## Expected Behavior
- FxR should *either*
  - Restore the user's screen size properly when entering full screen mode, and apply the proper CSS layout
  - Always set the screen size to 3x in full screen mode (I would suggest this option, since it provides the best analog to "full screen" in an immersive browser, and works nicely this way in Oculus Browser)

## Possible Solution
- When an application requests full screen support, change the size to 3x and fade the background, regardless of the user's non-full screen viewport settings.

## Context
We rely upon this behavior to provide nice, full UX for browsing, etc.

```
```

</details>

MortimerGoro:
Fixes #1362

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/NavigationBarWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/NavigationBarWidget.java
@@ -379,13 +379,19 @@ public class NavigationBarWidget extends UIWidget implements GeckoSession.Naviga
 
     private void setFullScreenSize() {
         mPlacementBeforeResize.copyFrom(mWindowWidget.getPlacement());
+        final float minScale = WidgetPlacement.floatDimension(getContext(), R.dimen.window_fullscreen_min_scale);
         // Set browser fullscreen size
         float aspect = SettingsStore.getInstance(getContext()).getWindowAspect();
         Media media = SessionStore.get().getFullScreenVideo();
         if (media != null && media.getWidth() > 0 && media.getHeight() > 0) {
             aspect = (float)media.getWidth() / (float)media.getHeight();
         }
-        mWindowWidget.resizeByMultiplier(aspect,1.75f);
+        float scale = mWindowWidget.getCurrentScale();
+        // Enforce min fullscreen size.
+        // If current window area is larger only resize if the aspect changes (e.g. media).
+        if (scale < minScale || aspect != mWindowWidget.getCurrentAspect()) {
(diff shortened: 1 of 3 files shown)


bluemarvin:
I'm confused what the issue is and what the expected behavior should be. I feel like there should be two screen sizes. One for normal browsing and one for full screen. But is seems to me that no mater what size I set for the full screen window, it reverts to the standard window size next time? I also could not reproduce the clipping described in the STR for the issue. Am I missing a step?

MortimerGoro:
I also see a problem outside of the Hubs layout issue. STR:

- Set the web size to 3X
- Load https://davidwalsh.name/demo/fullscreen.php
- Enter Fullscreen. The window size reduces to 1.75x.

Isn't it weird that fullscreen reduces the actual window size? If a user wants a youtube fullscreen video in 3X he can't because of this resize.

cc @jvonitter This may require design decision.

bluemarvin:
I agree that our window sizing is confusing. It rarely behaves the way I would expect and doesn't seem to retain the size I set it to.

jvonitter:
thanks for the heads up @MortimerGoro . I'll check in with UX about it

thenadj:
Looking at it and need some time to reply... expect answers early next week.

jvonitter:
@thenadj and I just discussed this and we think you're right, sizing the screen down like that is weird. What we'd like to do is update the logic to be:
If you enter full screen with a view that is 1.75x or smaller, it enlarges to 1.75x. 
Anything bigger than that size doesn't change size.

Sound ok?

MortimerGoro:
```
I just discussed this and we think you're right, sizing the screen down like that is weird. 
What we'd like to do is update the logic to be:
If you enter full screen with a view that is 1.75x or smaller, it enlarges to 1.75x.
Anything bigger than that size doesn't change size.
```
Thanks! that logic is what the PR is supposed to do.

PR #1389, opened 2019-07-12
WIP: FxA, Sync and Send Tab integrations

grigoryk:
This PR integrates latest version of the FxA account manager from android-components, and adds just enough code to allow signing-in via settings, signing out, synchronizing history & bookmarks and receiving tabs sent from other Firefox devices.

Follow-up:
- #1420 Enable glean telemetry support (so that we know how sync is doing)
- #1419 Obtain a production FxA client ID (currently using a sample one)
- #1397 megazord configuration for places and fxa libs
- #1395 bookmarks UI needs folder support
- #1398 better account management UI, currently there are just sign-in/sign-out/reconnect buttons

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -401,9 +401,13 @@ dependencies {
     implementation deps.android_components.browser_search
     implementation deps.android_components.browser_storage
     implementation deps.android_components.browser_domains
+    implementation deps.android_components.service_accounts
     implementation deps.android_components.ui_autocomplete
     implementation deps.android_components.concept_fetch
     implementation deps.android_components.lib_fetch
+
+    // TODO this should not be necessary at all, see Services.kt
+    implementation deps.work.runtime
 
     // Kotlin dependency
     implementation deps.kotlin.stdlib
--- a/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
@@ -29,6 +29,7 @@ import android.view.View;
 import android.view.ViewTreeObserver;
 import android.widget.FrameLayout;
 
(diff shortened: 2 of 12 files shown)


bluemarvin:
FYI, QR code probably isn’t feasible on most headsets due to limited or no camera support. A PIN number would probably be best.

keianhzo:
@grigoryk It seems that when starting the app we are getting a null account as it seems like the accountmanager initAsync hasn't finished yet so you have to pause/resume the activity to actually start getting events as the the corroutine has already finished.

keianhzo:
@jvonitter This will require some disgn and product decision making before merging, as @grigoryk said, at least bookmarks folders support and a sync button for an initial v1 with manual polling. Also finding the right place in the settings for the sign-in button.

grigoryk:
Updated this PR:
- bookmarks are now displayed right after signing-in
- account password changes (and other 'need to reconnect') events are now handled; there's now a "Reconnect" button
- some further cleanup and bugfixes

Started on bookmarks navigation improvement work in #1396.

grigoryk:
Filed a few follow-up issues that can be handled separately from this PR (see https://github.com/MozillaReality/FirefoxReality/pull/1389#issue-296890586 for a list).

grigoryk:
grigoryk:combinedFxaDesktopBookmarks has the rebased versions of this PR on top of #1395.

MortimerGoro:
great job @grigoryk! I tried `grigoryk:combinedFxaDesktopBookmarks` branch, I was able to log in on the first attempt and bookmark Sync worked perfectly.

Regarding `Send tab to device`, I have to restart FxR to detect it. Is there and event to receive it live?

grigoryk:
Filed some more follow-up tickets. Closing this, work will land on a feature branch in #1417.

PR #1396, opened 2019-07-13
Bookmark UI: folder navigation and "desktop bookmarks" support

### LINKED ISSUE #1395: Better bookmarks UI: navigation
Currently bookmarks library view displays the `mobile` root. #1389 introducing bookmark syncing, which brings an entirety mobile and desktop bookmarks from other Firefox devices, and at that point the current UI is no longer sufficient.

Things that will need to happen here:
-  proper treatment of separators - display in the UI, ignore clicks
- add synthetic `Desktop Bookmarks` folder if logged into `Sync` (or if there are desktop bookmarks around)
- handle folder navigation
- display human-friendly names for the set of default desktop folders

Fenix's codebase could be a good starting point here (if a bit convoluted..).
See: https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt and https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/ext/BookmarkNode.kt#L16

grigoryk:
This is ready for review. I've only tested this on a Pixel 3, not a real VR headset though.

Added/changed in this PR:
- moved BookmarksView to use Kotlin
- added folder navigation
- basic treatment of separators - ignore user clicks on them.
- add synthetic `Desktop Bookmarks` folder, which displays "desktop roots" - "toolbar", "menu" and "other" folders
  - these aren't allowed to be deleted
  - custom, localized names are used

Should probably also display folder icons.

### screenshots
Screenshot_20190714-030146
Screenshot_20190714-030158
Screenshot_20190714-030139


MortimerGoro:
I tested this with the `combinedFxaDesktopBookmarks` branch and worked really well on the Oculus Go.

grigoryk:
This will land on a feature branch as part of #1417.

PR #1401, opened 2019-07-16
Add support for performance monitor

bluemarvin:


app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
```diff
@@ -862,6 +868,9 @@ public void dismiss() {
     @SuppressWarnings("unused")
     private void handlePoorPerformance() {
         runOnUiThread(() -> {
+            if (!sPerformanceMonitorEnabled) {
```
keianhzo: I think we could avoid all the static code in the activity just using here: 
`if (!mSettings.isPerformanceMonitorEnabled()) {`

app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/settings/DeveloperOptionsView.java
```diff
@@ -200,6 +210,18 @@ private void setMultiprocess(boolean value, boolean doApply) {
...
     }
 
+    private void setPerformance(boolean value, boolean doApply) {
+        mPerformanceSwitch.setOnCheckedChangeListener(null);
+        mPerformanceSwitch.setValue(value, false);
+        mPerformanceSwitch.setOnCheckedChangeListener(mPerformanceListener);
+
+        SettingsStore.getInstance(getContext()).setPerformanceMonitorEnabled(value);
+
+        if (doApply) {
```
keianhzo: This block could probably be removed to if calling setting directly from the activity.

keianhzo: You might also want to rebase with https://github.com/MozillaReality/FirefoxReality/pull/1403.

bluemarvin:
> Other than that works great, just miss a way to manage the whitelist but I guess that can wait.

The white list is only per-run and is not saved. I figure people can just turn the performance monitor off if they want. We could maybe add a button to clear the white list but I don't know that it is worth the effort.

bluemarvin:
I think I've addressed all review comments.

PR #1405, opened 2019-07-16
Bump a-c dependency to 4.0.0

grigoryk:
Starting to split-off parts of the FxA/sync work into PRs that can land on `master` now, and PRs that will land on a feature branch.

4.0.0 is the latest version of `android-components`. By extension, this gets us a very recent version of application-services dependencies, as well.

In the future we may consider using the -SNAPSHOT variant (essentially, a nightly version)
to keep up with the latest changes in a-c and a-s.


grigoryk:
I didn't observe any regressions while testing this as part of #1395 and #1389.

bluemarvin:
@MortimerGoro and @keianhzo Probably only needs one review, just who ever can get to it first.

PR #1417, opened 2019-07-17
Combined PR: FxA+Sync+Send Tab integration & Bookmarks navigation

grigoryk:
First `feature/fxa` PR!

Combination of #1389 rebased on top of #1395, and both rebased on top of current `master` now that #1406 and #1405 have landed.

Issue #1394, opened 2019-07-13
syncing logins


grigoryk:
As a follow-up to https://github.com/MozillaReality/FirefoxReality/pull/1389 & #717.

Once that #1389 lands, it's fairly trivial to also sync logins. However, they need to be exposed somehow. There are two main avenues here:
- auto-fill support
- basic "password manager" UI, which lets people search and copy/paste passwords

Ideally, both are added at some point. Copy/paste support is tracked in #189.

keianhzo:
@emstanley Maybe we should move this to v12 when Autofill is ready as it doesn't make much sense to show the logins SyncEngine without support for actually exposing the auto fill in the UI.

PR #1418, opened 2019-07-18
WIP Megazord support

grigoryk:
This is meant to land on top of #1417, so pertinent commits here are the last two. The rest are just #1417.

Currently merging this is blocked on https://github.com/mozilla-mobile/android-components/pull/3806 landing.


grigoryk:
Once mozilla-mobile/android-components#3806 lands (and by then #1417, hopefully), I'll push an updated version of this PR.

grigoryk:
Hmm, that kotlin conversion _may_ make it tricky to rebase `feature/fxa`...

We should consider just landing the megazord stuff on `master`. One downside is that we'll increase the apk size, as the default megazord comes with `fxaclient` and `logins` libraries - and `master` isn't using them yet. But, with #1417 and work for #1394 we'll end up using all of these libraries, anyway.

bluemarvin:
@grigoryk Does this need to be part of the FxA feature branch?

PR #1425, opened 2019-07-18
Added support for bookmarks and history in the awesome bar

### LINKED ISSUE #1257: Show history and bookmarks in awesomebar results
needs #1141

keianhzo:
Fixes #1257 Added support for showing bookmarks and history items inside the awesome bar.


app/src/common/shared/org/mozilla/vrbrowser/search/SearchEngineWrapper.java
```diff
@@ -102,13 +104,14 @@ private String getSuggestionURL(String aQuery) {
         return mSearchEngine.buildSuggestionsURL(aQuery);
     }
 
-    public void getSuggestions(String aQuery, SuggestionsDelegate delegate) {
+    public CompletableFuture<List<String>> getSuggestions(String aQuery) {
+        CompletableFuture<List<String>> future = new CompletableFuture<>();
         // TODO: Use mSuggestionsClient.getSuggestions when fixed in browser-search.
```
bluemarvin: Looks like this comment can be removed?

keianhzo: Ideally we should use that method but I couldn't make it work using the current Kotlin-Java interop for calling suspend coroutines. I think it's ok to leave and I'll take a deeper look or maybe ask the AC team, they might know if there is a way.

bluemarvin: Okay, I saw `SuggestionsClient.getSuggestions() and assumed we were now calling what the `TODO` was highlighting.

Issue #1408, opened 2019-07-17
On Oculus Quest: 6 stale frames per second


kennardconsulting:
## Hardware
Oculus Quest

## Steps to Reproduce

**First:**

* Visit http://moonrider.xyz using the Oculus Browser on Quest
* Click to enter VR (glasses icon in bottom right corner)
* Bend your knees. Move your whole torso from left to right (i.e. like a 'strafing' action, not turning your head)
* Notice the edges of things like the 'Start' button and the grey star you are standing on (look down). They move in a choppy way.

**Then Compare:**

* Open Beat Saber demo (native app) and start tutorial
* Move to the corner of the black square you are standing on (with the white edge) and look down
* Repeat same left/right strafing movement
* Notice the white edges of the square. Notice movement is not choppy.

**Notes**

Is not Moon Rider specific. Is easily seen on most AFrame demos, such as Anime-UI (https://aframe.io/aframe/examples/showcase/anime-UI) and Towers

I believe this is a bug in the positional tracking. UPDATE: a bug related to stale frames (see below)

## Current Behavior
Positional tracking is choppy

## Expected Behavior
Positional tracking is as good as native apps (such as Beat Saber, and the Firefox Reality UI)

## Possible Solution
This issue also affects the Oculus Browser. But it does *not* affect Exokit Native running a WebVR app.

## Context
​It has a significant impact on comfort for anything less than 2 metres away.

## Error Logs and Stack Traces

Other reports:

https://www.reddit.com/r/OculusQuest/comments/c6p3d5/jitter_judder_stutter_in_the_oculus_browser/?utm_source=share&utm_medium=ios_app

<details open>

```
```

</details>

philip-lamb:
@fernandojsg This sounds like a content bug and it mentions A-Frame. Do you have any comment?

philip-lamb:
/remind me in 2 days close if no response.

kennardconsulting:
Thanks for your response!

I do not believe this is a content bug, since the same content is choppy on Oculus Browser and Firefox Reality, but runs smooth on Exokit native. See:

"I tried https://webvr.info/samples/03-vr-presentation.html on Quest Oculus Browser and I am seeing visual frame misses... I tried the same code in Exokit native and it's buttery smooth"
https://github.com/aframevr/aframe/issues/4222#issuecomment-509728422

kennardconsulting:
I don't think it's content related, because it works in Exokit Native. Please don't close :)

[...]

Issue #304, opened 2018-08-02
Oculus optimization: Render straight to an Android SurfaceTexture based texture swap chain


MortimerGoro:
The new Oculus SDK includes and API to create an Android SurfaceTexture based texture swap chain (`vrapi_CreateAndroidSurfaceSwapChain`). We could optimize the render path in the future to render WebVR straight to the swap chain and remove some texture copys.

MortimerGoro:
Hopefully other SDKs will allow this in the future too

philip-lamb:
@MortimerGoro I believe this is already implemented at https://github.com/MozillaReality/FirefoxReality/blob/master/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp#L291 Please close if so.

philip-lamb:
/remind @MortimerGoro in 1 day Check status and close or update.

MortimerGoro:
@philip-lamb We have implemented this for the browser chrome. We also could try to implement it in WebVR (currently we receive a WebGL texture from Gecko, which creates it's own surface). We can create another issue for that or make this one more explicit about WebVR, whichever you prefer.

philip-lamb:
Tracking activity on this in https://bugzilla.mozilla.org/show_bug.cgi?id=1537961

PR #1429, opened 2019-07-19
Render WebVR frames to OVR VR surfaces

daoshengmu:
I put my Gecko work at https://bugzilla.mozilla.org/show_bug.cgi?id=1562134.

### DIFF
--- a/app/src/main/cpp/BrowserWorld.cpp
+++ b/app/src/main/cpp/BrowserWorld.cpp
@@ -1299,6 +1299,7 @@ void
 BrowserWorld::DrawWorld() {
   m.externalVR->SetCompositorEnabled(true);
   m.device->SetRenderMode(device::RenderMode::StandAlone);
+  m.device->EnableExternalSurfaceRender(false);
   if (m.fadeAnimation) {
     m.fadeAnimation->UpdateAnimation();
   }
@@ -1363,23 +1364,29 @@ BrowserWorld::DrawImmersive() {
   m.device->StartFrame();
   VRB_GL_CHECK(glDepthMask(GL_FALSE));
   m.externalVR->PushFramePoses(m.device->GetHeadTransform(), m.controllers->GetControllers(), m.context->GetTimestamp());
+  mozilla::gfx::VRLayerTextureType surfaceType;
   int32_t surfaceHandle, textureWidth, textureHeight = 0;
   device::EyeRect leftEye, rightEye;
   bool aDiscardFrame = !m.externalVR->WaitFrameResult();
-  m.externalVR->GetFrameResult(surfaceHandle, textureWidth, textureHeight, leftEye, rightEye);
+  m.externalVR->GetFrameResult(surfaceType, surfaceHandle, textureWidth, textureHeight, leftEye, rightEye);
   ExternalVR::VRState state = m.externalVR->GetVRState();
   if (state == ExternalVR::VRState::Rendering) {
     if (!aDiscardFrame) {
       if (textureWidth > 0 && textureHeight > 0) {
         m.device->SetImmersiveSize((uint32_t) textureWidth/2, (uint32_t) textureHeight);
       }
-      m.blitter->StartFrame(surfaceHandle, leftEye, rightEye);
-      m.device->BindEye(device::Eye::Left);
-      m.blitter->Draw(device::Eye::Left);
+      m.device->EnableExternalSurfaceRender(surfaceType ==
+        mozilla::gfx::VRLayerTextureType::LayerTextureType_ExternalVRSurface);
+      // In Oculus platform, we can render WebGL immersive frames info AndroidSurface.
+      if (surfaceType != mozilla::gfx::VRLayerTextureType::LayerTextureType_ExternalVRSurface) {
+        m.blitter->StartFrame(surfaceHandle, leftEye, rightEye);
+        m.device->BindEye(device::Eye::Left);
+        m.blitter->Draw(device::Eye::Left);
 #if !defined(VRBROWSER_NO_VR_API)
-      m.device->BindEye(device::Eye::Right);
-      m.blitter->Draw(device::Eye::Right);
+        m.device->BindEye(device::Eye::Right);
+        m.blitter->Draw(device::Eye::Right);
 #endif // !defined(VRBROWSER_NO_VR_API)
+      }
     }
     m.device->EndFrame(aDiscardFrame);
     m.blitter->EndFrame();
--- a/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
+++ b/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
@@ -635,6 +635,8 @@ const vrb::Vector kAverageHeight(0.0f, 1.7f, 0.0f);
 const vrb::Vector kAverageOculusHeight(0.0f, 1.65f, 0.0f);
 
 struct DeviceDelegateOculusVR::State {
+  const static uint32_t EXTERNAL_SURFACE_BUFFER_SIZE = 1;
+
   struct ControllerState {
     const int32_t index;
     const ElbowModel::HandEnum hand;
@@ -663,6 +665,7 @@ struct DeviceDelegateOculusVR::State {
   ovrJava java = {};
   ovrMobile* ovr = nullptr;
   OculusEyeSwapChainPtr eyeSwapChains[VRAPI_EYE_COUNT];
+  ovrTextureSwapChain* eyeSurfaceSwapChain[EXTERNAL_SURFACE_BUFFER_SIZE];
   OculusLayerCubePtr cubeLayer;
   OculusLayerEquirectPtr equirectLayer;
   std::vector<OculusLayerPtr> uiLayers;
@@ -725,6 +728,9 @@ struct DeviceDelegateOculusVR::State {
       VRB_LOG("Failed to initialize VrApi!. Error: %d", status);
       exit(status);
       return;
+    }
+    for (int i = 0; i < EXTERNAL_SURFACE_BUFFER_SIZE; ++i) {
+      eyeSurfaceSwapChain[i] = nullptr;
     }
     initialized = true;
     SetRenderSize(device::RenderMode::StandAlone);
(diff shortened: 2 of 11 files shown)


app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -610,6 +610,8 @@ struct DeviceDelegateOculusVR::State {
   ovrJava java = {};
   ovrMobile* ovr = nullptr;
   OculusEyeSwapChainPtr eyeSwapChains[VRAPI_EYE_COUNT];
+  const static uint32_t EYE_SURFACE_COUNT = 2;
```
MortimerGoro: nit: you can use exiting `VRAPI_EYE_COUNT` constant

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -1098,6 +1100,17 @@ DeviceDelegateOculusVR::SetRenderMode(const device::RenderMode aMode) {
   vrb::RenderContextPtr render = m.context.lock();
   for (int i = 0; i < VRAPI_EYE_COUNT; ++i) {
     m.eyeSwapChains[i]->Init(render, m.renderMode, m.renderWidth, m.renderHeight);
+    // auto surfaceOut = vrapi_GetTextureSwapChainAndroidSurface(m.eyeSwapChains[i]->ovrSwapChain);
+    // surfaceOut is null from vrapi_GetTextureSwapChainAndroidSurface(m.eyeSwapChains[i]->ovrSwapChain);.
+  }
+
+  for (int i = 0; i < DeviceDelegateOculusVR::State::EYE_SURFACE_COUNT; ++i) {
+    if (!m.eyeSurfaceSwapChain[i]) {
+      m.eyeSurfaceSwapChain[i] = vrapi_CreateAndroidSurfaceSwapChain(m.renderWidth, m.renderHeight);
```
MortimerGoro: SwapChains are being recreated twice, first for standalone mode then for immersive. If they need to be created before immersive render mode is set, you can use GetImmersiveRenderSize() instead of m.renderWidth (they may have different sizes for immersive and standalone)

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -1098,6 +1100,17 @@ DeviceDelegateOculusVR::SetRenderMode(const device::RenderMode aMode) {
...
+      auto surfaceOut = vrapi_GetTextureSwapChainAndroidSurface(m.eyeSurfaceSwapChain[i]);
+      surfaceOut = m.java.Env->NewGlobalRef(surfaceOut);
```
MortimerGoro: swapChains and surface needs to be deallocated each time the app goes to background an resumes, check `DeviceDelegateOculusVR::EnterVR` and `DeviceDelegateOculusVR::LeaveVR`

app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -1174,23 +1174,32 @@ BrowserWorld::DrawImmersive() {
...
   if (state == ExternalVR::VRState::Rendering) {
     if (!aDiscardFrame) {
       if (textureWidth > 0 && textureHeight > 0) {
         m.device->SetImmersiveSize((uint32_t) textureWidth/2, (uint32_t) textureHeight);
       }
-      m.blitter->StartFrame(surfaceHandle, leftEye, rightEye);
-      m.device->BindEye(device::Eye::Left);
-      m.blitter->Draw(device::Eye::Left);
+      // In Oculus platform, we can render WebGL immersive frames info AndroidSurface.
+      if ((m.device->GetDeviceType() != device::OculusGo) &&
```
MortimerGoro: Can we use `if (surfaceType == mozilla::gfx::VRLayerTextureType::LayerTextureType_ExternalVRSurface)`instead of checking by device?

app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -1199,6 +1208,12 @@ BrowserWorld::DrawImmersive() {
       m.blitter->CancelFrame(surfaceHandle);
     }
     DrawLoadingAnimation();
+
+    if ((m.device->GetDeviceType() != device::OculusGo) &&
```
MortimerGoro: Can we use `if (surfaceType == mozilla::gfx::VRLayerTextureType::LayerTextureType_ExternalVRSurface)`instead of checking by device?

MortimerGoro: Or just always call  m.device->SetExternalSurfId(-1) here

MortimerGoro: Also this call seems to be in the else branch which shows the loading animation. Should it be in the other branch or at the end of the if conditions? next to the end `m.device->EndFrame` I mean

MortimerGoro (changes requested):

app/src/main/cpp/BrowserWorld.cpp
daoshengmu: It looks like `DrawLoadingAnimation ()` is at the else branch of `if (state == ExternalVR::VRState::Rendering)`. I didn't change the logic of this part, and it seems make sense to let draw loading animation when it is not at the rendering mode.

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
daoshengmu: I would like to keep this because it means we are doing doubled-buffer surface instead two buffers each for one eye.

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
daoshengmu: It looks like I just need to deallocate the `m.eyeSurfaceSwapChain` at `LeaveVR()`, and It will be recreated at ` SetRenderMode()`.

app/src/main/cpp/BrowserWorld.cpp
daoshengmu: Sounds great!

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -1443,13 +1461,30 @@ DeviceDelegateOculusVR::EndFrame(const bool aDiscard) {
...
+    // Flip texCoord in vertical when using WebGL frame textures.
+    const ovrMatrix4f texFlip = ovrMatrix4f_CreateScale(1, -1, 1);
+    ovrMatrix4f proj = ovrMatrix4f_Multiply(&projectionMatrix, &texFlip);
+    proj = ovrMatrix4f_TanAngleMatrixFromProjection(&proj);
+
+    for (int i = 0; i < DeviceDelegateOculusVR::State::EYE_SURFACE_COUNT; ++i) {
+      const auto eyeSwapChain = m.eyeSurfaceSwapChain[swapChainIndex];
+      // Set up OVR layer textures
+      projection.Textures[i].ColorSwapChain = eyeSwapChain;
+      projection.Textures[i].SwapChainIndex = swapChainIndex;
```
MortimerGoro: `projection.Textures[i].SwapChainIndex` should be `0`. It's used when Oculus creates the double/triple buffer internally, but we are creating the buffer externally (`vrapi_CreateAndroidSurfaceSwapChain` doesn't support bufferCount parameter)

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -1098,6 +1101,21 @@ DeviceDelegateOculusVR::SetRenderMode(const device::RenderMode aMode) {
   vrb::RenderContextPtr render = m.context.lock();
   for (int i = 0; i < VRAPI_EYE_COUNT; ++i) {
     m.eyeSwapChains[i]->Init(render, m.renderMode, m.renderWidth, m.renderHeight);
+    // auto surfaceOut = vrapi_GetTextureSwapChainAndroidSurface(m.eyeSwapChains[i]->ovrSwapChain);
```
MortimerGoro: nit: remove comments

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -582,6 +582,8 @@ const vrb::Vector kAverageHeight(0.0f, 1.7f, 0.0f);
 const vrb::Vector kAverageOculusHeight(0.0f, 1.65f, 0.0f);
 
 struct DeviceDelegateOculusVR::State {
+  const static uint32_t EYE_SURFACE_COUNT = 2;
```
MortimerGoro: nit: I'd use a better name to identify with external, eg. `EXTERNAL_SURFACE_BUFFER_SIZE`

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -1443,13 +1461,30 @@ DeviceDelegateOculusVR::EndFrame(const bool aDiscard) {
...
+      projection.Textures[i].TexCoordsFromTanAngles = ovrMatrix4f_TanAngleMatrixFromProjection(&projectionMatrix);
+    }
+  } else {
+    const int swapChainIndex = externalSurfaceId;
+    // Flip texCoord in vertical when using WebGL frame textures.
+    const ovrMatrix4f texFlip = ovrMatrix4f_CreateScale(1, -1, 1);
+    ovrMatrix4f proj = ovrMatrix4f_Multiply(&projectionMatrix, &texFlip);
+    proj = ovrMatrix4f_TanAngleMatrixFromProjection(&proj);
+
+    for (int i = 0; i < DeviceDelegateOculusVR::State::EYE_SURFACE_COUNT; ++i) {
```
MortimerGoro: `VRAPI_EYE_COUNT` should be used for the loop instead of the `EYE_SURFACE_COUNT`. We need to set the values for both eyes, if the EYE_SURFACE_COUNT was set to > 2 it could overflow.

MortimerGoro (dismissed):
Thanks for the updates! This is getting closer to be ready! 

Other than the mentioned issues WebVR samples demo looks zoomed compared to the master build. It seems something related to bad sizes or bad eye offsets. Could you check that?

daoshengmu:
> Thanks for the updates! This is getting closer to be ready!
> 
> Other than the mentioned issues WebVR samples demo looks zoomed compared to the master build. It seems something related to bad sizes or bad eye offsets. Could you check that?

ExternalVRSurface
externalVR

EGLTexture
egltexture

@MortimerGoro 
Regarding to the immersive texture size, I have checked they both are 1024 x 1024, and their difference is not so obvious as my attachments.

Besides, I notice that if I release my surfaces at `LeaveVR()`, the surface will not be presented properly after re-entering the immersive mode. However, if I don't release my surfaces, it will still be good after exiting FxR and re-enter.

kennardconsulting:
Looking forward to this being merged! Any progress on resolving conflicts?

daoshengmu:
> Looking forward to this being merged! Any progress on resolving conflicts?

I have solved conflicts in local and am working on other new issues after merging with master branch.

kennardconsulting:
Awesome. Thanks for the update. Can you comment on whether it helps this issue in the sample app you show?

https://github.com/MozillaReality/FirefoxReality/issues/1408

Specifically, whether the moving cube (and the stationary cubes if moving your head) appear to judder every few seconds

daoshengmu:
I am going to give another commit to provide an UI switch in the developer setting to turn on/off external surface feature.

kennardconsulting:
Awesome. Thanks again for all your hard work on this. I really appreciate it and look forward to seeing the end result.

If you get time, could you please comment on how this feature affects the 'judder' on the moving cubes (and the stationary cubes if moving your head) in the WebVR samples demo on Oculus Quest?

daoshengmu:
It seems like I am using a wrong projection matrix when adopting Swapchain surfaces to do immersive render at [Sketchfab.com](https://sketchfab.com/3d-models/the-owl-knight-724ff67781f244fe856bea42df1816d4)

daoshengmu:
Per #304

kennardconsulting:
I'm unclear if this issue is now closed?

daoshengmu:
> I'm unclear if this issue is now closed?

Not yet. We are blocked by Bugzilla review. Sorry about that.

daoshengmu:
Now, doing profiling experiments from [webgfx-tests](https://github.com/MozillaReality/FirefoxReality/wiki/Profiling), and run it via `webgfx-tests run webvr_cubes -a "SN" -b fxr -c ./examples/tests/webgfx-tests.config.json -o result.json -n 8`

disable_vs_enable

We can see we can get a little higher FPS if we adopt this approach (enable). The first time to draw or page loading time would be more than the original one (disable) is because we need to ask OVR to resize our external surface buffer from the content side, the only way I can get is asking the content thread sleep for `80` ms to make sure we can get the resized surface.

disable_vs_2surf

However, if we choose to use two external surface buffers. The FPS will be low. It seems like the framebuffer memory pressure (we have a couple of screen size framebuffers) causes the FPS goes down. I have checked disabling all extra JNI bindings and `WebGLContext::BlitBackbufferToCurDriverFB` for this experiment, and we still get the same result. So, I pretty sure this would be our bottleneck.

PR #1438, opened 2019-07-20
Reduce false positive performance monitor detection

bluemarvin:


bluemarvin:
Depends on https://github.com/MozillaReality/vrb/pull/66

bluemarvin:
Now depends on https://github.com/MozillaReality/vrb/pull/67

MortimerGoro:
I tried ShaderToy and it took a long time to detect the jank. It's that on purpose to avoid false positives?

bluemarvin:
How did you test it? If you launch with the broken url it will take some time to detect. But I don’t consider that the typical use case. But I could have broken it. The way it should work is after the frame rate has stabilized junk should take about two seconds to detect.

MortimerGoro:
> How did you test it? If you launch with the broken url it will take some time to detect. But I don’t consider that the typical use case. But I could have broken it. The way it should work is after the frame rate has stabilized junk should take about two seconds to detect.

I launched with the default URL, typed shadertoy in the awesome bar and open it from the Google search

bluemarvin:
Okay, sounds broken 😓. I’ll take a look when back to my computer.

keianhzo:
I've made some tests on the Go and false positives detection has improved for regular pages but shadertoy takes too long time to be detected and sometimes it crashes the app.

keianhzo:
@bluemarvin See to work much better than before, less false positives and although it still takes a while to jump in for performance intensive sites like shadertoy, it doesn't feel as long as before.

MortimerGoro:
I get some crashes when trying Shadertoy:

```
2019-07-24 10:28:09.825 976-1334/? W/ActivityManager: Force removing ActivityRecord{b8a0723 u0 org.mozilla.vrbrowser.dev/org.mozilla.vrbrowser.VRBrowserActivity t13930}: app died, no saved state
2019-07-24 10:35:50.867 976-1087/? I/ActivityManager: Process org.mozilla.vrbrowser.dev (pid 3474) has died
```

Probably related to memory limit because I don't see a stacktrace.

bluemarvin:
I landed this because the false positives on the Focus were getting really annoying. We can tune further with follow ups or just turn it off if we don't feel it is helping.

Issue #1329, opened 2019-06-13
Telemetry for multi-window 


jvonitter:
- How many too-many-window dialogs happen / what percentage
- Long press discovery
- +tray button use
- How long is a window open for / window life
- Frequency of window moves
- Frequency of window resizes (and most common size if we can)
- Curved windows setting - how many users use it / what percentage
- When a session is multi-window, what percentage of time is each position the active window
- Average of how many windows are open at a time, per session

Needed by #1319

thenadj:
@jvonitter Can we add three more?

+ What position is most often used for active window (left, middle, right)
+ Resizing: What window sizes are used? How often?
+ How many windows are open at a time?

jvonitter:
@thenadj Yeah, we can add those. I'll update the ticket above

keianhzo:
@jvonitter do we have a final list of events that we want to add to multiwindow?

jvonitter:
updated the description to be a bit clearer

jvonitter:
I've reached out to the data steward for ET to review this data collection. Stay tuned!

keianhzo:
Closed by #1467

jvonitter:
Data review form, attn Janice Tsai - 
[datareview-multiwindow.txt](https://github.com/MozillaReality/FirefoxReality/files/3518546/datareview-multiwindow.txt)

Wiki update in progress

jvonitter:
https://github.com/MozillaReality/FirefoxReality/wiki/Telemetry has been updated with the info here. closing this!

Issue #1353, opened 2019-07-01
Telemetry fails to upload


bluemarvin:
## Hardware

All
## Steps to Reproduce

1. Visit Mozilla telemetry

## Current Behavior

Notice drop off in stats.
## Expected Behavior

Telemetry should continue to be uploaded.

## Possible Solution

Update the android-component telemetry service to use an alternative method for uploading telemetry.

## Context

PR https://github.com/MozillaReality/FirefoxReality/pull/1085 broke telemetry when we removed the android permission `RECEIVE_BOOT_COMPLETED`. The service was using this permission to upload telemetry. With this permission removed, we need to find an alternative solution that is permissible in the Oculus store for uploading telemetry.

daoshengmu:
@pocmo Per our discussion at https://github.com/MozillaReality/FirefoxReality/pull/1300#issuecomment-505661982, we are asking for assistance to avoid to use `RECEIVE_BOOT_COMPLETED` permission, and our workaround fix at issue #1300 still only be able to upload about 25% data to the telemetry system.

jvonitter:
Can we turn the permission back on for non-Oculus builds?

MortimerGoro:
@jvonitter yes we can do that. The permission removal was only done for Quest, so maybe we can even enable it for Oculus go (we have already separate builds)

daoshengmu:
> Can we turn the permission back on for non-Oculus builds?

That's a good thought. We should roll back it for other platforms except Quest. We also can see if our MAU number in telemetry is really getting back.

MortimerGoro:
@jvonitter @daoshengmu I'll create a PR with the permission enabled again except for Quest

pocmo:
> @pocmo Per our discussion at #1300 (comment), we are asking for assistance to avoid to use RECEIVE_BOOT_COMPLETED permission

So we did see the telemetry drop for *all* builds / devices? From the docs I would expect it to still run the job and retry it. What's weird is that we take care of persisting pings so the next schedule should send all ping that have not uploaded yet. So in the worst case this should make pings come in slower but not lose them.

Some things we could try:

* Calling `Telemetry.scheduleUpload()` early, e.g. on app start. It will schedule a job to pick up all the pings that are still on disk and try to upload them.

[...]

PR #1443, opened 2019-07-22
Added Multi-window Telemetry

keianhzo:
Added telemetry events from #1329

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
@@ -297,6 +297,10 @@ public class VRBrowserActivity extends PlatformActivity implements WidgetManager
     @Override
     protected void onStop() {
         super.onStop();
+
+        if (SettingsStore.getInstance(this).getCylinderDensity() > 0.0f)
+            TelemetryWrapper.queueCurvedModeActiveEvent();
+
         TelemetryWrapper.stop();
     }
 
--- a/app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
@@ -27,16 +27,20 @@ import org.mozilla.vrbrowser.utils.DeviceType;
 import org.mozilla.vrbrowser.utils.UrlUtils;
 
 import java.net.URI;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Map;
 
+import androidx.annotation.NonNull;
 import androidx.annotation.UiThread;
 
 import static java.lang.Math.toIntExact;
+import static org.mozilla.vrbrowser.ui.widgets.Windows.*;
 
 
 public class TelemetryWrapper {
     private final static String APP_NAME = "FirefoxReality";
-    private final static String LOGTAG = "VRB";
+    private final static String LOGTAG = TelemetryWrapper.class.getSimpleName();
     private final static int MIN_LOAD_TIME = 40;
     private final static int LOADING_BUCKET_SIZE_MS = 100;
     private final static int MIN_IMMERSIVE_TIME = 1000;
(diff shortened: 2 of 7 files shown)


MortimerGoro:
```
/opt/FirefoxReality/app/src/main/res/layout/context_menu_item.xml:15: AAPT: error: resource drawable/context_menu_item_background_color (aka org.mozilla.vrbrowser:drawable/context_menu_item_background_color) not found.
error: failed linking file resources.
```

app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
```diff
@@ -344,6 +348,8 @@ protected void onResume() {
 
     @Override
     protected void onDestroy() {
+        TelemetryWrapper.destroy();
+
```
daoshengmu: I am curious why you want to queue your multiple window telemetry at `onDestory()` instead of be like our previous telemetry, queue them at `onStart()` and `onStop()`?

keianhzo: Some windows related events like window lifetime histogram or window placement pct are better measured at the end of the session as you can stop and resume but the windows are not recreated so the stat wouldn't be completely accurate if we flush the data in the middle. We could artificially trigger a window destroy/re-create at onStop but this just made things simpler.

keianhzo:
@daoshengmu I've added some logs so all the multi-window related telemetry can be seen at the end of the session. Let me know if you catch anything.

app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
daoshengmu: put in onDestroy, I would be afraid the metric will not be uploaded for Oculus release. We usually do it `onStart/onStop` for `recordSessionStart/recordSessionEnd` and then scheduleUpload().

IIUC, `onDestroy ` will be called after `onStop`?

daoshengmu:
We probably will start to replace our existing telemetry service to `Glean` in Q3, because the telemetry service will not be maintained soon. We need to keep an eye on it for the coming changes.

app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
```diff
@@ -292,5 +344,217 @@ public static void uploadImmersiveToHistogram() {
...
+        Log.d(LOGTAG, "\tLEFT: " + pcts[WindowPlacement.LEFT.getValue()]);
+        Log.d(LOGTAG, "\tRIGHT: " + pcts[WindowPlacement.RIGHT.getValue()]);
+
+        for (int index = 0; index< MAX_WINDOWS; index++) {
+            activePlacementTime[index] = 0;
+            activePlacementStartTime[index] = 0;
+        }
+    }
+
+    private static void queueOpenWindowsAvgEvent() {
```
MortimerGoro: Should we measure time here too (or in a extra data) as you are doing in `queueActiveWindowPctEvent`? You can have 1.5 windows open in a session but how do you know how long they have used 1 windows or 2 windows?

keianhzo: I think the goal of that stat is knowing what's the average of open windows in a session but we can also make a a % of how much time the user spends with 1, 2 or 3 open windows. It just depends of what's more useful for us.

jvonitter: % of how much time the user spends with 1, 2 or 3 open windows would be great! probably more useful than the average number.

keianhzo: @jvonitter ok, I'll add both :)

keianhzo:
@daoshengmu Just FYI I've removed the the Telemetry queueing from `start` as it’s always empty at that point (we always flush in on `stop`) so it doesn’t make sense to queue it. I've just removed the data queuing NOT the upload scheduling, we still need that to solve the boot permission issues.

app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
keianhzo: Fixed this

daoshengmu:
> @daoshengmu Just FYI I've removed the the Telemetry queueing from `start` as it’s always empty at that point (we always flush in on `stop`) so it doesn’t make sense to queue it. I've just removed the data queuing NOT the upload scheduling, we still need that to solve the boot permission issues.

I just took a look at `queueHistogram()`, it will write file stream into a storage. Every bucket is 0 value from `start`, so it will not affect the final result. But, they are still redundant data.

According to https://github.com/MozillaReality/FirefoxReality/issues/1353#issuecomment-508170240, our `BOOT` permission will only cause the regression of the job scheduler. So, I agree we can remove the `queueHistogram()` at `start`. Thanks.

PR #1466, opened 2019-07-26
Allow keyboard to be repositioned

### LINKED ISSUE #1146: Allow keyboard to be repositioned
Allow keyboard to be repositioned.

(UX and design aspects are covered in related user testing master issue #480. This issue is to track implementation work.)

- [ ] Implement dynamic positioning (i.e relative to web content) based on design specification.
- [ ] Implement a UI element which allows user to reposition temporarily.
- [ ] Allow the user to pin the keyboard in a preferred location, and to unpin and revert to dynamic positioning.

needed-by #480

MortimerGoro:
Fixes #1146


app/src/main/cpp/WidgetPlacement.h
```diff
@@ -37,10 +37,12 @@ struct WidgetPlacement {
   int32_t GetTextureWidth() const;
   int32_t GetTextureHeight() const;
 
+  static const float WORLD_DPI_RATIO;
```
bluemarvin: nit: I would prefer we reserve all caps for `#define` macro and use camel case for variables even `const` ones.

app/src/main/cpp/WidgetPlacement.h
```diff
@@ -37,10 +37,12 @@ struct WidgetPlacement {
...
   static WidgetPlacementPtr FromJava(JNIEnv* aEnv, jobject& aObject);
+  static WidgetPlacementPtr Create(const WidgetPlacement& aPlacement);
 private:
   WidgetPlacement() = default;
-  VRB_NO_DEFAULTS(WidgetPlacement)
+  WidgetPlacement(const WidgetPlacement&) = default;
```
bluemarvin: I think you need `WidgetPlacement& operator=(const WidgetPlacement&)` also?

app/src/main/cpp/WidgetMover.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+  State()
+      : widget(nullptr)
+      , attachedController(-1)
+      , moveBehaviour(WidgetMoveBehaviour::GENERAL)
+      , endRotation(0)
+  {}
+
+
+  WidgetPlacementPtr HandleKeyboardMove(const vrb::Vector& aDelta) {
```
bluemarvin: nit: you can probably return `WidgetPlacementPtr&` to reduce incrementing/decrementing the `std::shared_ptr<>`.

app/src/common/shared/org/mozilla/vrbrowser/browser/SettingsStore.java
```diff
@@ -521,5 +525,40 @@ public void setNotificationsEnabled(boolean isEnabled) {
...
+    }
+
+    public void saveKeyboardMove(WidgetPlacement aPlacement) {
+        try {
+            SharedPreferences.Editor editor = mPrefs.edit();
+            JSONObject json = new JSONObject();
+            json.put("x", aPlacement.translationX);
+            json.put("y", aPlacement.translationY);
+            json.put("z", aPlacement.translationZ);
+            json.put("r", aPlacement.rotation);
```
bluemarvin: Should the rotation be saved since it is calculated from the `x, y, z` values? That way if the layout changes and the calculation changes the rotation will still be calculated correctly when restored?

jvonitter:
> Additionally, It doesn't seem there is a way to reset the keyboard to the default position?
This is by design. Once users move it, they can move it wherever they want including the default position. It should reset once they leave the app.

bluemarvin:
> > Additionally, It doesn't seem there is a way to reset the keyboard to the default position?
> 
> This is by design. Once users move it, they can move it wherever they want including the default position. It should reset once they leave the app.

The keyboard position is saved between runs and there is no indication when the user has it back in the default position.

Additionally, I notice that you can move the keyboard farther to the right than to the left. This seems like an odd constraint.

jvonitter:
> The keyboard position is saved between runs and there is no indication when the user has it back in the default position.

There's nothing really special about the specific default position. If a user wants it below the awesomebar, they can move it there. If they want it somewhere else, that's ok too. I think whether or not it should reset between runs is something we should consider after we use for a while. If it persists between sessions now, let's see how that feels and switch it if it's bad.

> Additionally, I notice that you can move the keyboard farther to the right than to the left. This seems like an odd constraint.

That does seem weird and not intentional.

MortimerGoro:
PR updated with:
- Do not save keyboard movement after app is restarted

MortimerGoro:
I'll improve curved mode movement in a follow-up because I need the math improvements from the `multiwindow` branch not landed in master yet

app/src/common/shared/org/mozilla/vrbrowser/browser/SettingsStore.java
```diff
@@ -5,12 +5,16 @@
 import android.graphics.Color;
 import android.os.StrictMode;
 import android.preference.PreferenceManager;
+import android.util.Log;
 
+import org.json.JSONException;
+import org.json.JSONObject;
```
bluemarvin: Looks like `org.json.*` can be removed?

MortimerGoro: Fixed

PR #1476, opened 2019-07-30
Support for WaveVR appId 

### LINKED ISSUE #1458: Support staging specific appId
Some platform like HTC Vive come with preinstalled versions of FxR. At the moment there is no way of testing the staging builds on those devices as they are signed with a different key than the store build and so when doing:

`adb install -r FirefoxReality-1.3-32051617-wavevr-arm-staging-signed-aligned.apk`

we get:

```Failure [INSTALL_FAILED_UPDATE_INCOMPATIBLE: Package org.mozilla.vrbrowser signatures do not match the previously installed version; ignoring!]```

We should provide a different `appId` for staging builds so app install/updates can be performed on all the platforms regardless of them having a preinstalled versions of FxR.

Another option could be to sign the staging builds using the stores key but the previous alternative is preferred over this one.

**Additional context**

### PR DESCRIPTION
**Pull request by keianhzo**
Fixes #1458 Added a .wavevr for non-store WaveVR builds so we can install debug and staging builds without clashing with the preinstalled app id.

Also replace noapi x86 builds with x86_64.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -18,7 +18,7 @@ def getCrashRestartDisabled = { ->
     return "false"
 }
 
-def getDevApplicationIdSufix = { ->
+def getDevApplicationIdSuffix = { ->
     if (gradle.hasProperty("simultaneousDevProduction")) {
         return gradle.simultaneousDevProduction == "true" ? ".dev" : ""
     }
@@ -78,7 +78,7 @@ android {
             signingConfig getUseDebugSigningOnRelease() ? debug.signingConfig : release.signingConfig
         }
         debug {
-            applicationIdSuffix getDevApplicationIdSufix()
+            applicationIdSuffix getDevApplicationIdSuffix()
             pseudoLocalesEnabled true
         }
     }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by keianhzo** [comment]
@MortimerGoro updated!

File: versions.gradle [hunk 1]
```diff
@@ -47,6 +47,7 @@ def gecko_view = [:]
 gecko_view.nightly_armv7a = "org.mozilla.geckoview:geckoview-nightly-armeabi-v7a:$versions.gecko_view"
 gecko_view.nightly_arm64 = "org.mozilla.geckoview:geckoview-nightly-arm64-v8a:$versions.gecko_view"
 gecko_view.nightly_x86 = "org.mozilla.geckoview:geckoview-nightly-x86:$versions.gecko_view"
```
philip-lamb: I don't think we want to retain the x86 dependency.

app/build.gradle
```diff
@@ -455,6 +494,7 @@ if (findProject(':geckoview-local')) {
         armImplementation deps.gecko_view.nightly_armv7a
         arm64Implementation deps.gecko_view.nightly_arm64
         x86Implementation deps.gecko_view.nightly_x86
```
philip-lamb: I don't think we need to retain the x86 target here.

PR #1487, opened 2019-08-01
Ensure the vrb::RunnableQueue has been created before the Activiy tries to call it in onCreate()

bluemarvin:
This fixes the missing UI on start up for Oculus and probably SVR.


MortimerGoro:
Good catch!

PR #1488, opened 2019-08-02
Fixes correct version code generation and APK naming

keianhzo:
Fixes correct version code generation for arm64 builds and correct version code in the APK name


bluemarvin:
Per discussion it was decided we don't need to version based on CPU type and the code in question should just be removed.

philip-lamb:
@keianhzo Do you intend to open a new request with the other changes? We still need the WaveVR non-store ID.

PR #1500, opened 2019-08-05
Capture pointer input during move gestures. Improve curved keyboard move.

MortimerGoro:
- Fixes the pointer jump to a different widget during move gestures
- Improved curved mode keyboard movement


bluemarvin:
I see a strange jump some times when moving the keyboard left and right.

keianhzo:
I also see the micro jumps while moving the keyboard. Also when moving the keyboard the pointer seems to move faster than the keyboard itself so it doesn't stay at the same point during the movement. Other than that looks great for the rest of the widgets.

MortimerGoro:
I'll address the strange jump in a follow-up

PR #1506, opened 2019-08-05
Fixes non working reset for privacy settings

keianhzo:
The privacy settings were not being reset correctly.


bluemarvin:
Environment panel also say `Reset Display Settings`

keianhzo:
Fixed reset buttons descriptions and moved out the override env button from the env scrollview.

bluemarvin:
I'm going to land this. If @MortimerGoro has any feedback it can be done in a follow up.

PR #1507, opened 2019-08-05
Dialog for requesting permission for storing voice samples

### LINKED ISSUE #641: Speech/Voice Input Permissions
We need to add a dialog to ask users for permission to collect speech/voice data. 

From Andre:
"Mozilla would like (with user permission) to collect and process anonymous user voice samples to improve our voice recognition models. In order to do so we need to implement UI to explain to users what we will do with their voice, and allow them to opt in (and opt-out later if they change their mind).

Possible UI needs include:

    A button in settings that turns on/off sending voice data to Mozilla (default off)
    a way to prompt user the first time they try the voice icon to agree to sending voice data to Mozilla to improve our services - and explain that they can continue to use voice search even if they do not wish to share data with us"

keianhzo:
Fixes #641 A dialog has been added to ask the user for permission to store voice samples and transcriptions.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
@@ -1274,6 +1274,11 @@ public class VRBrowserActivity extends PlatformActivity implements WidgetManager
             newWindow.getSessionStack().newSessionWithUrl(uri);
     }
 
+    @Override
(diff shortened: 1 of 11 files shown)


keianhzo:
@bluemarvin Fixed

bluemarvin:
I'm going to land this. If @MortimerGoro has any feedback it can be done in a follow up.

PR #1509, opened 2019-08-06
Address playstore feedback

bluemarvin:


bluemarvin:
Will create a 1.3.1 once this lands.

PR #1513, opened 2019-08-06
Fix JPG skyboxes. Do not require restart to switch to a custom env.

### LINKED ISSUE #1493: Custom skyboxes not loading on Oculus Quest
## Hardware
Oculus Quest, Firefox Reality 1.3

## Steps to Reproduce
Follow the exact steps in https://github.com/MozillaReality/FirefoxReality/wiki/Environments

## Current Behavior
Skybox / Cubemap does not load (skybox is black)

## Expected Behavior
Expected one of the demo skybox's to load

```
```

</details>

MortimerGoro:
Fixes #1493

Also makes it possible to switch between stock and custom environments without a restart, even if they have different formats.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/settings/EnvironmentOptionsView.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/settings/EnvironmentOptionsView.java
@@ -66,26 +66,28 @@ class EnvironmentOptionsView extends SettingsView {
         mBinding.envOverrideSwitch.setOnCheckedChangeListener(mEnvOverrideListener);
 
         SettingsStore.getInstance(getContext()).setEnvironmentOverrideEnabled(value);
+        mWidgetManager.updateEnvironment();
     }
 
     private OnClickListener mResetListener = (view) -> {
-        boolean restart = false;
+        boolean updated = false;
         if (mBinding.envOverrideSwitch.isChecked() != SettingsStore.ENV_OVERRIDE_DEFAULT) {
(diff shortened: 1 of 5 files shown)


keianhzo:
LGTM, great feature!

MortimerGoro:
@jvonitter I have some ideas to improve the custom environments override. We can support many folders and automatically populate the environment list with the custom ones. That way you can switch between different stock and custom environments without having to enable the `override` env option (which is a bit confusing)

bluemarvin:
> @jvonitter I have some ideas to improve the custom environments override. We can support many folders and automatically populate the environment list with the custom ones. That way you can switch between different stock and custom environments without having to enable/disable the `override` env option (which is a bit confusing)

Well, the feature was only intended for testing new skyboxes. We should add telemetry to the feature to get an idea of how much it is being used (I'm also curious what are the more popular not-default built in skyboxes). There is a lot we can do to improve this if it is a popular feature.

jvonitter:
made a new issue for the telemetry - #1524 
we need to get through the data collection review before we can do it, but i'll update when that's done.

elliotttate:
@bluemarvin SideQuest now has Firefox Skybox sharing. In addition to telemetry, you can see stats on the number of downloads on each Skybox here: https://sidequestvr.com/#/app/184

Not sure if this an indication of people wanting it, but this post about the Firefox skyboxes was the top post yesterday in the Oculus Quest sub: https://www.reddit.com/r/OculusQuest/comments/cm235m/watch_netflix_youtube_etc_and_surf_the_internet/

PR #1517, opened 2019-08-06
Adds support for clear cache

### LINKED ISSUE #1454: Add developer option to clear cache
Requested by @ScottWitte - 
"As a developer, I need to be certain that the changes I make in code and upload are indeed being used when viewing in FxR. If a change doesn't work, is it because I didn't really fix it or the old code is stuck in cache?

BTW, I really hate Safari wiping out everything when I just want a clean cache. Please don't do THAT.

keianhzo:
Fixes #1454 Adds support for clearing the cache. At the moment just clears all the caches until with have a more complete spec to do it more granular. Also I've refactored sessions code and fixed:
- Not all sessions were being recreated for methods that required session recreation
- Removed the intermediate session stack as it's not really necessary we can handle it with the linked hashmap

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStack.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStack.java
@@ -24,6 +24,7 @@ import org.mozilla.geckoview.GeckoRuntime;
 import org.mozilla.geckoview.GeckoSession;
 import org.mozilla.geckoview.GeckoSessionSettings;
 import org.mozilla.geckoview.MediaElement;
+import org.mozilla.geckoview.StorageController;
 import org.mozilla.geckoview.WebRequestError;
 import org.mozilla.vrbrowser.R;
 import org.mozilla.vrbrowser.browser.Media;
@@ -833,6 +834,27 @@ public class SessionStack implements ContentBlocking.Delegate, GeckoSession.Navi
         }
 
         recreateAllSessions();
+    }
+
+    public void clearCache(final long clearFlags) {
+        if (mRuntime != null) {
+            // Per GeckoView Docs:
+            // Note: Any open session may re-accumulate previously cleared data.
+            // To ensure that no persistent data is left behind, you need to close all sessions prior to clearing data.
+            // https://mozilla.github.io/geckoview/javadoc/mozilla-central/org/mozilla/geckoview/StorageController.html#clearData-long-
+            for (Map.Entry<Integer, SessionState> entry : mSessions.entrySet()) {
+                SessionState state = entry.getValue();
+                if (state != null) {
+                    state.mSession.stop();
+                    state.mSession.close();
+                }
+            }
+
+            mRuntime.getStorageController().clearData(clearFlags).then(aVoid -> {
+                recreateAllSessions();
+                return null;
+            });
+        }
     }
 
     // NavigationDelegate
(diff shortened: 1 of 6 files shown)


app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStack.java
```diff
@@ -803,8 +815,29 @@ protected void setTrackingProtection(final boolean aEnabled) {
...
+            // https://mozilla.github.io/geckoview/javadoc/mozilla-central/org/mozilla/geckoview/StorageController.html#clearData-long-
+            for (Map.Entry<Integer, SessionState> entry : mSessions.entrySet()) {
+                SessionState state = entry.getValue();
+                if (state != null) {
+                    state.mSession.stop();
+                    state.mSession.close();
+                }
             }
+
+            mRuntime.getStorageController().clearData(StorageController.ClearFlags.ALL).then(aVoid -> {
```
bluemarvin: This does more than just clear the cache. I think `StorageController.ALL_CACHES` would be better? If we want to purge all data I think it should be called something other than clear cache?

keianhzo: Moved to Privacy and updated to ALL_CACHES.

keianhzo:
@MortimerGoro @bluemarvin I've updated with the change request and also made some refactoring to remove the intermediate session stack that is not really necessary and that simplifies the sessionstack code a bit.

bluemarvin:
@keianhzo we should probably wait to land this until the UI spec has been updated? @jvonitter

keianhzo:
@bluemarvin There was never UI spec for this one, was just adding a clear cache button. You mean to have a more complete spec make the "clear" more granular?

bluemarvin:
> @bluemarvin There was never UI spec for this one, was just adding a clear cache button. You mean to have a more complete spec make the "clear" more granular?

Right, I was asking if we should wait for the spec with more details or land this and follow up once there is a spec.

jvonitter:
yes please wait for the spec. it's on Nadja's list and it'll be done soon

app/src/main/res/values/strings.xml
```diff
@@ -290,6 +290,17 @@
     <!-- The string labels an On/Off switch in the developer options dialog and is used to toggle enabling Servo. -->
     <string name="developer_options_servo">Enable Servo</string>
 
+    <!-- The string labels the description text for the clear cookies and site data button in the
+         privacy options dialog. -->
+    <string name="developer_options_clear_cache_cookies_site_description">Cookies &amp; SiteData</string>
```
bluemarvin: Should `SiteData` be `Site Data`?

keianhzo: Fixed

keianhzo:
We have to leave out the description from the spec as GV doesn't currently expose any API for getting the cache used disk space.

ScottWitte:
I'm the original requester for this feature and I'm a little concerned. Am I right that you will have one button that clears cache, history, and cookies, much as with Mobile Safari? If so, in my experience that would be a mistake. A major reason for clearing the cache is so you can reload a page and be certain you have the most current content. That is especially important when developing and is just about the first thing any support or FAQ tells a user to do when troubleshooting. When something doesn't work as expected you can be certain it isn't just because some old code is still being used. 

If you nuke everything just to clear the cache you force users to retype URLs, which is tedious and error-prone, especially in a VR interface, or to search for lost URLs, which can be difficult. I suggest you use the approach of all other browsers (except mobile Safari) and give users control over just what they want to clear. Look at mobile FireFox as a model.

Thanks for your consideration. Really excited to see how FxR is developing!

bluemarvin:
There are two buttons, one clears cache, one clears cookies and data.

ScottWitte:
Excellent. Thanks for the clarification.

Issue #1511, opened 2019-08-06
Browser window is too low in Oculus Quest


MortimerGoro:
@thenadj and @jvonitter mentioned that  the position of the window on the Quest feels different (and really low) than on the Go.

It's probably related to the initial height set up in Oculus Quest. I'll take a look.

MortimerGoro:
@jvonitter @thenadj I tested this and I think it works as expected.

The initial height of the user is always the same as Oculus Go: `1.75m`. From that position your movements can affect the scene height:
- If you open the app while seated and you stand up the browser window will look in a lower position because you are "taller" now
- If you open the app while standing and then you sit or lie in the floor the window will be in a higher position because you are looking it from above.

In both cases you can reset the pose to the standard height just using the recenter button in the controller or using the HMD off/on button (each time you remove and put the headset this is recentered too).

@jvonitter @thenadj can you confirm the issue come from any of those position changes or give STR if that's not the case.

The alternative would be to always use a fixed scene height based on the floor level. But this may come with some problems:
- Each user will see a slightly different view initially based on their height. 
- You can't make both sitting and standing comfortable. If we adjust the perfect view for one the other would be affected (browser high or low). 
- You can't reset the height as we are doing now

MortimerGoro:
I created this PR to simplify some math we are doing: https://github.com/MozillaReality/FirefoxReality/pull/1512. It shouldn't affect this unless there was some bug in the offsets we calculated.

MortimerGoro:
After some conversations with @jvonitter we confirmed that the initial view is the same in Oculus Go and Quest. So this problem is not related to the Quest positional tracking math.

Janice thinks we should place the window a bit higher:

![image](https://user-images.githubusercontent.com/1070794/62638344-8bbbe000-b93d-11e9-8e58-57b4318ae167.png)

On the image above the yellow box is where the center of my view is. The circle is where @jvonitter  thinks it should be.

MortimerGoro:
Fixed in #1548 

@jvonitter if there is any feedback to address please reopen it

jvonitter:
We reviewed it a bunch on different devices and decided higher is better, but that this is a little too high. Please lower it to be between where it was before and where it is now.

PR #1535, opened 2019-08-07
Place the window a bit higher

MortimerGoro:
Please check the conversation and the images posted in #1511.

This is an attempt to try a bit higher window height. We should get feedback before landing it.

One problem we have compared to Oculus is that they have a fixed window height, so they have always centered in the same position. For us placing the window higher may feel good in the 1x size but has a bigger impact for the 2X 3X sizes. So we need to find a good balance.

### DIFF
--- a/app/src/main/res/values/dimen.xml
+++ b/app/src/main/res/values/dimen.xml
@@ -5,7 +5,7 @@
     <!-- Browser window -->
     <dimen name="browser_width_pixels">800px</dimen>
     <item name="window_world_width" format="float" type="dimen">4.0</item>
-    <item name="window_world_y" format="float" type="dimen">0.2</item>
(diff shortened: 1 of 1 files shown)


bluemarvin:
One thing with the window higher is that when watching a 2D video in fullscreen, the back,projection,brightness menu is higher and it is really bright. I think it would be cool to have the option to hide the controllers and menus when in fullscreen like in immersive video mode. Otherwise I guess moving the window up a little doesn't bother me too much.

jvonitter:
On first glance, I love it! I'm passing the apk to Nadja and I want to use it for a while, but I think it's ok to merge and test. We should check it on different devices/platforms as I've only checked Quest so far.

bluemarvin:
Let's land so people can start testing. It's easy to change back if needed.

PR #1539, opened 2019-08-08
Strip GeckoView logs from release builds.

MortimerGoro:


bluemarvin:
I'll try and add an API to GV so we don't need to use the pref.

PR #1542, opened 2019-08-08
Use original URI when switching UA on a page that was redirected

### LINKED ISSUE #1473: The mobile version of a webpage is displayed even though the desktop user agent is selected
**[Affected Versions]:**
- Firefox Reality 1.4-32101950

**[Affected Platforms]:**
- Oculus Go - Android 7.1.2

**[Prerequisites]:**
- Go to "Settings"(gear icon), "Display" tile and for the "User-Agent Mode" setting select the "Desktop" option.

**[Steps to reproduce]:**
1. Visit the following webpage: "facebook.com".
2. Click the "Switch User Agent" button in the search bar. (The mobile version of the webpage is displayed)
3. Click the "Switch User Agent" button again. 
4. Observe the loaded webpage.

**[Expected results]:**
- The desktop version of the webpage is displayed.

**[Actual results]:**
- The mobile version of the webpage is displayed even though the desktop user agent is selected.

**[Notes]:**
- The issue is also reproducible with the twitter.com webpage.
- Attached a screen recording of the issue: [link](https://drive.google.com/file/d/1pGqr74iTvhJWjNyU6XVFQvBHo0deQrRI/view?usp=sharing)

bluemarvin:
Partially fixes #1473

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStack.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStack.java
@@ -36,6 +36,10 @@ import org.mozilla.vrbrowser.telemetry.TelemetryWrapper;
 import org.mozilla.vrbrowser.utils.InternalPages;
 
 import java.util.ArrayList;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.ArrayDeque;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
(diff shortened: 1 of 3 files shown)


bluemarvin:
To test:
1) Go to `twitch.tv` (note: gets redirected to `m.twitch.tv`)
2) switch to desktop mode

URL should be `www.twitch.tv` (`twitch.tv` redirects to `www` on desktop)

Note for what will not work:
1) Go to `twitch.tv` (note: gets redirected to `m.twitch.tv`)
2) click on one of the videos
3) Switch to desktop mode

The domain will remain `m.twich.tv` even in desktop mode.

This patch only works when switching UA on a domain that was redirected. Once you click a link, FxR no longer knows it has been redirected.

To address always dropping the `m.` prefix FxR would need to handle that on a per domain basis.

keianhzo:
@bluemarvin I've seen this when switch back and forth from VR to Desktop:
1. Go to twitch.tv (note: gets redirected to m.twitch.tv)
2. Click on one of the videos
3. Switch to desktop mode
4. Click on the username of the highlighted video
5. Switch back to VR mode

**Result**: You go back to twitch.tv
**Expected**: You should reload the current site

bluemarvin:
That’s what fennec does too and it’s wrong. I’ll see if I can figure out what isn’t working.

bluemarvin:
This turning out to be more difficult that I first thought. The problem I am having is that in the twich.tv case specifically, it is a single page site. So we don't actually get `onLoadRequest()` for most link clicks on the site. so it is impossible to know when to not use the original URI that triggered the redirection. This also explains Fennec's similar behavior. What's needed is for `onLocationChange()` which does fire to include a parameter that indicates if the location change was caused by human interaction of clicking on a link or if it is just the page playing with the history.

bluemarvin:
It was suggested I look at the history delegate to see if it is possible to determine what is causing the location change.

bluemarvin:
After trying a lot of different things with GeckoView, until GV can tell us if a visited URI was human initiated or script initiated there isn't a way to track redirects. This is a brute force method that just drops the `m.` prefix if the URI contains it before switching to desktop mode.

keianhzo:
@bluemarvin LGTM just the merge conflicts and maybe we could also add support for `mobile.`? Twitter uses that instead of `m.` and probably some others.

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStack.java
```diff
@@ -777,13 +779,51 @@ public int getUaMode() {
...
+                    state.mSettings.setViewportMode(GeckoSessionSettings.VIEWPORT_MODE_DESKTOP);
+                    overrideUri = checkForMobileSite(state.mUri);
+                } else {
+                    state.mSettings.setViewportMode(GeckoSessionSettings.VIEWPORT_MODE_MOBILE);
+                }
+                mCurrentSession.getSettings().setViewportMode(state.mSettings.getViewportMode());
+                if (overrideUri != null) {
+                    mCurrentSession.loadUri(overrideUri, GeckoSession.LOAD_FLAGS_BYPASS_CACHE);
+                } else if (state.mUri != null){
+                    mCurrentSession.loadUri(state.mUri, GeckoSession.LOAD_FLAGS_BYPASS_CACHE);;
```
keianhzo: nit: double `;`at the end of this line.

bluemarvin:
> After trying a lot of different things with GeckoView, until GV can tell us if a visited URI was human initiated or script initiated there isn't a way to track redirects. This is a brute force method that just drops the `m.` prefix if the URI contains it before switching to desktop mode.

It looks like twitter uses cookies. Even if I drop the `mobile.` prefix, it just comes back even with the desktop UA.

PR #1552, opened 2019-08-09
User agent polishing

### LINKED ISSUE #1527: user agent polish- tooltip & icon update
Right now the user agent toggle in the awesome bar says "switch user agent"
It should say something more user friendly/familiar. I suggest "Request desktop site" to align with mobile.

We should also remove the "VR" icon and use just the monitor icon with a checkmark when it's active.

needs #1473

keianhzo:
Fixes #1527 User agent polishing


keianhzo:
@bluemarvin For what I understood from #1527 we were going to mimic the mobile behavior where the text is always "Request desktop site" and the indicator if the state is the check in the icon. @jvonitter ?

bluemarvin:
I find that behavior confusing. On mobile the text makes sense since it is next to a check box while I expect a tool tip to tell me what will happen when I click a button.

PR #1566, opened 2019-08-13
Prevent crash reporter from getting stuck in an infinte loop

bluemarvin:


bluemarvin:
It looks like sometimes FxR is getting re-launched faster than the crash service can start and so the service will see the Pid of the newly created activity and gets stuck in an infinite loop waiting for the new activity to quit. This prevents it from waiting forever.

keianhzo:
@bluemarvin I've been trying to reproduce the issue but I couldn't. Could you provide STRs to repro this?

bluemarvin:
@keianhzo The STR is to trigger a crash on start up where FxR gets restarted before the crash service starts watching for PIDs. You will see this in the log while FxR is running:
```
08-13 14:09:42.054 11549 11566 E VRB     : ======> Found PID 11714
08-13 14:09:42.155 11549 11566 E VRB     : ======> Found PID 11714
08-13 14:09:42.257 11549 11566 E VRB     : ======> Found PID 11714
08-13 14:09:42.359 11549 11566 E VRB     : ======> Found PID 11714
08-13 14:09:42.460 11549 11566 E VRB     : ======> Found PID 11714
08-13 14:09:42.562 11549 11566 E VRB     : ======> Found PID 11714
08-13 14:09:42.666 11549 11566 E VRB     : ======> Found PID 11714
08-13 14:09:42.767 11549 11566 E VRB     : ======> Found PID 11714
08-13 14:09:42.870 11549 11566 E VRB     : ======> Found PID 11714
08-13 14:09:42.971 11549 11566 E VRB     : ======> Found PID 11714
```
I used this patch to capture an intermittent start up crash in GeckoView:

https://crash-stats.mozilla.com/report/index/0665bd92-f30e-4f69-9026-0eb100190813

With out the patch, FxR would crash, restart and the above message would be in the log and the Crash Reporter dialog would never be shown.

bluemarvin:
If there is a better solution, feel free to suggest one, but this patch unblocked me and prevents the crash reporter from running forever.

PR #1574, opened 2019-08-15
Add debug logging switch to release build developer options

bluemarvin:


bluemarvin:
Note, I've post a patch for GV to include the pref used in the 1.3.3 release. We will need to bump GV versions once this lands:

https://bugzilla.mozilla.org/show_bug.cgi?id=1574341

PR #1575, opened 2019-08-15
Support placing tray in front of the current active window

### LINKED ISSUE #1521: multiwindow polish - center tray with active window
The tray should be centered beneath the active window (instead of staying fixed in the center)

keianhzo:
Closes #1521 Always place the tray in front of the active window


MortimerGoro:
New implementation here: #1632

PR #1600, opened 2019-08-19
Added logs and handle CompletableFutures exception

### LINKED ISSUE #1570: Awesome bar and bookmarks don't work in noapi nmode
## Hardware
Emulator

## Steps to Reproduce

1. Click on the URL bar
2. Start typing

## Current Behavior

The awesome bar is not shown.

## Expected Behavior

The awesome bar should appear and start displaying suggestions.

## Possible Solution

It seems the are getting `java.lang.RuntimeException: Places functionality not available` from the `PlacesBookmarksStorage` service but only on the emulator. I've checked with the AC team. 

Also we should handle exceptions when using CompletableFutures to at least log the cause. In the awesome bar we should handle exceptions when getting suggestions as now we are just halting and not getting any other type of suggestions just because the bookmarks call is throwing an exception.

<details open>

```
java.lang.RuntimeException: Places functionality not available
```

</details>

keianhzo:
Fixes #1570 Better exception handling for Places related content in suggestions. There seems to be some problem with Places in x86_64 https://mozilla.slack.com/archives/CA4JKKLSG/p1565878585191200 already informed the AC team. In the meantime this helps to show the available content in the awesome bar instead of showing nothing.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/search/suggestions/SuggestionsProvider.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/search/suggestions/SuggestionsProvider.java
@@ -1,6 +1,7 @@
 package org.mozilla.vrbrowser.search.suggestions;
 
 import android.content.Context;
+import android.util.Log;
 
 import androidx.annotation.NonNull;
 
@@ -16,6 +17,8 @@ import java.util.List;
 import java.util.concurrent.CompletableFuture;
 
 public class SuggestionsProvider {
+
+    private static final String LOGTAG = SuggestionsProvider.class.getSimpleName();
 
     public class DefaultSuggestionsComparator implements Comparator {
 
@@ -90,6 +93,11 @@ public class SuggestionsProvider {
                 items.sort(mComparator);
             }
             future.complete(items);
+
+        }).exceptionally(th -> {
+            Log.e(LOGTAG, "Error getting bookmarks suggestions: " + th.getLocalizedMessage());
+            future.complete(items);
+            return null;
         });
 
         return future;
@@ -111,6 +119,11 @@ public class SuggestionsProvider {
                 items.sort(mComparator);
             }
             future.complete(items);
+
+        }).exceptionally(th -> {
+            Log.e(LOGTAG, "Error getting history suggestions: " + th.getLocalizedMessage());
+            future.complete(items);
+           return null;
         });
 
         return future;
@@ -152,6 +165,11 @@ public class SuggestionsProvider {
                 items.sort(mComparator);
             }
             future.complete(items);
+
+        }).exceptionally(th -> {
+            Log.e(LOGTAG, "Error getting search engine suggestions: " + th.getLocalizedMessage());
+            future.complete(items);
+            return null;
         });
 
         return future;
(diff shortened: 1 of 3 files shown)


app/src/common/shared/org/mozilla/vrbrowser/search/suggestions/SuggestionsProvider.java
```diff
@@ -17,6 +18,8 @@
 
 public class SuggestionsProvider {
 
+    private static final String LOGTAG = SuggestionsProvider.class.getSimpleName();
```
bluemarvin: Is this how we want to handle LOGTAGs going forward? It makes it hard to grep for FxR generated logs.

app/src/common/shared/org/mozilla/vrbrowser/search/suggestions/SuggestionsProvider.java
```diff
@@ -90,6 +93,11 @@ public void setComparator(Comparator comparator) {
                 items.sort(mComparator);
             }
             future.complete(items);
+
+        }).exceptionally(th -> {
+            Log.e(LOGTAG, "Error getting bookmarks suggestions: " + th.getLocalizedMessage());
```
bluemarvin: I assume the exceptions do not contain PII? I'm paranoid after all the work it took to clean up FxR.

app/src/common/shared/org/mozilla/vrbrowser/search/suggestions/SuggestionsProvider.java
keianhzo: Android Studio already gives you the option of showing only the current app logs so I think filtering by class it's more useful when you want to debug a specific class (or group of classes if you use `|`). I'd rather start moving logs to this format.

bluemarvin: So the problem I have with using AndroidStudio is that there are a number of log messages generated in the same process that are not FxR log message. For example on the Focus, there is that annoying foviation log message that spams the logcat. Being able to filter on VRB was how I was able to focus on just message from our code and not underlying (and often noisy) system log message.

app/src/common/shared/org/mozilla/vrbrowser/search/suggestions/SuggestionsProvider.java
keianhzo: Being part of AC I guess they don't, maybe @pocmo can answer that.

bluemarvin: I would hope not but since these API deal with bookmarks and history I could conceive of something get leaked out.

pocmo: Since this calls into Kotlin and Rust code of AS libraries we can't really guarantee that at the AC level.

bluemarvin: @keianhzo we should probably make the Logs debug then so they get stripped out for release?

keianhzo:
@bluemarvin updated

PR #1601, opened 2019-08-19
L10n fixes

keianhzo:
Fixes https://github.com/mozilla-l10n/android-l10n/pull/62 L10n fixes

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -733,7 +733,7 @@
 
     <!-- This string is for the tooltip that appears upon hovering the 'Request desktop site' button in the
          browser's navigation bar. The button it labels, when pressed, change the currently used user agent. -->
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -731,7 +731,7 @@
          speech-to-text recognition. -->
     <string name="voice_search_tooltip">Voice Search</string>
 
-    <!-- This string is for the tooltip that appears upon hovering the 'User Agent' button in the
+    <!-- This string is for the tooltip that appears upon hovering the 'Request desktop site' button in the
```
Delphine: The initial comment and string that landed was "Switch User Agent". Since this changes, there should be a new String ID to reflect that.

keianhzo: @Delphine Should be ok no

Delphine: Yes, thanks!

PR #1603, opened 2019-08-20
Implement multiwindow hover and click border effect.

### LINKED ISSUE #1518: multiwindow polish- selection/click state
Windows should have blue border on hover and click. Border should go away after click.

MortimerGoro:
Fixes #1518

### DIFF
--- a/app/src/main/cpp/WidgetBorder.cpp
+++ b/app/src/main/cpp/WidgetBorder.cpp
@@ -1,0 +1,239 @@
+/* -*- Mode: C++; tab-width: 20; indent-tabs-mode: nil; c-basic-offset: 2 -*-
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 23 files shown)


MortimerGoro:
Depth issues fixes, this is ready for testing

app/src/main/cpp/WidgetBorder.cpp
```diff
@@ -0,0 +1,239 @@
...
+    std::vector<int> index;
+    index.push_back(1);
+    index.push_back(2);
+    index.push_back(3);
+    index.push_back(4);
+    std::vector<int> normalIndex;
+    normalIndex.push_back(1);
+    normalIndex.push_back(1);
+    normalIndex.push_back(1);
+    normalIndex.push_back(1);
```
bluemarvin: We seem to do this, a lot, I wonder if we could refactor simple geometry creation out into a factory of some point?

bluemarvin: Actually, I see you sort of did that already here.

PR #1611, opened 2019-08-20
Update voice collection strings

jvonitter:
Updating legal language around voice collection permissions

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -860,11 +860,11 @@
 
     <!-- This string is displayed in the title of the Voice Search samples collect dialog. That dialog is shown the first time
          a user tries to use the voice search. -->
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -860,11 +860,11 @@
 
     <!-- This string is displayed in the title of the Voice Search samples collect dialog. That dialog is shown the first time
          a user tries to use the voice search. -->
-    <string name="voice_samples_collect_dialog_title">Allow Firefox to Anonymously Collect Voice Samples?</string>
+    <string name="voice_samples_collect_dialog_title">Allow Firefox Reality to Collect Voice Samples for Research?</string>
```
bluemarvin: I just realized we shouldn't have Firefox or Firefox Reality in the strings file. See: 

https://github.com/MozillaReality/FirefoxReality/blob/master/app/src/main/res/values/strings.xml#L243

I actually saw a few other places where the app name was hardcoded in the strings.xml file.

PR #1612, opened 2019-08-20
Replace all the hardcoded Firefox strings

keianhzo:
Replace all the hardcoded Firefox strings


keianhzo:
@bluemarvin you are right, updated.

PR #1613, opened 2019-08-20
Title bar support

### LINKED ISSUE #519: Render the current document's `<title>` in the browser UI/chrome
## Hardware
All

## Steps to Reproduce
1. Load in Firefox Reality a page with a `<title>`.
2. Notice that the `<title>`/`document.title` is not rendered anywhere in the UI/browser chrome.

## Current Behavior
The document title is not visible in the UI/browser chrome.

image

## Expected Behavior
The document title is visible somewhere in the UI/browser chrome.

## Possible Solution
This will likely need a design spec. I am not sure where it should appear, but the document title could appear above the browser window, for example. I don't have any good recommendations that come to mind at the moment.

keianhzo:
Fixes #519 Render the current document's `<title>` in the browser UI/chrome


keianhzo:
I have updated the title bar texts for home, private home and bookmarks. I've added a "Private Browsing" title when it the private browsing landing as it looked weird to be empty and hiding it doesn't feel consistent but let me know if that doesn't sound right and we can revert.

MortimerGoro:
LGTM

PR #1622, opened 2019-08-21
Fixes tooltips

keianhzo:
Fixes tooltips not being displayed.


keianhzo:
Not too much, there are alternatives but we are already using it across the app. Maybe we can file a follow up to refactor and land this?

bluemarvin:
> Not too much, there are alternatives but we are already using it across the app. Maybe we can file a follow up to refactor and land this?

Okay, I was just curious what functionally the gecko.ThreadUtils added that the previous solution did not.

bluemarvin:
https://github.com/MozillaReality/FirefoxReality/issues/1626

PR #1659, opened 2019-08-26
What until Youtube video has loaded data to request a video quality change

MortimerGoro:
Fixes 1433


keianhzo:
So much better! I hated that background spinner.

PR #1688, opened 2019-08-27
Add dontobfuscate option to pro-guard rules. Fixes #1683

### LINKED ISSUE #1683: Disable ofuscation in release builds
The stack traces from release builds are obfuscated and that makes them difficult to use. 

Disable obfuscation on release builds.

Uploading trace files for each build would require crash-stats support.

**Additional context**
Obfuscation does not seem to add much value for an open source based application.

bluemarvin:


keianhzo:
Yes, it seems minification itself mangles the line number and that cannot be changed so I guess we will need to upload the mappings to crash-stats to get useful stacktraces. Not sure though if the stores themselves support uploading mapping files. I know that Google Play does so you can get the original stack traces in the dashboard but I'm not sure if Oculus/HTC have that feature.

At least this commit let us know where the crash happened so it might be worth to land it while we wait for mappings support in crash-stats.

PR #1697, opened 2019-08-28
Add support for clearing window color before GV does the first composition

### LINKED ISSUE #1641: Browser window and nav bar should appear together when making a new window.
## Configuration

Firefox Reality version: all
Firefox Reality build ID: all

Hardware: all

## Steps to Reproduce

1. Create a new window.

## Current Behavior

The nav bar is shown but the window is not show until gecko renders to the surface texture.
## Expected Behavior

A window should be shown even if it is blank.
## Possible Solution

Draw a temporary quad placeholder for the window. It could be white for regular windows and purple for private browsing. Not drawing the window immediately on window creation makes FxR seem unresponsive and slow. We would need to have different implementations for layers and non-layer window rendering.

MortimerGoro:
Fixes #1641

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
@@ -675,19 +675,14 @@ public class VRBrowserActivity extends PlatformActivity implements WidgetManager
             if (aTexture == null) {
                 Log.d(LOGTAG, "Widget: " + aHandle + " (" + aWidth + "x" + aHeight + ") received a null surface texture.");
             } else {
-                aTexture.setOnFrameAvailableListener(new SurfaceTexture.OnFrameAvailableListener() {
-                    @Override
-                    public void onFrameAvailable(SurfaceTexture surfaceTexture) {
-                        surfaceTexture.setOnFrameAvailableListener(null);
-                        if (!widget.getFirstDraw()) {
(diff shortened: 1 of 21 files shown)


MortimerGoro:
We are considering postponing this to v1.5 until the new GV compositor callback API ready. It's hard to get the perfect delay without that callback.

philip-lamb:
OK, LMK what you decide. If not landed by 2019-08-29 17:00PDT I'll bump to 1.5.

bluemarvin:
I've created two GeckoView issues to help enable this feature:
https://bugzilla.mozilla.org/show_bug.cgi?id=1578947
https://bugzilla.mozilla.org/show_bug.cgi?id=1578949

bluemarvin:
I rebased this so that I could look into working on the GV fixes needed.

bluemarvin:
After rebasing this no longer works on WaveVR devices.

bluemarvin:
HTC is working, now just waiting for https://bugzilla.mozilla.org/show_bug.cgi?id=1578947 to land in GV nightly.

bluemarvin:
Looks like the custom frag shader does not work on NoAPI/Daydream and draws a white border around the window. Investigating...

bluemarvin:
> Looks like the custom frag shader does not work on NoAPI/Daydream and draws a white border around the window. Investigating...

After more testing, this appears to be an Android 9 bug as this patch works on the Mirage Daydream device which is Android 8.

PR #1699, opened 2019-08-28
Update FxR Java logging to include class name in LOGTAG

bluemarvin:


bluemarvin:
> For Widget classes would it be possible to create the LOGTAG on the base class so we don't need to add it on all widgets?

It probably is, but the intention was to know which component it was coming from. @keianhzo Do you have an opinion on the granularity of the logtag?

keianhzo:
Maybe we make the LOGTAG static field non final in UIWidget and initialize it as `this.getClass().getSimpleName()` and just use that one from all the derived classes?

bluemarvin:
> I guess what @MortimerGoro means is to make the LOGTAG static field non final in UIWidget and initialize it as `this.getClass().getSimpleName()` and just use that one from all the derived classes?

Wouldn't it also need to be not static either, otherwise the last class instantiated would overwrite the LOGTAG?

bluemarvin:
In fact shouldn't it be final but not static?

MortimerGoro:
Yes, I mean create a protected member using `SystemUtils.createLogtag(this.getClass())`. It should still use the real Widget class based tag on all widgets

bluemarvin:
> Yes, I mean create a protected member using `SystemUtils.createLogtag(this.getClass())`. It should still use the real Widget class based tag on all widgets

I think I understand the concept but not sure how to do this in Java. Where would it go? I the constructor? The Widget.java is an interface so it can't go in there.

bluemarvin:
Okay, I think I figured it out. I put it in UIWidget base class. My only reservation is now every instance of any Widget class has a string. Before it was just in the ones that need it and it was static.

MortimerGoro:
> Okay, I think I figured it out. I put it in UIWidget base class. My only reservation is now every instance of any Widget class has a string. Before it was just in the ones that need it and it was static.

We could create a `getLogTag()` method to lazy initialize it, but we don't have that many widgets for this string having an impact

PR #1778, opened 2019-09-05
Restore window size/position if the app is exited while in fullscreen

### LINKED ISSUE #1774: Windows not restored correctly if exiting app from inside VR Video
## Configuration

Firefox Reality version: 1.4
Firefox Reality build ID: 66a6788e

Hardware: All

## Steps to Reproduce

1. Open a seconds window to the left
2. In the window on the left go to youtube
3. Play a video and make it fullscreen
4. Click on the home button and quit the app
5. Open the app again

## Current Behavior

Both left and front windows are in the front position

screencap

## Expected Behavior

Windows should be restored in their original positions

## Possible Solution

Persist the original window position when in immersive/fullscreen modes where the window is moved to the front during immersive/fullscreen.

<details open>

```
```

</details>

keianhzo:
Fixes #1774 Fixes restoring the window position and size when quitting the app while in fullscreen mode. Before this PR the restored window position was not correct and the restored windows size was the fullscreen window size when quitting the app is windows > 1.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/NavigationBarWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/NavigationBarWidget.java
@@ -456,6 +456,7 @@ public class NavigationBarWidget extends UIWidget implements GeckoSession.Naviga
         }
 
         setFullScreenSize();
+        mAttachedWindow.setBackupPlacement(mPlacementBeforeFullscreen);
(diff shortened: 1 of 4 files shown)


keianhzo:
@MortimerGoro I've made the refactor to remove the backup placements from the NavigationBar to the WindowWidget, that way we can also handle all the cases when the session can be saved while a backup placement is used (fullscreen and resize ATM). 

Also fixed a crash when quitting while in resize mode.

PR #1782, opened 2019-09-06
Implement inverse touch function for UI automated testing in noapi

MortimerGoro:
This is a helper function for QA to start doing some automating UI testing in noapi


MortimerGoro:
Here is an example using the API: https://github.com/MozillaReality/FirefoxReality/tree/experimental/inverse_touch_test

It automatically clicks the private browsing button from the tray regardless of camera position

jrbenny35:
Is there a way this could be used with a test runner?

bluemarvin:
I'm hesitant to land code that doesn't get used as it tends to rot quickly. Does it make more sense to create a feature branch for testing and land this there?

MortimerGoro:
@bluemarvin I updated the PR to land it in a feature branch. Also removed this from https://github.com/MozillaReality/FirefoxReality/milestone/14

PR #1809, opened 2019-09-11
Update .taskcluster.yml to v1

djmitche:


djmitche:
OK, ready for review now, after a few fixes :)

As we [migrate to the new community deployment of taskcluster](https://bugzilla.mozilla.org/show_bug.cgi?id=1574661), we'd like to have all projects using v1, allowing us to eventually drop support for v0 :)

PR #1825, opened 2019-09-12
Include VisitType.RELOAD query type when re-visiting.

daoshengmu:
I don't why we want to remove the previous URL from the history even though the URLs are the same. IMHO, we should just keep it because they are **HISTORY**.


daoshengmu:
> Let me do some research about the current behavior of FF desktop. We need to query if it is existing, if so, removing the older URL and add it the top. If didn't find it, just adding it.

It works perfectly now, just like Desktop FF. It will update the URL from today's URL. If doing reload, it will update the visiting time.

bluemarvin:
This didn't work when I tried it before. I will look again when I have some time.

PR #1845, opened 2019-09-16
Fix crash from URI.create unhandled exceptions. Fixes #1844

### LINKED ISSUE #1844: Crash Report [@ java.net.URISyntaxException: at java.net.URI$Parser.fail(URI.java) ]
## Configuration

Firefox Reality version: 1.4 
Firefox Reality build ID: e6ec2b6c757f17a5ded6c3225adbbd8db8c6ee85

Hardware: Oculus (maybe others).

## Steps to Reproduce
This is the second highest crasher on https://crash-stats.mozilla.org/home/product/FirefoxReality for the FxR 1.4 release.

## Current Behavior

https://crash-stats.mozilla.org/report/index/d16a9ec8-f1cb-4f0c-90e9-c6cdb0190916
## Expected Behavior

FxR should not crash.

bluemarvin:


daoshengmu:
I am trying to catch the `URISyntaxException` from
[URI.java:2856](http://hg.openjdk.java.net/jdk9/dev/jdk/file/a37774add79a/src/share/classes/java/net/URI.java#l2851)
```
 private void failExpecting(String expected, String prior, int p)
            throws URISyntaxException
        {
            fail("Expected " + expected + " following " + prior, p);
        }
```

Because this exception is from `URI.create(urlString);`. But it only throws an `IllegalArgumentException`. So, I think this is all we can do.

bluemarvin:
As stated here: https://developer.android.com/reference/java/net/URI.html#create(java.lang.String) The `URISyntaxException` is wrapped in an `IllegalArgumentException` when thrown from `URI.create()`

PR #1847, opened 2019-09-16
Write a file containing the crash files location. Fixes #1846

### LINKED ISSUE #1846: Prevent FxR from loosing crash reports when it fails to restart after a crash
When a crash occurs, sometimes the FxR fails to restart and the crash report is lost. It would be nice to persist the crash report so that the next time FxR is run, the user has a chance to deal with the crash report.

Write out a file when the crash occurs. Then every time FxR starts it looks for the file and prompts the user if they want to submit the crash.

We have tried several different approaches to try and ensure that FxR restarts buy nothing has proven to be reliable. Have a file that persists is the best way to cover as many crashes as possible.

bluemarvin:


daoshengmu:
What is the main purpose you wanna send `extrasPath` in our crash reports? Is it a local file path?

bluemarvin:
> What is the main purpose you wanna send `extrasPath` in our crash reports? Is it a local file path?

The crash reporter writes out two files: dump file and extra file. They both have to be uploaded.

app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
```diff
@@ -481,44 +489,106 @@ private void handleConnectivityChange() {
...
+            mCrashDialog.setCrashDialogDelegate(() -> postCrashFiles(dumpFile, extraFile));
+            mCrashDialog.show(UIWidget.REQUEST_FOCUS);
+        }
+    }
+
+    private void sendCrashFiles(@NonNull final String aDumpFile, @NonNull final String aExtraFile) {
+        try {
+            GeckoResult<String> result = CrashReporter.sendCrashReport(VRBrowserActivity.this, new File(aDumpFile), new File(aExtraFile), getString(R.string.crash_app_name));
+
+            result.then(crashID -> {
```
snorp: You can use `result.accept` here which is a little nicer.

bluemarvin: Thanks, done.

PR #1858, opened 2019-09-19
Add vrbopen.sh script for opening URL in FxR

bluemarvin:


tools/utils/vrbopen.sh
```diff
@@ -0,0 +1,41 @@
...
+    exit 1
+  fi
+fi
+
+if [ "$VRB_EXT" == "default" ] ; then
+  VRB_EXT=""
+fi
+echo Using $VRB_APP$VRB_EXT
+
+if [ "$URL" = x ] ; then
```
daoshengmu: I'm thinking if it should be `"X$URL" = "X"` if you would like to check if $URL is null  or empty

bluemarvin: If it is null there is no need to check. I can add quotes around `x` for consistency. I might also be able to get rid of the `x` entirely.

keianhzo:
@bluemarvin can we add info about this to the README/Wiki so people are aware of it's existence? I guess most people will look into those places before start using or developing for FxR.

bluemarvin:
> @bluemarvin can we add info about this to the README/Wiki so people are aware of it's existence? I guess most people will look into those places before start using or developing for FxR.

Yep, was already working on adding it, just other things have been taking precedent. I'll just close it until I have more time to work on it.

PR #1867, opened 2019-09-20
Remove most 32bit builds from taskcluster builds. Fixes #1855

### LINKED ISSUE #1855: Dissable 32 bit builds
Now that we are shipping 64 bit builds to the stores. We should disable 32 bit builds.

**Additional context**
Now that we aren't using 32 bit builds we could reduce our taskcluster build times by almost half.

bluemarvin:


tools/taskcluster/build_targets.py
```diff
@@ -33,14 +33,14 @@
...
 platforms = {
-   'oculusvr': ['arm', 'arm64'],
-   'oculusvrStore': ['arm', 'arm64'],
-   'oculusvr3dofStore': ['arm', 'arm64'],
-   'wavevr': ['arm', 'arm64'],
+   'oculusvr': ['arm64'],
+   'oculusvrStore': ['arm64'],
+   'oculusvr3dofStore': ['arm64'],
+   'wavevr': ['arm64'],
    'wavevrStore': ['arm', 'arm64'],
```
daoshengmu: Do we still wanna support `arm` as the first option for wavevrStore? `return [archList[0]]` at Ln 66 will make `arm` to be the first choice.

bluemarvin: https://github.com/MozillaReality/FirefoxReality/pull/1867/files#diff-0d8ef6f62b99db4845f293a1ada2cd85R59
Makes `arm64` the default if nothing is specified..

PR #1873, opened 2019-09-24
Increase click target size for buttons inside Library views

### LINKED ISSUE #1852: Increase click target size for buttons inside Library views
In history or bookmarks, the click targets for the 'delete' and 'more' icons are too small. Extend them all the way to the top and bottom of the hover bar and further to the left and right. 

(cc @thenadj )

keianhzo:
Fixes #1852 Increased the click target area and reduced the library items row height to match the spec.


MortimerGoro:
@keianhzo this should be ready for merge when conflicts are fixed

keianhzo:
@MortimerGoro fixed

PR #1874, opened 2019-09-24
Clear GV history after wiping of the history

### LINKED ISSUE #1749: Clearing history does not update the forward button.
## Configuration

Firefox Reality version: 1.4
Firefox Reality build ID: 03c4d09c54bfc736d2ac69f7352bd13e815af626

Hardware: All

## Steps to Reproduce

1. click on a link
2. press back button
3. clear history

## Current Behavior

The forward button is still clickable.
## Expected Behavior

After clearing history, the forward button should not be clickable.

keianhzo:
Fixes #1749 Clear navigation buttons after history is cleared by recreating the session. Also fixed the way history is cleared, before we could clear yesterday and last week's entries but seeing the way desktop does it it makes more sense to always clean from now to those ranges.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStack.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStack.java
@@ -267,7 +267,7 @@ public class SessionStack implements ContentBlocking.Delegate, GeckoSession.Navi
         mVideoAvailabilityListeners.remove(aListener);
     }
 
-    public void restore(SessionStack store, int currentSessionId) {
+    public void restore(@NonNull SessionStack store, int currentSessionId) {
         mSessions.clear();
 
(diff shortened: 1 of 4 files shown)


bluemarvin:
I think we should wait until GV support the session history clear. Having the windows disappear when clearing history is jaring. I think having the forward/back buttons still work is the lesser of the two issues. Maybe break out the fix for what history gets cleared to a different PR?

bluemarvin:
https://bugzilla.mozilla.org/show_bug.cgi?id=1583265

bluemarvin:
`GeckoSession.purgeHistory()` will probably be in the September 27th build of Nightly GeckoView.

bluemarvin:
@keianhzo The GeckoSession.purgeHistory() function is now available in GeckoView.

bluemarvin:
This is now addressed in #1903

PR #1879, opened 2019-09-25
Library fixing/polishing

### LINKED ISSUE #1832: History Page animations do not seem to run when page does not have focus
## Configuration

Firefox Reality version: 1.4

Hardware: Quest (probably all)

## Steps to Reproduce

1. Open history
2. Open second window
3. open pages in new second window

## Current Behavior

The history page sporadically update and gets stuck mid animation until it is given focus.

## Expected Behavior

Animations should run smoothly in the history page even when it doesn't have focus.

### LINKED ISSUE #1842: The "TODAY" string from the "History" page glitches while scrolling
**[Affected Versions]:**
- Firefox Reality 1.5-32570012 (6090cf4)

**[Affected Platforms]:**
- Oculus Go - Android 7.1.2

**[Prerequisites]:**
- The history logs exceed the "History" page view. 

**[Steps to reproduce]:**
1. Click the "History" button from the tray menu.
2. Click>hold and scroll the page so that the "TODAY" string fades in the background on the top of the page. 
3. Observe the string.

**[Expected results]:**
- The string fades without glitches.

**[Actual results]:**
- At a certain point the string glitches while fading.

**[Notes]:**
- I could not verify the issue with the "YESTARDAY" and "LAST WEEK" strings because I had no available feeds for this sections.
- Attached is a screen recording of the issue: [link](https://drive.google.com/open?id=1qG1t7zneXsqfzV0YrXx1m2sX10bfGX31)

keianhzo:
Fixes #1842 Fixes #1832 Fixes the glitch from the first history header and issue with animations and list refresh while a library panel is opened. Also did some refactoring/polishing and fixed some colors that didn't match the spec.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/BookmarksStore.kt
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/BookmarksStore.kt
@@ -44,7 +44,6 @@ class BookmarksStore constructor(val context: Context) {
 
     fun addBookmark(aURL: String, aTitle: String) = GlobalScope.future {
         storage.addItem(BookmarkRoot.Mobile.id, aURL, aTitle, null)
-        notifyListeners()
         notifyAddedListeners()
     }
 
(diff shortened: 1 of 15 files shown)


bluemarvin:
If the navbar is clicked with only one window, the keyboard is shown as expected. Then if a second window is created and it's navbar is clicked, the keyboard is visible. It is only not visible if there is more than one window when the keyboard is shown for the first time.

PR #1883, opened 2019-09-25
Implement Copy Paste

### LINKED ISSUE #189: allow users to copy/paste text
When selecting a text and long pressing it an actions menu should show up with actions such as copy, share, select all, Google Search, etc

This applies to:
- Navigation Bar
- Browser window

MortimerGoro:
Fixes #189


MortimerGoro:
I've adressed the mentioned issues. Ready for review again.

bluemarvin:
@MortimerGoro I'm seeing one issue with the nav bar. STR:
1) Copy contents of navbar
2) Clear navbar
3) Type something new into navbar untill there isn't a recomendation
4) Long press to show action menu (Note it only says paste which is expected)
5) Now long click again on the text to select it.

Expected:
Action menu should add the option to copy

Actual:
Action menu doesn't change.

bluemarvin:
Another thing I seeing is some weirdness around the selection when typing in the navbar and their are suggestions you can make multiple selections:
selection

MortimerGoro:
We are merging current branch to a feature branch.

I created a new issue for the follow-ups: https://github.com/MozillaReality/FirefoxReality/issues/1983

PR #1887, opened 2019-09-26
Improve depth sorting: do not clamp hitPoint used for ndc calculation

MortimerGoro:
Also backouts workaround dimension changes included in https://github.com/MozillaReality/FirefoxReality/pull/1767


bluemarvin:
The taskcluster failure was from maven.

PR #1891, opened 2019-09-27
Unify MediaControls back button and controller back action

MortimerGoro:
Fixes inconsistency mentioned by @bluemarvin  

> "For 360 video, is there a reason the back B button on the controller drop you out of immersive and fullscreen at the same time while the back button on the media controls only drop you out of immersive mode? Seem inconsistent to me."

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -13,7 +13,7 @@ tasks:
       else: ${event.repository.clone_url}
     isFeatureBranch:
       $if: 'tasks_for == "github-push"'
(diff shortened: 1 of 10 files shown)


.taskcluster.yml
```diff
@@ -11,7 +11,10 @@ tasks:
       $if: 'tasks_for == "github-pull-request"'
```
bluemarvin: Looks like you were rebasing while I was abusing master on Friday. It should be safe to rebase now.

PR #1911, opened 2019-10-03
Change app version to use new system

bluemarvin:


bluemarvin:
org mozilla vrbrowser-20191003-094959

Issue #1125, opened 2019-04-24
`User-Agent` Override always returns default `User-Agent`


cvan:
## Steps to Reproduce

1. Load a site from the `userAgentOverride.json` list. Or add this entry to the file, and install it on FxR:
    
    ```
    "020506e9b65ec049e227e0a25dfafa84ebfc8eb366b6ce9731d757ae96ee223a14c0a51d6d8a72d45b71b6c779f2dd58841f29180827096a129145c5bcf608e6": "Mozilla/5.0 (Linux; Android 7.1.1; NGI77B) AppleWebKit/537.36 (KHTML, like Gecko) OculusBrowser/5.7.1.22.147827412 SamsungBrowser/4.0 Chrome/66.0.3359.203 Mobile VR Safari/537.36 example.com"
    ```
    
2. Using Remote Debugging, run `navigator.userAgent` in the Console:
    
    ```
    Mozilla/5.0 (Android 7.1.1; Mobile VR; rv:68.0) Gecko/68.0 Firefox/68.0
    ```

bluemarvin:
Looks like a GeckoView bug where if we launch FxR with an intent, `NavigationDelegate.onLoadRequest()` does not seem to get fired. If FxR is already running, it does.

philip-lamb:
@bluemarvin Is there a GV bug we can track?

philip-lamb:
Tested with build from current master on one of the sites on the list and can't reproduce, either with regular launch or launch by intent.

cvan:
I can consistently reproduce. Try a clean uninstall, build + run to Go, close and force stop FxR, launch a URL by adb using the VIEW activity. This trips me up a lot during testing and debugging, but for end users this bug will not surface. Working with Delight-VR over email to fix their UA sniffing. Barring any web-compat changes on any other sites, I should be able to clear the UA override list by EOW. We should still keep the UserAgentOverride API in FxR. Thanks, everyone!

bluemarvin:
It's definitely a GV issue, I'm root causing it right now.  It will impact Fenix far more than FxR right now. To work around, make sure to launch FxR before sending the URL intent.

bluemarvin:
This should be working now.

PR #1913, opened 2019-10-03
Restore userAgentOverride when restoring state (#1125)

agi90:
When restoring state using session.restoreState, `onLoadRequest` is not called
which causes FxR's UA override logic to not run. To fix this, we remember the
UA override string and restore it.


agi90:
Can someone merge this for me? :)

Issue #1530, opened 2019-08-07
Visual polish


jvonitter:
- [x] URL text box truncates too soon. [BUG-29a_URL_Bar_001.pdf](https://github.com/MozillaReality/FirefoxReality/files/3474677/BUG-29a_URL_Bar_001.pdf)

- [x] Voice search language setting dialog - height is too big and spacing is incorrect [BUG-29b_Voice_Search_Language_001.pdf](https://github.com/MozillaReality/FirefoxReality/files/3474678/BUG-29b_Voice_Search_Language_001.pdf)

- [x] Padding within tooltips is incorrect [BUG-30_Tooltip_001.pdf](https://github.com/MozillaReality/FirefoxReality/files/3474684/BUG-30_Tooltip_001.pdf)
- [x] Tray tooltips should be formatted differently from the rest of the tooltips [BUG-30_Tooltip_001.pdf](https://github.com/MozillaReality/FirefoxReality/files/3474684/BUG-30_Tooltip_001.pdf)

- [x] Editable text inside settings (such as Display->Density) should not be blue and chunky. It should be Open Sans Bold size 14, Color: Rhino (see https://trello.com/c/P52DGGmp/88-uis-16-settings) (MM)

- [x] Vertical separators are missing between the "Private Browsing" and "Bookmarks" buttons and between the "Downloads" and "Settings" buttons. https://trello.com/c/ENtbk8O4/352-uis-73-tray-with-library-buttons (MM)

- [x] Tray buttons are in the wrong order. New Window should be before Private Browsing.

- [x] At larger than 1x size, there should be padding on left and right side of bookmarks. Currently the content extends all the way to the left and right border of the window.  https://trello.com/c/jBtMLMKO/234-uis-62-bookmarks (MM)
- [x] Scrollbars should all move the same way. In the awesomebar suggestions in 1.3, the scrollbar is correct. In settings it moves in the opposite direction. When the bar is dragged, it should move together with the pointer. (MM)
- [x] Hover color for long press menu should be blue (instead of white) (see https://trello.com/c/AEOHvjHr/427-uis-85-copy-paste)

jvonitter:
@keianhzo if you want to do any of this between everything else... :-)

jvonitter:
do break these into individual issues if needed

PR #1915, opened 2019-10-04
UI Polishing

keianhzo:
This addresses most of the issues from #1530. Also added support for responsive library panels.

- The copy&paste issue will be addressed in the copy&paste PR
- The scrollbar standarization fix will the another PR


keianhzo:
@bluemarvin I have reverted the hint fix in favor of @MortimerGoro's one. Rg the try tooltips, I saw in the spec a note about leaving it as is. The margins, honestly I couldn't see any difference in margins either.

PR #1931, opened 2019-10-08
Making Focus Plus use WVR_InputId_Alias1_Trigger to check trigger only.

### LINKED ISSUE #1930: HTC Focus Plus make trigger button state be pressed when pressing touchpad in WebVR
## Configuration

Firefox Reality version: Dev. 5
Firefox Reality build ID: 

Hardware: Focus Plus

## Steps to Reproduce

1. Go to https://webvr.info/samples/XX-vr-controllers.html and enter VR
2. Press left on the touchpad, the button state of trigger will be true.

## Current Behavior

The button state of trigger will be true.

## Expected Behavior

The button state of trigger should be false.

<details open>

```
```

</details>

daoshengmu:
Fixes #1930. It looks like after the recent updates, Focus Plus only can use `WVR_InputId_Alias1_Trigger` to listen the trigger state. Otherwise, when pressing the left on the touchpad, it will send a `WVR_InputId_Alias1_Digital_Trigger`.

### DIFF
--- a/app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
+++ b/app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
@@ -303,8 +303,8 @@ struct DeviceDelegateWaveVR::State {
 
       delegate->SetVisible(controller.index, !WVR_IsInputFocusCapturedBySystem());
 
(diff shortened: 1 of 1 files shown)


daoshengmu:
@bluemarvin Please help confirm if it happens on your device too. I am not sure if it is because of my controller's glitch.

PR #1934, opened 2019-10-08
UI transitions animations

### LINKED ISSUE #66: Hover colors: ease in/out
Browser window: ease the colors in and out to make it look a bit smoother than just abrupt color change

keianhzo:
Fixes #66 This adds support for animation transitions in most of the UI elements. Although the are still some elements that need to be updated but will require some more work:

- Keyboard keys
- VRB related elements (outlines, widgets show/hide, ...)


keianhzo:
@jvonitter @thenadj In case you want to test this one. Maybe the fade in-out time is a bit high, we can lower that to make the fade effect faster but still appreciable.

Not applied to the keyboard yet as that's a quite custom implementation and would require some more work that this quick polish.

https://send.firefox.com/download/9df66d0c312ba842/#ef3n76vDZ8JfM4IoGVfmcw

jvonitter:
Thanks @keianhzo! we'll review it.

jvonitter:
This feels fine to me. We can keep an eye out for any slowness as we test FxR6 generally, but I'm ok with it as is.

PR #1936, opened 2019-10-08
Added support for a telemetry enabled/disabled ping

### LINKED ISSUE #1477: Add ping for telemetry enable/disable
We don't have a clear idea of what percentage of users disable telemetry. When a user elects to disable it, we should transmit a "goodbye" ping, and conversely a "hello" ping when a user enables the function.

This will help gauge user sentiment around collecting data to improve the browser.

keianhzo:
Closes #1477 This adds support for sending a ping when the telemetry is enabled/disabled.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
@@ -15,6 +15,7 @@ import org.mozilla.telemetry.measurement.SearchesMeasurement;
 import org.mozilla.telemetry.net.TelemetryClient;
 import org.mozilla.telemetry.ping.TelemetryCorePingBuilder;
 import org.mozilla.telemetry.ping.TelemetryMobileEventPingBuilder;
+import org.mozilla.telemetry.ping.TelemetryPingBuilder;
 import org.mozilla.telemetry.schedule.TelemetryScheduler;
 import org.mozilla.telemetry.schedule.jobscheduler.JobSchedulerTelemetryScheduler;
 import org.mozilla.telemetry.serialize.JSONPingSerializer;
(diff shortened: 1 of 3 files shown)


daoshengmu:
@keianhzo One thing we might miss is the users who have already set telemetry enable/disable in settings. [1] For this kind of users, we need to figure out a way to collect their data.

[1] https://github.com/MozillaReality/FirefoxReality/blob/85ab9b82e05664f28d0a3310508bd882ee2a80e0/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/settings/PrivacyOptionsView.java#L113.

daoshengmu:
> @keianhzo One thing we might miss is the users who have already set telemetry enable/disable in settings. [1] For this kind of users, we need to figure out a way to collect their data.
> 
> [1]
> 
> https://github.com/MozillaReality/FirefoxReality/blob/85ab9b82e05664f28d0a3310508bd882ee2a80e0/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/settings/PrivacyOptionsView.java#L113
> 
> .

@keianhzo did you update it for covering the users have been opt-out ? If so, please ask me review again.

keianhzo:
Good catch @daoshengmu I'll wait to see if this gets approved and implement this if that's the case.

keianhzo:
@daoshengmu I've added support for sending the telemetry status ping in case this is an update and the Telemetry was disabled. I think this should work. We are only sending if it was disabled as enabled is the default state.

app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
```diff
@@ -132,6 +134,7 @@
         private static final String THREE_OPEN_WINDOWS_TIME_PCT = "three_windows_open_time_pct";
         private static final String WINDOWS_OPEN_W_MEAN = "window_open_w_mean";
         private static final String WINDOWS_PRIVATE_OPEN_W_MEAN = "window_open_private_w_mean";
+        private static final String TELEMETRY_STATUS = "telemetry_status";
```
daoshengmu: Please try to avoid the string length be longer than 15. The key will be cut in our database.

keianhzo:
@daoshengmu I've shorten the status string and rebased.

keianhzo:
@daoshengmu @MortimerGoro Review comments have been addressed and data review approved. This should be ready to merge.

PR #1948, opened 2019-10-09
Update title bar position after neighbour window is resized

keianhzo:
When having >1 windows and you resize the front window the title from the side windows don't update their positions.

STRs:
- Open three windows
- Resize the front window horizontally

Actual Result:
The title bars of the side windows are not updated

Expected result:
The title bars should always be centered horizontally in the window


MortimerGoro:
LGTM

For the future we should improve the way we update recursive widgets. I created this issue: https://github.com/MozillaReality/FirefoxReality/issues/1953

PR #1987, opened 2019-10-16
Pop Ups blocking

### LINKED ISSUE #593: Add support for popup blocking UI
## Hardware

All

## Current Behavior

All popups are blocked.
## Expected Behavior

user should be prompted if a popup should be blocked.
## Possible Solution

GeckoView added a popup request the dialog delegate. FxR should add support for this so users can decide if a popup should be blocked. This should allow users to log in with facebook on 3rd party websites.

keianhzo:
Closes #593 Adds support for a pop-up blocking dialog. Also refactored the Prompt delegate from the Window widget.

You can test with: http://popuptest.com/

The dialog style is not exactly the spec one, I have tried to reuse the style of other similar dialogs (we now have a Base dialog widget) to avoid having multiple slightly different styles.

Also the pop-up allowed sites list is not the same as the spec one. I've used the same spec as the content languages list as they are both lists that live in the settings menu and makes sense that use the same style.


jvonitter:
@keianhzo the spec isnt current yet (it's in the queue) but that shouldn't hold this up. if the switches are there and they work, we can polish later if there are differences from the spec

bluemarvin:
The switch state in the `Advanced` menu seems to be swapped. I blocked the popups for both page and they both show as being allowed:
org mozilla vrbrowser-20191018-133321

PR #1994, opened 2019-10-16
Copy paste improvements

### LINKED ISSUE #1983: Copy Paste follow-ups
Reported by @bluemarvin: I'm seeing one issue with the nav bar. STR:
1) Copy contents of navbar
2) Clear navbar
3) Type something new into navbar untill there isn't a recomendation
4) Long press to show action menu (Note it only says paste which is expected)
5) Now long click again on the text to select it.

Expected:
Action menu should add the option to copy

Actual:
Action menu doesn't change.

nocopy

Another thing I seeing is some weirdness around the selection when typing in the navbar and their are suggestions you can make multiple selections:
selection

MortimerGoro:
Fixes #1983

- Correctly update Copy/Cut actions on the URLbar when the selection changes
- Hide or commit autocompletion selection when the user starts a select gesture
- Fix conflict between URL bar scroll gesture and select text gesture
- Do not recreate the CopyPaste menu if the new gesture has the same actions as before (avoids a minor flicker)

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/views/CustomInlineAutocompleteEditText.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/views/CustomInlineAutocompleteEditText.java
@@ -1,0 +1,40 @@
+package org.mozilla.vrbrowser.ui.views;
+
+import android.content.Context;
+import android.util.AttributeSet;
+
+import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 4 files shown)


keianhzo:
@MortimerGoro The only thing I see is that in the URL bar you don't get the menu when you double click on the text but in a web site you do. But I'd consider it a nit so I'm going to merge this and open a new issue for that.

PR #2027, opened 2019-10-21
Survey feedback link

### LINKED ISSUE #2011: add user feedback survey link
**Current**
In the settings main page, we currently show the logo, app name, version number and date

**Change to**
1. app name -> app name + version number  ("Firefox Reality 5")
2. link to survey - "Send us Your Feedback" in Azure #518fe1
    link is - https://qsurvey.mozilla.com/s3/FxR
3. date (no change)

Like this -
survey link

keianhzo:
Closes #2011


app/src/main/res/values/strings.xml
```diff
@@ -257,6 +252,10 @@
         opens the Firfox Reality support web site in the browser window. -->
     <string name="settings_help">Help</string>
 
+    <!-- This string is displayed in the settings header under the Firefox Reality logo. When clicked the settings panel is closed
+         a the survey web site in opened in the currently focused window -->
+    <string name="settings_send_your_feedback">&lt;a href="https://qsurvey.mozilla.com/s3/FxR">Send us Your Feedback&lt;/a></string>
+
```
bluemarvin: Is there a better way to handle the URL instead of including it in the l10n string? Seems like we should only have it in one place. Also, @jvonitter Do we have localized versions of this survey?

jvonitter:
@bluemarvin No localization for the survey at this point. We'll see what sort of responses we get and decide if we want to loc future surveys.

app/src/main/res/values/strings.xml
keianhzo: Updated with separate link/text.

PR #2056, opened 2019-10-24
Fix long press menu highlight state. Fixes #2047

### LINKED ISSUE #2047: fix long press menu highlight state
In the long-press menu for links-

- The top two options (open in new link and open in new tab) highlight together when you hover. 
- Other times you can hover over them and they dont highlight at all.
- Options should not have periods at the end of them

org mozilla vrbrowser-20191023-230023
org mozilla vrbrowser-20191023-230011

MortimerGoro:
Fixes #2047

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -903,10 +903,10 @@
     <string name="private_browsing_exit_tooltip">Exit Private Browsing</string>
 
     <!-- This string is shown in the context menu after a longpress on a link. When clicked it opens the target url in a new window. -->
-    <string name="context_menu_open_new_window">Open in a new window.</string>
+    <string name="context_menu_open_new_window_1">Open in a new window</string>
 
     <!-- This string is shown in the context menu after a longpress on a link. When clicked it opens the target url in a new tab. -->
-    <string name="context_menu_open_new_tab">Open in a new tab.</string>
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -903,10 +903,10 @@
     <string name="private_browsing_exit_tooltip">Exit Private Browsing</string>
 
     <!-- This string is shown in the context menu after a longpress on a link. When clicked it opens the target url in a new window. -->
-    <string name="context_menu_open_new_window">Open in a new window.</string>
+    <string name="context_menu_open_new_window_1">Open in a new window</string>
 
     <!-- This string is shown in the context menu after a longpress on a link. When clicked it opens the target url in a new tab. -->
-    <string name="context_menu_open_new_tab">Open in a new tab.</string>
+    <string name="context_menu_open_new_tab">Open in a new tab</string>
```
bluemarvin: Looks like you changed the version on `context_menu_open_new_window` but not `context_menu_open_new_tab`

MortimerGoro: context_menu_open_new_tab landed today, is not translated yet. Should we change that one too?

bluemarvin: I'm not sure how it works but I would change it just to be safe.

MortimerGoro: done

PR #2071, opened 2019-10-25
Hamburger menu

### LINKED ISSUE #1866: Send tab from FxR
Logged-in users can send a single tab from their VR headset to desktop/mobile devices.

Flow: https://trello.com/c/xkdu19Gw/422-uf-85-send-tab-from-fxr-to-device

### LINKED ISSUE #1898: add hamburger button next to the awesomebar for page options
Since we're adding a new page option for Send Tab, we need to make room for it in the page controls.

- [x] Replace resize button (to the right of the awesomebar) with hamburger button
- [x] Remove User Agent toggle from awesomebar
- [x] When clicked, hamburger button should trigger a context menu with three options:
1. `Send Tab` with send icon.  See UF-85 and #1866 
2. `Resize Window` with resize icon (no change to icon or behavior)
3. `Request Desktop Site` with UA icon either with or without checkmark, depending on settings.

### LINKED ISSUE #1939: URL hover highlight disapears when one of the internal buttons is hovered
## Configuration

Firefox Reality version: 5
Firefox Reality build ID: 0c9f13ee3cc23bf37d070e3ed721f067fca144fb

Hardware: All

## Steps to Reproduce

1. Hover over URL bar
2. Move to hover one of the URL Bar buttons such as the "voice search" or "create book mark button".

## Current Behavior

The URL bar should remain highlighted.
org mozilla vrbrowser-20191008-132800
org mozilla vrbrowser-20191008-132749
## Expected Behavior

The URL bar highlight disappears.

keianhzo:
Closes  #1866 Closes #1898 Closes #1939


jvonitter:
re @bluemarvin's comment above:
1) it should say "Menu"
4) There are a few options here. I passed the question up to Nadja. Check [here ](https://trello.com/c/rg6PsZgI/85-uis-21-window-controls)for the final answer - 
5) I've asked for an updated design. Check [here](https://trello.com/c/xkdu19Gw/422-uf-85-send-tab-from-fxr-to-device) for final answer

keianhzo:
@jvonitter I can't find the updated spec for the tabs dialog when the user is not connected in [UIS-21](https://trello.com/c/rg6PsZgI/85-uis-21-window-controls)

@bluemarvin I've fixed #1 #5 and #6. For #2 I think we don't intercept that URL and I don't see any logs of possible Gecko issues. Maybe we should force mobile UA when loading that URL?

bluemarvin:
> @bluemarvin I've fixed #1 #5 and #6. For #2 I think we don't intercept that URL and I don't see any logs of possible Gecko issues. Maybe we should force mobile UA when loading that URL?

My guess is it is a FxA error since it is just a web page, they must be UA sniffing or something. If it is possible to force the UA it might be a good idea. But it also wouldn't hurt to raise the issue with the FxA people who maintain the login page.

bluemarvin:
> 3. I find the `Switch to desktop` option still confusing.

I think it should just say `Desktop Mode`.

> 5. the checkmark when sending a tab is gigantic

I don't see any checkmark now. Is that what's intended?
 
> 6. When I edit the devices attached at accounts.firefox.com that list is not updated on the device. I had seven different Quests listed as targets so I went and disconnected them. On desktop, they disappeared instantly, but in FxR they are still show. The list should probably be update when the share dialog is opened.

So I logged in and went to the send tab list and it was empty. I open an closed it several times and it remained empty. If it is working to download the list then there should be some sort of visual indicator it is working/syncing. I went to take a screen shot and when FxR resumed the list was visible so either pause/resume triggered something or it was working the whole time.

keianhzo:
@bluemarvin updated to "Desktop Mode". The big check has been removed, now we show a notification in the tray. I think it's more consistent with what we are doing for things like bookmarks or received tabs.

The list should always be updated after logging in, we explicitly request a devices update when showing the dialog unless . I've added a "Syncing..." message while the devices are being updated although it should be pretty fast.

The FxA logging url is now opened in a new foreground tab.

keianhzo:
Updated

Issue #2081, opened 2019-10-28
Restore tray UI animation transitions


keianhzo:
## Configuration

Firefox Reality version: 6
Firefox Reality build ID: 

Hardware: All

## Steps to Reproduce

1. Click on the history tray button
2. Click on the Bookmarks tray button

## Current Behavior

The history button is still in active state

## Expected Behavior

History button should be in normal state

## Possible Solution

We have temporarily removed the animations for tray buttons as there seems to be an android bug while combining XML transition animations and setting the View background programmatically. Further investigation is required.

## Error Logs and Stack Traces

<details open>

```
```

</details>

keianhzo:
This seems to be an Android bug, it'll require and undefined amount of research time to get this fix.

PR #2082, opened 2019-10-28
Disable tray animations

### LINKED ISSUE #2054: UI transition animation causes some UI glitches
- Transparent background on tray tabs icon
- Invalid right icon background color in a focused URL bar
- On the hover animation in the URL bar you can notice that the URL bar and the right icon background have different colors

[ui_glitches.mp4.zip](https://github.com/MozillaReality/FirefoxReality/files/3768662/ui_glitches.mp4.zip)

keianhzo:
Closes #2054 It seems that dynamically settings the background doesn't play well with animations in certain cases. I was starting to be a time sink so I've just disabled the animation for the tray and opened a follow-up #2081

The other issues should be fixed by https://github.com/MozillaReality/FirefoxReality/pull/2071


MortimerGoro:
ok, it makes sense, when a tried to find a fix I thought it seemed like an Android bug

PR #2088, opened 2019-10-28
Attach the text input listener to the current active session

keianhzo:


app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/KeyboardWidget.java
```diff
@@ -287,6 +287,7 @@ protected void initializeWidgetPlacement(WidgetPlacement aPlacement) {
 
     @Override
     public void detachFromWindow() {
+        mAttachedWindow.removeWindowListener(this);
```
bluemarvin: I think we should null check `mAttachedWindow`

PR #2093, opened 2019-10-29
Capture tab snapshots when stacking sessions and opening the tabs dialog.

MortimerGoro:
Capture tab snapshots when stacking sessions and opening the tabs dialog. It uses a background display/surface when stacking sessions. For other situations we don't need to capture in background because the actions are originated from the tabs panel, which captures snapshots of active sessions now.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/utils/BitmapCache.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/utils/BitmapCache.java
@@ -203,6 +203,7 @@ public class BitmapCache {
         if (mCapturedAcquired) {
             return null;
         }
+        mCapturedAcquired = true;
         mCaptureSurfaceTexture.setDefaultBufferSize(width, height);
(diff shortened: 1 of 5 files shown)


app/src/common/shared/org/mozilla/vrbrowser/utils/BitmapCache.java
```diff
@@ -182,4 +193,42 @@ private void runIO(Runnable aRunnable) {
...
+
+    public void setCaptureSurface(SurfaceTexture aSurfaceTexture) {
+        mCaptureSurfaceTexture = aSurfaceTexture;
+        mCaptureSurface = new Surface(aSurfaceTexture);
+    }
+
+    public @Nullable Surface acquireCaptureSurface(int width, int height) {
+        if (mCapturedAcquired) {
+            return null;
+        }
```
bluemarvin: Shouldn't `mCaptureAcquired` be set to true here?

PR #2108, opened 2019-10-31
Fixes misalignments in the url bar items

keianhzo:
STRs:
- Load the content feed, see the left margin in the URL bar
- Open bookmarks/history, see the left margin in the URL bar

This PR fixes that misalignment along with the padding between the popup blocking icon and the secure/loading images.


bluemarvin:
I think I liked it more with the extra padding.

PR #2113, opened 2019-10-31
Make the UI react faster to page loads

keianhzo:
Improve the UI change reaction time when loading pages. ie. when loading a history/bookmarks item it when initiating a login flow. 

Instead of reacting on page load callback we just hide on page start, helps improve the responsiveness perception.


bluemarvin:
Why wait for page start? Why not just hide when clicking on a history item or bookmark?

keianhzo:
@bluemarvin I the case of History/Bookmarks we use that to cover the case where the user types a new URL int he URL bar, so we use it for all the navigation events instead of adding more callbacks.

PR #2114, opened 2019-10-31
Fix popup notification in curved mode

keianhzo:
This fixes the pop-up notification z-fighting in curved mode


bluemarvin:
I was going to report this and forgot. Thanks for fixing it anyway :)

Issue #1397, opened 2019-07-13
Megazord configuration for native Rust libraries


grigoryk:
To efficiently consume multiple appservices libraries, a megazord configuration is necessary.

Currently only `libplaces` is being used, for bookmarks storage.
#1389 adds a dependency on `libfxaclient` as well, and we'll need `liblogins` as well for #1394 and password management.

And so, we need to use a "megazord" which contains all the libraries we're using, compiled together.
In the near future consuming non-megazord appservices libraries will not be supported, so this needs to happen on a fairly quick timeline.

The actual work should be fairly straightforward: configuration gradle module substitutions, rust logging, and initializing megazord with a `concept-fetch` client implementation (e.g. gecko fetch client). Follow either what `reference-browser` or `fenix` are doing here.

grigoryk:
Changes to how megazords are consumed are coming in https://github.com/mozilla-mobile/android-components/pull/3806, so let's wait for that to land.

bluemarvin:
Is this required for v6 and FxA? Is it still on track or can we push it to v7?

keianhzo:
This has been already included in #1973 I followed the instructions above. @grigoryk can you confirm that everything is ok?

keianhzo:
Megazrod wasn't fully integrated. Addressed in https://github.com/MozillaReality/FirefoxReality/pull/2129

keianhzo:
Megazord wasn't fully integrated. Addressed in #2129

PR #2129, opened 2019-11-04
FxA updates

keianhzo:
Closes https://github.com/MozillaReality/FirefoxReality/issues/1397 
This addresses some comments from @grigoryk that didn't make it to the main PR. 

- getAuthenticationUrlAsync function refactoring: https://github.com/MozillaReality/FirefoxReality/pull/1973#discussion_r337533325
- Use a GeckWebExecutor for FxA requests
- Megazord integration: https://mozilla.github.io/application-services/docs/applications/consuming-megazord-libraries.html

The WorkManager auto-initialization fix is still there as crashes are still happening without this and we haven't found the cause.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStore.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStore.java
@@ -14,8 +14,6 @@ import org.mozilla.geckoview.ContentBlocking;
 import org.mozilla.geckoview.GeckoRuntime;
 import org.mozilla.geckoview.GeckoRuntimeSettings;
 import org.mozilla.geckoview.GeckoSession;
-import org.mozilla.geckoview.WebExtension;
-import org.mozilla.vrbrowser.BuildConfig;
 import org.mozilla.vrbrowser.VRBrowserApplication;
 import org.mozilla.vrbrowser.browser.BookmarksStore;
 import org.mozilla.vrbrowser.browser.HistoryStore;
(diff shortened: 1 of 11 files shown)


bluemarvin:
The tc failure was maven. I re-ran the task and it passed.

keianhzo:
@MortimerGoro @bluemarvin This was pushed to v7, I guess we can review/merge now.

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStore.java
```diff
@@ -64,48 +57,10 @@ private SessionStore() {
...
-        } else {
-            mRuntime.attachTo(context);
-        }
-    }
+        mRuntime = EngineProvider.INSTANCE.getOrCreateRuntime(context);
 
-    public GeckoRuntime getRuntime() {
-        return mRuntime;
+        // FIXME: Once GeckoView has a prefs API
+        SessionUtils.vrPrefsWorkAround(context, aExtras);
```
bluemarvin: I think the work around function should be called before the runtime is created.

keianhzo: Fixed

keianhzo:
I have updated this PR with Megazord support. After reading some more abut it at https://mozilla.github.io/application-services/docs/applications/consuming-megazord-libraries.html I've seen that Megazord was not properly integrated. The last commit adds proper support for it using the Fenix Megazord. Also bumps AC version to 22.

keianhzo:
@MortimerGoro @bluemarvin I haven't noticed any issues after Megazord + AC bump, give it a try and let me know. Using the custom Fenix Megazord should save us some APK size.

PR #2157, opened 2019-11-05
Audit promises exceptions

keianhzo:
Make sure all the promises exceptions are caught and logged and printed as stacktraces.


keianhzo:
I'm using `Log.d`to make sure they are stripped as they may contain personal info coming from places/accounts.

PR #2159, opened 2019-11-05
Fix the open windows weighted mean calculation

keianhzo:
This fixes an issue with the window count weighted mean that was accounting for the already opened windows after the telemetry was flushed.

Also after talking to @daoshengmu we decided that is better to force a telemetry status ping every time the app is created for easier telemetry status tracking in the backend. 

After this patch the telemetry status ping is always sent:
- When the app is opened
- When the user changes the telemetry status using the privacy options switch

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
@@ -509,7 +509,7 @@ public class TelemetryWrapper {
             Log.d(LOGTAG, "\tTWO: " + openWindowsTime[WindowPlacement.LEFT.getValue()]);
             Log.d(LOGTAG, "\tTHREE: " + openWindowsTime[WindowPlacement.RIGHT.getValue()]);
 
-            Log.d(LOGTAG, "Open Private Windows Count:");
+            Log.d(LOGTAG, "Open Windows Count:");
             Log.d(LOGTAG, "\tFRONT: " + openWindows[WindowPlacement.FRONT.getValue()]);
             Log.d(LOGTAG, "\tLEFT: " + openWindows[WindowPlacement.LEFT.getValue()]);
             Log.d(LOGTAG, "\tRIGHT: " + openWindows[WindowPlacement.RIGHT.getValue()]);
--- a/app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
@@ -601,7 +601,7 @@ public class TelemetryWrapper {
                 openPrivateWindows[i] = 0;
             }
             if (number > 0) {
-                openPrivateWindows[number -1 ] = 1;
+                openPrivateWindows[number-1] = 1;
             }
 
         } else {
@@ -610,7 +610,7 @@ public class TelemetryWrapper {
             }
 
             if (number > 0) {
-                openWindows[number - 1] = 1;
+                openWindows[number-1] = 1;
             }
         }
     }
--- a/app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
@@ -595,6 +595,26 @@ public class TelemetryWrapper {
         }
     }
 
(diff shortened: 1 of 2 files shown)


daoshengmu:
The window count weighted fix looks good to me.

daoshengmu:
It looks like the previous approach for tracking telemetry opt-out is the only way we could do for now. So, you can consider to remove your second commit and just land the first commit. @keianhzo

keianhzo:
Updated, although per our meeting discussion this will be hold until v7, right @bluemarvin ?

bluemarvin:
Yep, let's hold it. Should be able to start landing v7 on Wednesday (Nov. 6)

app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
```diff
@@ -595,6 +595,26 @@ private static void queueActiveWindowPctEvent() {
...
+                openPrivateWindows[number -1 ] = 1;
+            }
+
+        } else {
+            for (int i=0; i<openWindows.length; i++) {
+                openWindows[i] = 0;
+            }
+
+            if (number > 0) {
+                openWindows[number - 1] = 1;
```
daoshengmu: Same here.

app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
```diff
@@ -595,6 +595,26 @@ private static void queueActiveWindowPctEvent() {
         }
     }
 
+    public static void resetOpenedWindowsCount(int number, boolean isPrivate) {
+        if (isPrivate) {
+            for (int i=0; i<openPrivateWindows.length; i++) {
+                openPrivateWindows[i] = 0;
+            }
+            if (number > 0) {
+                openPrivateWindows[number -1 ] = 1;
```
daoshengmu: It makes sense to me for resetting to 0, but giving 1 looks odd to me

daoshengmu:
I also don't understand why making them to be 1 [1]. Ideally, `openWindows[index]++` only happens when creating a new window. But in your code, `openWindows[index] = 1;` at somewhere that with no reasons.

[1] https://github.com/MozillaReality/FirefoxReality/blob/51a797f0a5b4bbea87550256bb6c67dd19e95ca4/app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java#L624-L630

bluemarvin:
Moved to v7 so we can focus on the four other remaining PRs for v6

app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java
keianhzo: The `openWindows` and `openPrivateWindows` have three buckets for 1, 2 or 3 opened windows during the application lifetime so we can calculate the weighted mean later. `resetOpenedWindowsCount` gets the current windows number as a parameter (as it happens upon onresume), so when resetting we set that bucket to 1 as that's the initial windows count.

keianhzo: This also applies to above and below comments. `openWindows` has three buckets so when setting to one we are basically initializing the bucket for that windows number.

daoshengmu: nit: a redundant space beyond `]`

app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/Windows.java
```diff
@@ -426,6 +426,9 @@ public void onResume() {
         for (WindowWidget window: mPrivateWindows) {
             window.onResume();
         }
+
+        TelemetryWrapper.resetOpenedWindowsCount(mRegularWindows.size(), false);
+        TelemetryWrapper.resetOpenedWindowsCount(mPrivateWindows.size(), true);
```
daoshengmu: @keianhzo  Here still looks odd to me.
Like you said, `when openWindows has three buckets so when setting to one we are basically initializing the bucket for that windows number.`
If so, for example, when mRegularWindows(2) to do initialize, you need to make
openWindows[0] = 1;
openWindows[1] = 1;
But, you only make openWindows[2 -1] = 1;

keianhzo:
@daoshengmu Updated

keianhzo:
@daoshengmu I think this can be merged

PR #2198, opened 2019-11-08
Making WaveVR/OculusVR controllers support haptic feedback

### LINKED ISSUE #173: Support haptic feedback in immersive mode of HTC Vive Focus
This will be dependent on Bugzilla Bug 1473397 - Add Haptic Feedback support to gfxVRExternal and VRSessionOpenVR

daoshengmu:
Fixes #173. In this PR, we start to support gamepad haptics feedback in the immersive mode. We also can use `ControllerContainer->SetHapticsFeedback()` to give parameters to ask perform vibration for other usages.

@bluemarvin One of my My Vive Focus controllers seems to be broken, so please help verify if it works to you. Also, it shows `W/Wave_SDK: WVR_TriggerVibration with API Level 4 is not supported by the current Runtime version 3` when using `WVR_TriggerVibration` even though I have updated WaveSDK from [sdk-updates](https://github.com/MozillaReality/FirefoxReality-android-third-party/tree/sdk-updates).

### DIFF
--- a/app/src/main/cpp/BrowserWorld.cpp
+++ b/app/src/main/cpp/BrowserWorld.cpp
@@ -871,6 +871,7 @@ BrowserWorld::Draw() {
   m.device->ProcessEvents();
   m.context->Update();
   m.externalVR->PullBrowserState();
+  m.externalVR->SetHapticState(m.controllers);
 
   m.CheckExitImmersive();
   if (m.splashAnimation) {
--- a/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
+++ b/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
@@ -644,6 +644,9 @@ struct DeviceDelegateOculusVR::State {
     ovrInputTrackedRemoteCapabilities capabilities = {};
     vrb::Matrix transform = vrb::Matrix::Identity();
     ovrInputStateTrackedRemote inputState = {};
+    uint64_t inputFrameID = 0;
+    float remainingVibrateTime = 0.0f;
+    double lastHapticUpdateTimeStamp = 0.0f;
 
     bool Is6DOF() const {
       return (capabilities.ControllerCapabilities & ovrControllerCaps_HasPositionTracking) &&
@@ -691,7 +694,6 @@ struct DeviceDelegateOculusVR::State {
   vrb::Matrix reorientMatrix = vrb::Matrix::Identity();
   device::CPULevel minCPULevel = device::CPULevel::Normal;
   device::DeviceType deviceType = device::UnknownType;
-
 
   void UpdatePerspective() {
     float fovX = vrapi_GetSystemPropertyFloat(&java, VRAPI_SYS_PROP_SUGGESTED_EYE_FOV_DEGREES_X);
@@ -952,10 +954,13 @@ struct DeviceDelegateOculusVR::State {
             controller->CreateController(controllerState.index, int32_t(controllerState.hand),
                                          controllerName, beamTransform);
             controller->SetButtonCount(controllerState.index, 6);
+            controller->SetHapticCount(controllerState.index, 1);
           } else {
             // Oculus Go only has one kind of controller model.
             controller->CreateController(controllerState.index, 0, "Oculus Go Controller");
             controller->SetButtonCount(controllerState.index, 2);
+            // Oculus Go has no haptic feedback.
+            controller->SetHapticCount(controllerState.index, 0);
           }
           controllerState.created = true;
         }
@@ -1037,7 +1042,6 @@ struct DeviceDelegateOculusVR::State {
           const bool yTouched = (controllerState.inputState.Touches & ovrTouch_Y) != 0;
           const bool menuPressed = (controllerState.inputState.Buttons & ovrButton_Enter) != 0;
 
-
           controller->SetButtonState(controllerState.index, ControllerDelegate::BUTTON_X, 3, xPressed, xTouched);
           controller->SetButtonState(controllerState.index, ControllerDelegate::BUTTON_Y, 4, yPressed, yTouched);
 
@@ -1093,6 +1097,58 @@ struct DeviceDelegateOculusVR::State {
       controller->SetButtonState(controllerState.index, ControllerDelegate::BUTTON_TOUCHPAD, 0, trackpadPressed, trackpadTouched);
 
       controller->SetAxes(controllerState.index, axes, kNumAxes);
+      if (controller->GetHapticCount(controllerState.index)) {
+        UpdateHaptics(controllerState);
+      }
+    }
+  }
+
+  void UpdateHaptics(ControllerState& controllerState) {
+    vrb::RenderContextPtr renderContext = context.lock();
+    if (!renderContext) {
+      return;
+    }
+    if (!controller || !ovr) {
+      return;
+    }
+
+    uint64_t inputFrameID = 0;
+    float pulseDuration = 0.0f, pulseIntensity = 0.0f;
+    controller->GetHapticFeedback(controllerState.index, inputFrameID, pulseDuration, pulseIntensity);
+    if (inputFrameID > 0 && pulseIntensity > 0.0f && pulseDuration > 0) {
+      if (controllerState.inputFrameID != inputFrameID) {
+        // When there is a new input frame id from haptic vibration,
+        // that means we start a new session for a vibration.
+        controllerState.inputFrameID = inputFrameID;
+        controllerState.remainingVibrateTime = pulseDuration;
+        controllerState.lastHapticUpdateTimeStamp = renderContext->GetTimestamp();
+      } else {
+        // We are still running the previous vibration.
+        // So, it needs to reduce the delta time from the last vibration.
+        const double timeStamp = renderContext->GetTimestamp();
+        controllerState.remainingVibrateTime -= (timeStamp - controllerState.lastHapticUpdateTimeStamp);
+        controllerState.lastHapticUpdateTimeStamp = timeStamp;
+      }
+
+      if (controllerState.remainingVibrateTime > 0.0f && renderMode == device::RenderMode::Immersive) {
+        if (vrapi_SetHapticVibrationSimple(ovr, controllerState.deviceId, pulseIntensity > 1.0f ? 1.0f : pulseIntensity)
+            == ovrError_InvalidOperation) {
+          VRB_ERROR("vrapi_SetHapticVibrationBuffer failed.");
+        }
+      } else {
+        // The remaining time is zero or exiting the immersive mode, stop the vibration.
+        if (vrapi_SetHapticVibrationSimple(ovr, controllerState.deviceId, 0.0f) == ovrError_InvalidOperation) {
+          VRB_ERROR("vrapi_SetHapticVibrationBuffer failed.");
+        }
+        controllerState.remainingVibrateTime = 0.0f;
+      }
+    } else if (controllerState.remainingVibrateTime > 0.0f) {
+      // While the haptic feedback is terminated from the client side,
+      // but it still have remaining time, we need to ask for stopping vibration.
+      if (vrapi_SetHapticVibrationSimple(ovr, controllerState.deviceId, 0.0f) == ovrError_InvalidOperation) {
+        VRB_ERROR("vrapi_SetHapticVibrationBuffer failed.");
+      }
+      controllerState.remainingVibrateTime = 0.0f;
     }
   }
 
--- a/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
+++ b/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
@@ -1066,7 +1066,8 @@ struct DeviceDelegateOculusVR::State {
           VRB_WARN("Undefined hand type in DeviceDelegateOculusVR.");
         }
 
-        const bool thumbRest = (controllerState.inputState.Touches & ovrTouch_ThumbUp) != 0;
+        // This is always false in Oculus Browser.
+        const bool thumbRest = false;
         controller->SetButtonState(controllerState.index, ControllerDelegate::BUTTON_OTHERS, 5, thumbRest, thumbRest);
       } else {
         triggerPressed = (controllerState.inputState.Buttons & ovrButton_A) != 0;
--- a/app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
+++ b/app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
@@ -45,6 +45,9 @@ struct DeviceDelegateWaveVR::State {
     int32_t gripPressedCount;
     vrb::Matrix transform;
     ElbowModel::HandEnum hand;
+    uint64_t inputFrameID;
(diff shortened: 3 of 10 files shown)


daoshengmu:
Here is the test case, https://webvr.info/samples/XX-vr-controllers.html

bluemarvin:
> Here is the test case, https://webvr.info/samples/XX-vr-controllers.html

testing it now.

app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
```diff
@@ -350,6 +352,26 @@ struct DeviceDelegateWaveVR::State {
...
+      float pulseDuration = 0.0f, pulseIntensity = 0.0f;
+      delegate->GetHapticsFeedback(controller.index, inputFrameID, pulseDuration, pulseIntensity);
+      if (inputFrameID == 0 || pulseDuration <= 0 ) {
+        // Stopping haptics feedback.
+        if (controller.doingVibration) {
+          WVR_TriggerVibration(controller.type, WVR_InputId_Max, 0.0f, 0);
+          controller.doingVibration = false;
+        }
+      } else if (pulseIntensity > 0.0f) {
+        controller.doingVibration = true;
```
bluemarvin: I'm not sure what you need `doingVibration` for. Is it so you only stop it once? Why do we need to stop it since the duration is one of the parameter?

app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
```diff
@@ -350,6 +352,26 @@ struct DeviceDelegateWaveVR::State {
...
+        // THe duration time unit needs to be transformed from milliseconds to microseconds.
+        // The gamepad extensions API does not yet have independent control
+        // of frequency and intensity. It only has vibration value (0.0 ~ 1.0).
+        // In this WaveVR SDK, the value makes more sense to be intensity because frequency can't
+        // < 1.0 here.
+        WVR_TriggerVibration(controller.type, WVR_InputId_Max, pulseDuration * 1000.0f, 1, pulseIntensity > 0.5f ? WVR_Intensity_Normal : WVR_Intensity_Strong);
```
bluemarvin: Since it has five intensities wouldn't it be better to map the 0.0 to 1.0 rang into fifths? That way you get all the intensities. something like:
```
int intensity = ceil(pulseIntensity * 5);
intensity <= 5 ? intensity : 5;
```

app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -858,6 +858,14 @@ BrowserWorld::Draw() {
...
   m.externalVR->PullBrowserState();
 
+  const uint32_t count = m.controllers->GetControllerCount();
+  for (uint32_t i = 0; i < count; ++i) {
+    float pulseStart = 0, pulseDuration = 0, pulseIntensity = 0;
+    uint64_t inputFrameID = 0;
+    m.externalVR->GetHapticState(i, inputFrameID, pulseStart, pulseDuration, pulseIntensity);
+    m.controllers->SetHapticsFeedback(i, inputFrameID, pulseDuration + pulseStart, pulseIntensity);
+  }
+
```
bluemarvin: We should probably only be processing these when in immersive mode since there shouldn't be any controllers available at other times. I would also rather it look something like this: `m.externalVR->SetHapticsState(m.controllers);`

app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
daoshengmu: Ok. We can remove it because the stop function actually doesn't work for Focus+.
```
WVR_TriggerVibration(WVR_DeviceType_Controller_Left, WVR_InputId_Max, 10000000, 1);
WVR_TriggerVibration(WVR_DeviceType_Controller_Left, WVR_InputId_Max, 0, 0);
```

app/src/main/cpp/BrowserWorld.cpp
daoshengmu: I was thinking if our UX are interested in using haptics feedback when doing interaction like keyboard typing, etc...If so, we can just call `ControllerContainer->SetHapticsFeedback()` and set parameters, this API will be implemented in OculusVR as well.

bluemarvin: We will probably use the controller API with UI but I would rather that for loop be in the externalVR code and then only have it called when in immersive mode.

daoshengmu:
I haven't had a chance to use Focus to take a try because someone borrowed it. I suppose it should have no problem since I tested the vibration API on Focus.

daoshengmu:
Added Oculus Quest haptic feedback support. `vrapi_SetHapticVibrationSimple` usually do vibrate for a couple of seconds. So, we need to stop it once WebVR content already stops the haptic feedback. Its API allows us to stop it by giving the intensity to `0`.

For testing, please make `thumbRest: false` at
`controller->SetButtonState(controllerState.index, ControllerDelegate::BUTTON_OTHERS, 5, thumbRest, thumbRest);`

I also consider if we need to move the `GetTimeStamp()` to our util library.

daoshengmu:
@bluemarvin I think we can continue to review it again.

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -1061,6 +1072,46 @@ struct DeviceDelegateOculusVR::State {
 
         const bool thumbRest = (controllerState.inputState.Touches & ovrTouch_ThumbUp) != 0;
         controller->SetButtonState(controllerState.index, ControllerDelegate::BUTTON_OTHERS, 5, thumbRest, thumbRest);
+
+        // Only Oculus 6 DOF (Quest) has haptic feedback.
+        controller->SetHapticCount(controllerState.index, 1);
```
MortimerGoro: `UpdateControllers` is becoming very big. Can you move the haptics code to a separate function? (e.g. `UpdateHaptics`)

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -965,6 +967,16 @@ struct DeviceDelegateOculusVR::State {
     }
   }
 
+  double GetTimeStamp() {
```
MortimerGoro: You can use `m.context->GetTimestamp()` instead of duplicating this function

MortimerGoro:
@daoshengmu haptic doesn't work correctly in Oculus. Check this demo: https://webvr.info/samples/XX-vr-controllers.html

On Oculus Browser the controller vibrates when clicking any button. It doesn't do anything in FxR.

MortimerGoro:
After conversation with @bluemarvin and @daoshengmu we decided to move it to v8. Specially because Oculus requires the haptic to be shut off explicitly and there are some risks if there is some edge case where that isn't done correctly or an app sending a big times, etc

daoshengmu:
> @daoshengmu haptic doesn't work correctly in Oculus. Check this demo: https://webvr.info/samples/XX-vr-controllers.html
> 
> On Oculus Browser the controller vibrates when clicking any button. It doesn't do anything in FxR.

This is my bad. I should update my Oculus SDK first. It works now after using my latest update.

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -1096,6 +1105,48 @@ struct DeviceDelegateOculusVR::State {
...
+    vrb::RenderContextPtr renderContext = context.lock();
+
+    controller->GetHapticFeedback(controllerState.index, inputFrameID, pulseDuration, pulseIntensity);
+    if (inputFrameID > 0 && pulseIntensity > 0.0f && pulseDuration > 0) {
+      if (controllerState.inputFrameID != inputFrameID) {
+        // When there is a new input frame id from haptic vibration,
+        // that means we start a new session for a vibration.
+        controllerState.inputFrameID = inputFrameID;
+        controllerState.remainingVibrateTime = pulseDuration;
+        controllerState.lastHapticUpdateTimeStamp = renderContext->GetTimestamp();//GetTimeStamp();
```
bluemarvin: Please remove commented code.

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -1062,8 +1063,13 @@ struct DeviceDelegateOculusVR::State {
           VRB_WARN("Undefined hand type in DeviceDelegateOculusVR.");
         }
 
-        const bool thumbRest = (controllerState.inputState.Touches & ovrTouch_ThumbUp) != 0;
+        // This is always false in Oculus Browser.
+        const bool thumbRest = false; //(controllerState.inputState.Touches & ovrTouch_ThumbUp) != 0;
```
bluemarvin: Please clean up commented code.

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -1096,6 +1105,48 @@ struct DeviceDelegateOculusVR::State {
     }
   }
 
+  void UpdateHaptics(ControllerState& controllerState) {
+    uint64_t  inputFrameID = 0;
+    float pulseDuration = 0.0f, pulseIntensity = 0.0f;
+    vrb::RenderContextPtr renderContext = context.lock();
```
bluemarvin: locking a weak pointer can result in `renderContext` being a `nullptr`

bluemarvin:
I was thinking we should also have a case for the Oculus where if we exit immersive mode and the haptics are still active they should be stopped.

daoshengmu:
> I was thinking we should also have a case for the Oculus where if we exit immersive mode and the haptics are still active they should be stopped.

Good catch! I have confirmed in that case, they are `inputFrameID > 0 && pulseIntensity > 0.0f && pulseDuration > 0`, so we need to check if it exists the immersive mode.

Please use this link to take a try,
https://daoshengmu.github.io/webvr.info/samples/XX-vr-controllers-long-haptics.html

bluemarvin:
Something still isn't right. On the Quest When I visit https://webvr.info/samples/XX-vr-controllers.html
STR:
1) Enter immersive mode
2) Press and hold the left trigger
3) Press and hold the right trigger
3) Release the left trigger

Expected:
Right controller continues to vibrate

Actual:
Right controller stops vibrating. Pressing the left controller trigger while continuing to hold the right trigger will actually case the right controller to vibrate, but it will only vibrate while the left controller trigger is pressed.

Notes:
The opposite is not true where pressing the right trigger first does not affect the left controller.

bluemarvin:
I can reproduce the issue with the HTC Focus Plus as well.

bluemarvin:
For the focus it seems to vary based on what the system thinks is the first and second controller since it doesn't know left from right.

bluemarvin:
Also, when using https://daoshengmu.github.io/webvr.info/samples/XX-vr-controllers-long-haptics.html on the Focus plus, exiting immersive mode does not stop the long vibration. When I quit FxR it does stop so there is a way to stop it.

daoshengmu:
> Also, when using https://daoshengmu.github.io/webvr.info/samples/XX-vr-controllers-long-haptics.html on the Focus plus, exiting immersive mode does not stop the long vibration. When I quit FxR it does stop so there is a way to stop it.

It seems like Wave SDK gave a update recently, it can stop the vibration through 
`WVR_TriggerVibration(controller.type, WVR_InputId_Max, 0, 0, (WVR_Intensity_Normal));`

Please use this link to confirm,
https://daoshengmu.github.io/webvr.info/samples/XX-vr-controllers-stop-haptics.html

It will do vibration for 8 sec when press a button, once the button is release, it will stop the vibration.

daoshengmu:
> Also, when using https://daoshengmu.github.io/webvr.info/samples/XX-vr-controllers-long-haptics.html on the Focus plus, exiting immersive mode does not stop the long vibration. When I quit FxR it does stop so there is a way to stop it.

For this case, it is because `hapticState` are reset to 0 when exiting from immersive mode,  even `controllerIndex`. It makes us difficult to use `controllerIndex` to update the haptic status from VRShmem. Hopefully, we still can check `inputFrameID != 0` to assign its status.

app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
```diff
@@ -350,6 +357,55 @@ struct DeviceDelegateWaveVR::State {
...
+      UpdateHaptics(controller);
+    }
+  }
+
+  void UpdateHaptics(Controller& controller) {
+    vrb::RenderContextPtr renderContext = context.lock();
+    if (!renderContext) {
+      return;
+    }
+
```
bluemarvin: I think you need a
```
if (!delegate) {
  return;
}
```

app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
```diff
@@ -350,6 +357,55 @@ struct DeviceDelegateWaveVR::State {
...
+                             1, WVR_Intensity(intensity));
+      } else {
+        // The remaining time is zero or exiting the immersive mode, stop the vibration.
+        WVR_TriggerVibration(controller.type, WVR_InputId_Max, 0, 0, WVR_Intensity_Normal);
+        controller.remainingVibrateTime = 0.0f;
+      }
+    } else if (controller.remainingVibrateTime > 0.0f) {
+      // While the haptic feedback is terminated from the client side,
+      // but it still have remaining time, we need to ask for stopping vibration.
+      WVR_TriggerVibration(controller.type, WVR_InputId_Max, 0, 0, (WVR_Intensity_Normal));
```
bluemarvin: Extra `()` around `WVR_Intensity_Normal`

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -1093,6 +1098,55 @@ struct DeviceDelegateOculusVR::State {
...
+      }
+    }
+  }
+
+  void UpdateHaptics(ControllerState& controllerState) {
+    vrb::RenderContextPtr renderContext = context.lock();
+    if (!renderContext) {
+      return;
+    }
+
```
bluemarvin:
```
if (!controller || !ovr) {
  return;
}
```

app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp
```diff
@@ -350,6 +357,62 @@ struct DeviceDelegateWaveVR::State {
...
+        // of frequency and intensity. It only has vibration value (0.0 ~ 1.0).
+        // In this WaveVR SDK, the value makes more sense to be intensity because frequency can't
+        // < 1.0 here.
+        int intensity = ceil(pulseIntensity * 5);
+        intensity = intensity <= 5 ? intensity : 5;
+        WVR_TriggerVibration(controller.type, WVR_InputId_Max, controller.remainingVibrateTime * 1000.0f,
+                             1, WVR_Intensity(intensity));
+      } else {
+        // The remaining time is zero or exiting the immersive mode, stop the vibration.
+  #if !defined(__arm__) // It will crash at WaveVR SDK arm32, let's skip it.
```
bluemarvin: Remove extra space

PR #2233, opened 2019-11-12
Fix url bar focus issues

keianhzo:
Force URL bar top parent to gain focus after a URL is loaded to avoid children to get it and make the URL bar to look focused.

STRs:
- Login FxA
- Click on the URL bar
- Click on the Hamburger menu
- Share a tab from another device
- Click the home button
- Resume the app

Actual result:
- The URL bar is in a focused state

Expected result:
- The URL bar shouldn't be in a focused state


keianhzo:
@MortimerGoro I've updated. It seems to be related to the bindings execution. We make an async request for setting the button as bookmarked and after that we set the bookmarked binding. That for some reason makes the button to gain focus. I've tried other approaches forcing the parent layout be focusable and get focus before descendants but I think the bindings execution takes precedence so clearing the focus after the binding is executed seems to be the only working solution.

PR #2254, opened 2019-11-14
Only create one UserAgentOverride object

bluemarvin:


app/src/common/shared/org/mozilla/vrbrowser/browser/engine/Session.java
```diff
@@ -141,9 +142,9 @@ private void initialize() {
         InternalPages.PageResources pageResources = InternalPages.PageResources.create(R.raw.private_mode, R.raw.private_style);
         mPrivatePage = InternalPages.createAboutPage(mContext, pageResources);
 
-        if (mUserAgentOverride == null) {
-            mUserAgentOverride = new UserAgentOverride();
-            mUserAgentOverride.loadOverridesFromAssets((Activity)mContext, mContext.getString(R.string.user_agent_override_file));
+        if (sUserAgentOverride == null) {
```
MortimerGoro: I think only checking null for initialization can be risky. What happens if the activity is recreated but the static instance is still alive. The `Context` instance could be different but the `UserAgentOverride` would still keep the old reference

MortimerGoro: We can resolve this cleaning the static variable in `Activity.Ondestroy`, that way we also avoid leaking the activity

bluemarvin: The UserAgentOverride class doesn't hold a reference to the activity. So I don't think anything is getting leaked. Recreating the activity should be fine since the config file will not have changed so there is no reason to read it again.

bluemarvin: Additionally, the class has a cache so keeping it between runs is probably a good thing.

MortimerGoro: ok, if it doesn't hold any reference then is fine

PR #2342, opened 2019-11-21
Don't keep the text selected after it's copied.

### LINKED ISSUE #2215: Static text should not remain highlighted after it's copied
Question from QA: `Should the static text remain highlighted after it's copied? On desktop it remains highlighted and on mobile it's not and I wanted to be sure on what the right behavior should be in FxR`

From what I have tried:
- Desktop: it remains highlighted
- iPhone mobile: it remains highlighted
- Android mobile: it's dimissed

The context menu can also be cleared using the back button, But just found that the back button clears if there is a Context Menu but not if there is a selection without context menu.

We need a UX decision
**Decision is: highlight should dismiss after copy**

MortimerGoro:
fixes #2215

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WindowWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WindowWidget.java
@@ -1676,6 +1676,11 @@ public class WindowWidget extends UIWidget implements SessionChangeListener,
                 if (aSelection.isActionAvailable(action)) {
                     aSelection.execute(action);
                 }
+                if (GeckoSession.SelectionActionDelegate.ACTION_COPY.equals(action) &&
+                        aSelection.isActionAvailable(GeckoSession.SelectionActionDelegate.ACTION_UNSELECT)) {
+                    // Don't keep the text selected after it's copied.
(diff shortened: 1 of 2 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WindowWidget.java
```diff
@@ -1676,6 +1676,11 @@ public void onAction(String action) {
                 if (aSelection.isActionAvailable(action)) {
                     aSelection.execute(action);
                 }
+                if (GeckoSession.SelectionActionDelegate.ACTION_COPY.equals(action) &&
+                        aSelection.isActionAvailable(GeckoSession.SelectionActionDelegate.ACTION_UNSELECT)) {
+                    // Don't keep the text selected after it's copied.
+                    aSelection.execute(GeckoSession.SelectionActionDelegate.ACTION_UNSELECT);
```
bluemarvin: One thing we can do in a follow up is use the shorthand calls, so You can just call `aSelection.unselect()` instead of calling the more verbose `execute()`. The fact that you have to check if the action is available sort of sucks though.

PR #2343, opened 2019-11-21
Invert horizontal scrolling

### LINKED ISSUE #2045: Horizontal scroll is inverted
Currently the Horizontal scroll is inverted.
If you go to any code example on glitch.com or any page that is big enough to need horizontal scrolling, you can realise that the vertical scroll works as expected (in the direction you move the joystick) but the horizontal scroll is inverted (pressing right moves the scroll to left).
Tested also on Oculus Browser and it works as expected.

MortimerGoro:
Fixes #2045

### DIFF
--- a/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
+++ b/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
@@ -1025,7 +1025,7 @@ struct DeviceDelegateOculusVR::State {
         trackpadY = controllerState.inputState.Joystick.y;
         axes[0] = trackpadX;
         axes[1] = -trackpadY; // We did y axis intentionally inverted in FF desktop as well.
-        controller->SetScrolledDelta(controllerState.index, trackpadX, trackpadY);
+        controller->SetScrolledDelta(controllerState.index, -trackpadX, trackpadY);
 
         const bool gripPressed = (controllerState.inputState.Buttons & ovrButton_GripTrigger) != 0;
         controller->SetButtonState(controllerState.index, ControllerDelegate::BUTTON_OTHERS, 2, gripPressed, gripPressed,


app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
```diff
@@ -843,7 +843,7 @@ void handleScrollEvent(final int aHandle, final int aDevice, final float aX, fin
             }
             if (widget != null) {
                 float scrollDirection = mSettings.getScrollDirection() == 0 ? 1.0f : -1.0f;
-                MotionEventGenerator.dispatchScroll(widget, aDevice, aX * scrollDirection, aY * scrollDirection);
+                MotionEventGenerator.dispatchScroll(widget, aDevice, -aX * scrollDirection, aY * scrollDirection);
```
bluemarvin: I think we only want to invert it for the Quest joystick, I think it is correct for touch pad based controllers.

bluemarvin: I've verified that it should only be inverted for joysticks. Not sure where that fix belongs. Our horizontal scrolling on Go is the same as the Oculus browser.

MortimerGoro: Done. I also verified that WebVR axes matches the Oculus behavior

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -1025,7 +1025,7 @@ struct DeviceDelegateOculusVR::State {
         trackpadY = controllerState.inputState.Joystick.y;
         axes[0] = trackpadX;
         axes[1] = -trackpadY; // We did y axis intentionally inverted in FF desktop as well.
-        controller->SetScrolledDelta(controllerState.index, trackpadX, trackpadY);
+        controller->SetScrolledDelta(controllerState.index, -trackpadX, trackpadY);
```
bluemarvin: It's weird that we invert `Y` for immersive and `X` for browsing.

PR #2351, opened 2019-11-22
Fixes UIDialog focus issues

### LINKED ISSUE #2329: The "Send Tab to Device" prompt is auto dismissed when trying to send a tab from the Hamburger menu after the app is resumed
**[Affected Versions]:**
- Firefox Reality version 7 RC2 (ff64d298)

**[Affected Platforms]:**
- Oculus Go - Android 7.1.2

**[Prerequisites]:**
- Make sure that the FxA account is logged on the VR device and also on a second device.

**[Steps to reproduce]:**
1. Click the "Oculus" button from the controller.
2. Click the "Resume" button from the prompt displayed.
3. Click the Hamburger button.
4. Click the "Send Tab to Device" button from the Hamburger menu.
5. Observe the "Send Tab to Device" prompt.

**[Expected results]:**
- The prompt is displayed and remains displayed.

**[Actual results]:**
- The prompt is briefly displayed and after it's auto dismissed.

**[Notes]:**
- The issue is reproducible when trying to send a tab after the device exits sleep mode.
- The issue is also reproducible after reopening the app.
- This issue is not reproducible if the tab is sent from "View Tabs".
- The issue is no longer reproducible after trying to send a tab on a second attempt.
- I was not able to reproduce this issue on the HTC Vive Focus devices.
- Attached a screen recording of the issue: [here](https://drive.google.com/open?id=1Sds8T7xLEYey-YwssFToHIEHOsrFRu0w).

keianhzo:
Fixes #2329 This PR fixes the following issues:
- The "Send Tab to Device" prompt is auto dismissed when trying to send a tab from the Hamburger menu after the app is resumed issue should be fixed now.
- UIDialog focus. After coming from suspension the focus was being requested too early and for dialogs like the NoInternet dialog (that are shown right after the activity is resumed) the focus request was not being processed. Now we enqueue the focus request so it is correctly processed in those cases.
- Adds a world click listener to the UIDialog. Most dialogs were already adding this so makes sense to refactor it a to a common place. Also there is an issue when resuming the activity twice. In that case the app focus is reset and old focus is null so dialogs couldn't be dismissed clicking in the world. It can be tested in the voice search dialog showing it and going home twice in a row.
- Common behavior for dimming the world. Some dialogs are dimming the world some others aren't. This standardizes that behavior, now all the dialogs dim the world.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/dialogs/UIDialog.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/dialogs/UIDialog.java
@@ -8,7 +8,7 @@ import org.mozilla.vrbrowser.ui.widgets.UIWidget;
 import org.mozilla.vrbrowser.ui.widgets.WidgetManagerDelegate;
 import org.mozilla.vrbrowser.utils.ViewUtils;
 
-public abstract class UIDialog extends UIWidget implements WidgetManagerDelegate.FocusChangeListener {
+public abstract class UIDialog extends UIWidget implements WidgetManagerDelegate.FocusChangeListener, WidgetManagerDelegate.WorldClickListener {
     public UIDialog(Context aContext) {
         super(aContext);
         initialize();
@@ -26,10 +26,12 @@ public abstract class UIDialog extends UIWidget implements WidgetManagerDelegate
 
     private void initialize() {
         mWidgetManager.addFocusChangeListener(this);
+        mWidgetManager.addWorldClickListener(this);
     }
 
     @Override
     public void releaseWidget() {
+        mWidgetManager.removeWorldClickListener(this);
         mWidgetManager.removeFocusChangeListener(this);
         super.releaseWidget();
     }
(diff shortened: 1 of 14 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/dialogs/UIDialog.java
```diff
@@ -39,13 +41,33 @@ public boolean isDialog() {
...
+
+        mWidgetManager.popWorldBrightness(this);
+    }
+
     // WidgetManagerDelegate.FocusChangeListener
 
     @Override
     public void onGlobalFocusChanged(View oldFocus, View newFocus) {
-        if (!ViewUtils.isEqualOrChildrenOf(this, newFocus) && isVisible()) {
+        if (oldFocus == this && isVisible()) {
```
MortimerGoro: Is this reliable? what happens if a children view gets the focus?

bluemarvin: Do we know why `ViewUtils.isEqualOrChilderenOf()` was failing on the first open of the share dialog?

bluemarvin: If this doesn't cause any known issues I think we should land it and address in a follow up.

keianhzo: @bluemarvin was failing because when showing the sendtabs dialog we hide the radiobutton while the devices are loading. That made the activity change the focus to null but only if the activity has been paused as the activity seems to reset the focus to null in that case.

@MortimerGoro This is how it was before https://github.com/MozillaReality/FirefoxReality/commit/d0cd20b151acbe579564b23a7f1fd90f707b90ac#diff-4f3b7fc6e7c2e6cb9d5359555ffe7e67R46 but for what Ive seen it wasn't reliable and most dialogs just used on the world click listener to hide. 

Actually I've just removed the global focus change listener from UIDialog as it's not really necessary.

Issue #1955, opened 2019-10-11
The History view is dismissed when the background website finishes loading


Softvision-GeluHaiduc:
**[Affected Versions]:**
- Firefox Reality version 5-32821846 RC1 (58b59br3)

**[Affected Platforms]:**
- Oculus Quest - Android 7.1.1
- HTC Vive Focus Plus - Android 7.1.2

**[Prerequisites]:**
- The headset is turned on and the browser is opened.

**[Steps to reproduce]:**
1. Navigate to any website that takes longer to load. (e.g. amazon.com, yahoo.com, huffpost.com/)
2. Click the History button from the menu tray while the website is loading.
3. Observe the History view.

**[Expected results]:**
- The History view remains displayed.

**[Actual results]:**
- The History view is dismissed after the webpage from the background is loaded.

**[Additional Notes]:**
- This issue is not reproducible with the “Bookmarks” page.
- Attached a screen recording of the issue [here](https://drive.google.com/file/d/1aLGU6ZdGrofC0lMaNmW93YwaTmK86Ddp/view?usp=sharing).

keianhzo:
We use navigation callbacks to dismiss the library panels when the user initiates a navigation event to avoid to explicitly add callbacks from all the possible navigation buttons (reload, home, back, forward, etc).

We have updated the navigation callback to use the faster `PageStart` instead of `onLocationChange` so this should be harder to reproduce now. In case it's not or we think we need faster feedback we can always add explicit callbacks from all the buttons.

keianhzo:
I've added a PR for explicit panel dismiss. That should help speed up the panel dismiss after clicking on an item and also fix this issue.

Softvision-GeluHaiduc:
I have verified that this issue is no longer reproducible with the latest Firefox Reality version 7 RC3 (7b66fb14) installed on the Oculus Go, Quest, HTC Vive Focus and Focus Plus headsets.
Now, the History view is not dismissed when the background website finishes loading.

PR #2367, opened 2019-11-25
Use LoadRequest event to close library panels

### LINKED ISSUE #2366: The History view is not dismissed when navigating to another website using the awesomebar
**[Affected Versions]:**
- Firefox Reality version 7 RC3 (7b66fb14)

**[Affected Platforms]:**

**[Prerequisites]:**

**[Steps to reproduce]:**
1. Click the History button from the menu tray.
2. Enter a URL in the awesome bar (e.g. youtube.com).
3. Observe the behavior.

**[Expected results]:**
- The History view is dismissed and the page is loaded.

**[Actual results]:**
- The History view is not dismissed and the page is loaded in the background.

**[Notes]:**
- If you enter a partial URL and access the suggested web page, the awesome bar will continue to display the partial input until the History view is dismissed.
- Attached is a recording of the [issue](https://drive.google.com/file/d/1e_u7B9sJpvew-GTpdDiEqLrxH9V0lDFD/view?usp=sharing).

keianhzo:
Fixes #2366 This reverts https://github.com/MozillaReality/FirefoxReality/pull/2353 and moves the library panels dismiss to the `LoadRequest` event as it's the first event we get when a page load starts. 

This will cause a regression in https://github.com/MozillaReality/FirefoxReality/issues/1955 but there is really no better way of handling this as we need to close the panels with a new URL is loaded and that could happen from multiple places: URL buttons (back, forward, reload), Settings help/survey, history/bookmark links, voice search, etc. So looking again at the problem I think this is still our best option.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/callbacks/BookmarksCallback.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/callbacks/BookmarksCallback.java
@@ -13,5 +13,4 @@ public interface BookmarksCallback {
     default void onFxASynSettings(@NonNull View view) {}
     default void onShowContextMenu(@NonNull View view, Bookmark item, boolean isLastVisibleItem) {}
     default void onHideContextMenu(@NonNull View view) {}
-    default void onItemClicked(@NonNull View view, Bookmark item) {}
 }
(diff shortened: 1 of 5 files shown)


MortimerGoro:
@keianhzo the conflicts needs to be fixed before landing

keianhzo:
@MortimerGoro Rebased

Issue #2062, opened 2019-10-24
Settings for WebXR


jvonitter:
We'll need some settings to support WebXR.
I'm using Firefox desktop art here to get the point across, but @thenadj will deliver actually FxR specs for these. Should use the same paradigm as [popups blocking](https://github.com/MozillaReality/FirefoxReality/issues/593)

- [ ] new privacy permission for XR, defaulted to ON.
with link out to an informational page based on https://bugzilla.mozilla.org/show_bug.cgi?id=1579267
![InkedSecurity_And_Privacy_LI](https://user-images.githubusercontent.com/48839872/67528067-589c0980-f66d-11e9-9998-d352c92b6d3f.jpg)

- [ ] exceptions list. users can add a sight here to prevent them from using XR
![Permissions](https://user-images.githubusercontent.com/48839872/67528048-4e7a0b00-f66d-11e9-961e-e44092d20376.png)

This is related to #1966 - when we teach users how to get out of immersive mode, we should also include some messaging that tells them they can turn off immersive mode from Settings.

jvonitter:
![](https://github.trello.services/images/mini-trello-icon.png) [DE-78 Web XR Support](https://trello.com/c/KF11i74I/485-de-78-web-xr-support)

PR #2413, opened 2019-12-03
Add WebXR settings and permission flow

MortimerGoro:
Initial implementation of #2062 (draft design)
- WebXR permission is enabled by default for all websites
- You can handle individual WebXR site permissions using the new HMD icon in the nav bar (only shown in websites using WebVR/WebXR)
- If the global WebXR permission is disabled the HMD icon is no visible anymore in the nav bar (I didn't want to implement the visual for this case until the spec is final)
- When clicking on the HMD icon the original prompt is also shown for websites that has already been decided (I didn't want to implement different visuals for this case until the spec is final)
- The content feed shows the HMD icon (we could automaticallly hide it in this website if we want)


MortimerGoro:
Requires https://bugzilla.mozilla.org/show_bug.cgi?id=1580567 and https://bugzilla.mozilla.org/show_bug.cgi?id=1580567

MortimerGoro:
APK with the local GV compilation if you want to test it before all the GV bumps land in master

PR #2414, opened 2019-12-03
By passing WebXR permission prompt

daoshengmu:
Per discussion in Slack, we still need to land this if we are blocked by the UI design spec review at #2413.

This PR will come by a GV version update once https://bugzilla.mozilla.org/show_bug.cgi?id=1580567 and https://bugzilla.mozilla.org/show_bug.cgi?id=1580567 are landed.


bluemarvin:
This is only required if we need to upgrade to a version of GV that contains the change before the final solution is ready. This probably shouldn't have been posted until it could include the GV bump needed to actually compile it.

daoshengmu:
Agree! Let's close it and do it when updating GV version.

PR #2417, opened 2019-12-04
Omit redirections in history

### LINKED ISSUE #1839: Two or three history feeds are wrongly logged when URL is change via javascript
**[Affected Versions]:**
- Firefox Reality 1.5-32570012 (6090cf4)

**[Affected Platforms]:**
- Oculus Go - Android 7.1.2

**[Prerequisites]:**

**[Steps to reproduce]:**
1. Click the "+" button to open a new window (I opened a new window for a better observation of the issue).
2. Click the "History" button to open the "History" page.
3. Focus the other window and navigate to "www.youtube.com".
4. Click on a suggested video and observe the "History" page.

**[Expected results]:**
- A history feed is logged for the opened video.

**[Actual results]:**
- Two or three history feeds are logged for the opened video.

**[Notes]:**
- The logs have the same URL.
- Attached is a screen recording of the issue: [link](https://drive.google.com/open?id=19oP4kQI-bfkgv0nQlmM1LfUzD7fWxXUd)

keianhzo:
Fixes #1839 Omit history redirects. Now we should just log the url that initiated the request.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/HistoryStore.kt
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/HistoryStore.kt
@@ -73,14 +73,22 @@ class HistoryStore constructor(val context: Context) {
     fun getDetailedHistory(): CompletableFuture<List<VisitInfo>?> = GlobalScope.future {
         storage.getDetailedVisits(0, excludeTypes = listOf(
                 VisitType.NOT_A_VISIT,
+                VisitType.DOWNLOAD,
                 VisitType.REDIRECT_TEMPORARY,
+                VisitType.RELOAD,
+                VisitType.EMBED,
+                VisitType.FRAMED_LINK,
                 VisitType.REDIRECT_PERMANENT))
     }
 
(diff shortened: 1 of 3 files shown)


bluemarvin:
Does this actually fix the youtube problem since I think we only get `onLocationChange()` when it the history is manipulated directly. I know it is still a problem with twitch.tv.

keianhzo:
It seems to be working fine with Youtube and everything I've tried so far. We clear the redirect cache after a page is fully loaded so anything in between that is a redirect or comes from a redirect is ommited.

bluemarvin:
> It seems to be working fine with Youtube and everything I've tried so far. We clear the redirect cache after a page is fully loaded so anything in between that is a redirect or comes from a redirect is ommited.

Some redirects only come through `onLocationChange()`. Go to twitch.tv to see this in action.

bluemarvin:
When I visit twitch.tv it doesn't even show up in history now.

keianhzo:
@MortimerGoro @bluemarvin I've updated to handle this like Fenix does just using the page visited flags from GV.

bluemarvin:
It's cleaner but we will still need GV support for `onLocationChange()` calls.

PR #2423, opened 2019-12-05
Add support for tab/windows adb parameters

### LINKED ISSUE #2355: Add support for opening tabs from ADB
We currently support opening a URL from [ADB](https://github.com/MozillaReality/FirefoxReality/wiki/ADB-command-cheat-sheet) in the current active session/tab. but we don't currently support opening the URL in a new tab. This can useful for testing.

We can follow the same approach as chrome where you can do it through a `create_new_tab` boolean extra parameter:

`adb shell am start -a android.intent.action.VIEW --ez create_new_tab true -d https://www.mozilla.com/ -t application/browser -n org.mozilla.vrbrowser/org.mozilla.vrbrowser.VRBrowserActivity
`

Additionally we could also give the option to open it in background.

keianhzo:
Fixes #2355 Add support for tab/windows adb parameters. Also improves the way we handle intent parameters. Now you don't need to supply a data URI to change the homepage/e10s.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
@@ -483,9 +483,8 @@ public class VRBrowserActivity extends PlatformActivity implements WidgetManager
         setIntent(intent);
         final String action = intent.getAction();
         if (Intent.ACTION_VIEW.equals(action)) {
-            if (intent.getData() != null) {
-                loadFromIntent(intent);
-            }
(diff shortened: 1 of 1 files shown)


philip-lamb:
Nice addition. I'd suggest you document these new options on the wiki.

keianhzo:
@philip-lamb I already added it to https://github.com/MozillaReality/FirefoxReality/wiki/ADB-command-cheat-sheet. Should I add it somewhere else? Maybe I add it also to https://github.com/MozillaReality/FirefoxReality/wiki/Sideload

PR #2438, opened 2019-12-09
FxA WebChannels implementation

### LINKED ISSUE #2267: Migrate FxA implemetation to WebChannels instead of interceptor
Since we're using Gecko, we should use WebChannels for communicating with the FxA web content, instead of intercepting URL loads.

Based on this comment: https://github.com/MozillaReality/FirefoxReality/pull/1973#issuecomment-544984467

**Additional context**
Once we integrate the webchannels support, for free we get handling of "Choose what to sync" selection during FxA sign-up, which our current interceptor-based integration doesn't support. Also with a tiny bit of extra work, ability to add "Manage Account" button to your FxA settings page.

This requires WebChannels webext integration.

keianhzo:
Fixes #2267 This PR adds support for AC FxA WebChannels implementation replacing the current interceptor implementation (can be switched back through a hidden preference). There should be no difference with the previous interceptor implementation. The WebChannels implementation adds support for an after-signin sync page and also to the account management page (not yet implemented).

This sits on top of https://github.com/MozillaReality/FirefoxReality/pull/2466 and requires that PR to be landed.


app/src/common/shared/org/mozilla/vrbrowser/browser/engine/GeckoEngineSession.java
```diff
@@ -0,0 +1,112 @@
...
+    @Override
+    public void restoreState(@NotNull EngineSessionState engineSessionState) {
+
+    }
+
+    @NotNull
+    @Override
+    public EngineSessionState saveState() {
+        return null;
+    }
```
bluemarvin: This seems bad, return `null` from `@NotNull`

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/GeckoEngineSession.java
```diff
@@ -0,0 +1,112 @@
...
+    public GeckoSession getGeckoSession() {
+        return mSession.getGeckoSession();
+    }
+
+    @NotNull
+    @Override
+    public Settings getSettings() {
+        return null;
+    }
+
```
bluemarvin: This one too, returning `null` from `@NotNull`

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/GeckoEngineSession.java
keianhzo: We are not using it right now as this is just a bridge for using AC. I'd say to implement as we need as it would be quite time consuming to implement the whole bridge right now.

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/GeckoEngineSession.java
keianhzo: I've added support for Settings although we are not using it ATM. Our bridge is merely for handling the session state in the web extensions features.

keianhzo:
Closing as there is a new adapter based approach.

PR #2439, opened 2019-12-09
Glean build system update

### LINKED ISSUE #2388: Migrate Glean to the latest build system
Latest version of Glean uses a different build system based on a glean gradle plugin. We should keep up to date with the latest Glean version and migrate our build system to the new one.

We can safely follow what Fenix does on this regard.

**Additional context**
https://mozilla.github.io/glean/book/user/adding-glean-to-your-project.html

keianhzo:
Closes #2388 This PR migrates to the new Glean build system (as the previous one is tagged as deprecated).

Other than that nothing has changed, we still need to delete the generated Kotlin files and the markdown docs so they get generated again.

### DIFF
--- a/versions.gradle
+++ b/versions.gradle
@@ -50,6 +50,7 @@ versions.kotlin = "1.3.31"
 versions.kotlin_coroutines = "1.2.1"
 versions.snakeyaml = "1.24"
 versions.gson = "2.8.5"
+ext.versions = versions
 def deps = [:]
 
 def gecko_view = [:]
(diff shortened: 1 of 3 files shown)


versions.gradle
```diff
@@ -21,7 +21,7 @@
 * maven repositories.
 **/
 ext.deps = [:]
-def versions = [:]
+ext.versions = [:]
```
daoshengmu: I think we would like to put back `def versions = [:]` and add `ext.versions = versions` after Ln 52.

keianhzo: Fixed

PR #2478, opened 2019-12-16
Rename ping to avoid underscores

badboy:
Not ready for merge yet.
Requires an upgrade of android-components (for the updated sdk_generator script).
We're landing changes in the pipeline as well first.

There's no data loss. Data will still go to the same table for both `session_end` and `session-end` pings.
I'll come back here and put it out of Draft mode when we're ready to land it.

### DIFF
--- a/app/metrics.yaml
+++ b/app/metrics.yaml
@@ -16,7 +16,7 @@ distribution:
       - baseline
       - events
       - metrics
-      - session_end
+      - session-end
     bugs:
       - https://github.com/MozillaReality/FirefoxReality/issues/1420
     data_reviews:
@@ -30,7 +30,7 @@ url:
   domains:
     type: counter
     send_in_pings:
-      - session_end
+      - session-end
     description: >
       Counting how many domains are visited in a session.
     bugs:
@@ -44,7 +44,7 @@ url:
   visits:
     type: counter
     send_in_pings:
-      - session_end
+      - session-end
     description: >
       Counting how many URL links are visited in a session.
     bugs:
@@ -92,7 +92,7 @@ tabs:
     description: >
       Number of tabs opened during a session
     send_in_pings:
-      - session_end
+      - session-end
     labels:
       - context_menu
       - tabs_dialog
@@ -115,7 +115,7 @@ tabs:
     description: >
       Number of tabs activated during a session
     send_in_pings:
-      - session_end
+      - session-end
     bugs:
       - https://github.com/MozillaReality/FirefoxReality/issues/1609
     data_reviews:
@@ -229,7 +229,7 @@ control:
     description: >
       Counting how many general windows are opened in a session.
     send_in_pings:
-      - session_end
+      - session-end
     bugs:
       - https://github.com/MozillaReality/FirefoxReality/issues/2347
     data_reviews:
--- a/app/pings.yaml
+++ b/app/pings.yaml
@@ -4,7 +4,7 @@
 
 $schema: moz://mozilla.org/schemas/glean/pings/1-0-0
 
-session_end:
+session-end:
   description: >
     This ping is sent at the end of a session (when Firefox Reality switches to the background).
     We usually send search and UI control metrics at the end of a session.
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -10,7 +10,7 @@ This means you might have to go searching through the dependency tree to get a f
  - [baseline](#baseline)
  - [events](#events)
  - [metrics](#metrics)
- - [session_end](#session_end)
+ - [session-end](#session-end)
 
 
 ## baseline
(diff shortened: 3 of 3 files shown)


keianhzo:
@badboy If you refer to replacing the **sdk_generator** script for the **glean-gradle-plugin**, that's already been merged.

badboy:
Oh, that landed after I opened this PR.
That is nice for us, though right now the versioning seems incorrect. Published under `org.mozilla.telemetry:glean-gradle-plugin` it follows the Glean versioning, which is _not_ the same as that of a-c (we're working on distributing that plugin as an a-c component to mitigate this issue).

We don't yet have a plugin release based on that updated parser (whereas the sdk_generator script pulled from a-c _was_ updated).

I'll work on fixing that here (for now it should be fine), it's likely to not be done by tomorrow, so it has to wait for January.

daoshengmu:
If we are going to replace the ping name to `session-end`, please ask for a review from Glean team. I feel it will bring some penalty for querying data.

badboy:
@daoshengmu I am one of the members of the Glean team.
We ensured that data from both names will land in the same table. This is an update we made across all Glean users, but we will continue to support the old names for now (so that data from old clients still gets collected).

badboy:
Can't run the build locally right now. I hope CI will tell me.

This now includes:

* Upgrades to android-components, application-services and Glean to the latest (and matching/compatible) versions
* Uses the right version of the plugin
* Renames the ping

I haven't yet checked a-c/a-s changelogs for the large version bumps, we should probably check for (breaking) changes.

/cc @Dexterp37 (Glean team) for feedback.

Dexterp37:
> I haven't yet checked a-c/a-s changelogs for the large version bumps, we should probably check for (breaking) changes.
> 
> /cc @Dexterp37 (Glean team) for feedback.

This looks fine to me, but the Firefox Reality team should really check and see when this could fit their release schedule, given the big bump in the A-C version.

daoshengmu:
Using `versions.telemetry` for glean-plugin makes sense to me. However, our team need to consider the risk of bumping to the newest version of A-C.

build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
     addRepos(repositories)
     dependencies {
         classpath 'com.android.tools.build:gradle:3.4.2'
-        classpath "org.mozilla.telemetry:glean-gradle-plugin:$versions.android_components"
+        classpath "org.mozilla.telemetry:glean-gradle-plugin:$versions.telemetry"
```
Dexterp37: @badboy if we're considering doing a major version bump, maybe we could as well also make this repo use the Gradle plugin as exported by a-c?

/cc @mdboom

mdboom: Yes -- we definitely should do that.  There is still an [issue blocking that](https://github.com/mozilla-mobile/android-components/pull/5476).

My current plan is to just file PRs with all of our consumers once this is working, so I wouldn't worry about this just yet.

badboy:
Another problem (that's causing the CI failure): The glean_parser update is not actually included in the Glean version that's part of android-components (other folks are using the sdk_generator script still from a-c which we _did_ update)

That means we either:

1. Wait for an a-c release with the plugin and updated Glean (that's the long-term solution anyway)
2. Go back to using the sdk_generator script here for now
3. Special-case the plugin version here right now and use a different version from the Glean that is included.

What do you think, @Dexterp37 & @mdboom?
(That doesn't answer the question about the a-c upgrade for the Reality team, that needs to be evaluated independently, I'm going through the changelog right now)

badboy:
I went through the changelog files of [android-components](https://github.com/mozilla/application-services/blob/7583d12d16c46982f23b4d44aede9bfc4c1fd2e7/CHANGELOG.md) and [application-services](https://github.com/mozilla-mobile/android-components/blob/a53f8eaba324512188736fb1a9e45910580a973d/docs/changelog.md) for all versions between the old one and the new one used in this PR and looked at the breaking changes, then did a quick grep through the source code to see if anything was used.
It was not, so from that perspective the upgrade should be fine.

badboy:
We're just landing a Glean update to a-c that should solve the version mismatches (unfortunately a-c still does not deliver the plugin).

When that lands an upgrade to the next a-c version might be best.

@daoshengmu What's your take on the a-c upgrade in general? What's the release cycle for FirefoxReality?
(This is not critical to get in, we _do_ support the old format indefinitely, just offering my support here to land this)

badboy:
#2701 upgraded a-c and Glean.
I removed any version changes from this PR, it's now only the ping rename.

@daoshengmu or @keianhzo: would be good to land this now. :)

daoshengmu:
@Dexterp37 @badboy  I have a question for this ping name changing. Currently, our schema is called `org_mozilla_vrbrowser.session_end`. I wonder what will be the new schema name after this change? or it will still be the same?

Dexterp37:
> 
> 
> @Dexterp37 @badboy I have a question for this ping name changing. Currently, our schema is called `org_mozilla_vrbrowser.session_end`. I wonder what will be the new schema name after this change? or it will still be the same?

Nope, the table name will be the same.

badboy:
Whoops, sorry about that, not sure how I missed it. While I was at it I also regenerated the docs.

daoshengmu:
@Dexterp37 @mdboom Do you have your own opinion about this PR? If not, I am going to merge it because this Friday (02/28) is our release day.

bluemarvin:
> @Dexterp37 @mdboom Do you have your own opinion about this PR? If not, I am going to merge it because this Friday (02/28) is our release day.

Too late to make v9. Please hold for v10.

PR #2516, opened 2019-12-20
Add WebXR related cap flags to device delegates

kearwood:
WebXR changes and permission UI changes in GeckoView will break immersive mode in FxR, unless these flags are added for the VR devices.


kearwood:
This patch needs to land before bugzilla Bug 1603825 (Suppress the VR permission UI when no VR runtimes are detected), or WebVR immersive content will not detect the headset.

daoshengmu:
This change looks good to me. Let me take a try with my local build.

PR #2541, opened 2019-12-24
Making WaveVR app id suffix to be '.dev'.

### LINKED ISSUE #2514: Glean doesn't collect data from `org.mozilla.vrbrowser.wavevr`
## Configuration

Firefox Reality version: v.7
Firefox Reality build ID: 

Hardware: 
HTC Focus and Focus+

## Steps to Reproduce

1. Run WaveVR with the local build version.
2. Start Glean debugger with `adb shell am start -n org.mozilla.vrbrowser/mozilla.telemetry.glean.debug.GleanDebugActivity`. It will fail
3. If using `adb shell am start -n org.mozilla.vrbrowser.wavevr/mozilla.telemetry.glean.debug.GleanDebugActivity`. It works.

## Current Behavior

Getting failures to start Glean debugger, and there is no data be collected in `org_mozilla_vrbrowser` schema.

## Expected Behavior

The data from `org.mozilla.vrbrowser.wavevr` should still be able to uploaded to `org_mozilla_vrbrowser` schema.

## Context

`wavevr` is not an actual distribution channel, that is for separating the build for our partner whose distribution channel is `wavevrStore`. Therefore, we have two application id, `org.mozilla.vrbrowser` for deploying and `org.mozilla.vrbrowser.wavevr ` for developing.

```
```

</details>

daoshengmu:
Fixes #2514. We should not have a specific app id for WaveVR. The original purpose at  #1476 is in order to allow developers can deploy their local build into WaveVR devices. However, that will make our application id has an another branch. That will be hard to do data collection.

So, the idea is we should unifiy non `org.mozilla.vrbrowser` application to be `org.mozilla.vrbrowser.dev` includes `simultaneousDevProduction=true` that mean this build is deployed by developers or other unofficial distribution channels.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -8,6 +8,8 @@ apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
 apply plugin: "org.mozilla.telemetry.glean-gradle-plugin"
 
+def devSuffix = ".dev"
+
 def getGitHash = { ->
     def stdout = new ByteArrayOutputStream()
     exec {
@@ -26,7 +28,7 @@ def getCrashRestartDisabled = { ->
 
 def getDevApplicationIdSuffix = { ->
     if (gradle.hasProperty("userProperties.simultaneousDevProduction")) {
-        return gradle."userProperties.simultaneousDevProduction" == "true" ? ".dev" : ""
+        return gradle."userProperties.simultaneousDevProduction" == "true" ? devSuffix : ""
     }
     return ""
 }
@@ -44,6 +46,7 @@ ext.gleanDocsDirectory = "$rootDir/docs"
 
 android {
     compileSdkVersion build_versions.target_sdk
+
     defaultConfig {
         applicationId "org.mozilla.vrbrowser"
         minSdkVersion build_versions.min_sdk
@@ -193,7 +196,7 @@ android {
                     arguments "-DVR_SDK_LIB=wavevr-lib", "-DWAVEVR=ON"
                 }
             }
-            applicationIdSuffix ".wavevr"
+            applicationIdSuffix devSuffix
         }
 
         wavevrStore {


daoshengmu:
I have checked either Quest or Go uses `org.mozilla.vrbrowser` no matter FxR is downloaded from the store or run in local.

Only Focus Plus and Focus use `org.mozilla.vrbrowser.wavevr` as the application id when running in local, other ppl distribute our app will use this app id as well. If downloading from WaveVR store, the app id will be `org.mozilla.vrbrowser` though.

daoshengmu:
Another question is if we can separate the local build application id of WaveVR to be `org.mozilla.vrbrowser.dev` and the one distributes to WaveVR store to be `org.mozilla.vrbrowser`. Do we still need their`FLAVOR_platform` to be different? (`wavevr` vs. `wavevrStore`).

keianhzo:
The main purpose of this wavevr specific appId is that QA is able to test staging app updates as they mostly test staging builds and hardly ever test debug ones. Staging builds let them test the release builds of the app that include things like minification, different endpoints and assets.

Now we have:
- WaveVR Store: org.mozilla.vrbrowser (Store)
- WaveVR Staging: org.mozilla.vrbrowser.wavevr (QA)
- WaveVR Debug: org.mozilla.vrbrowser.wavevr.debug (Dev)

So I don't think we want to get rid of that appId. What's the problem of using `adb shell am start -n org.mozilla.vrbrowser.wavevr/mozilla.telemetry.glean.debug.GleanDebugActivity` to test Glean locally? This shouldn't affect Glean production analytics, right?

daoshengmu:
@keianhzo I didn't see we have `org.mozilla.vrbrowser.wavevr.debug` case even though the WaveVR Debug, it still uses `org.mozilla.vrbrowser.wavevr`.

The main reason I dug into this is because we have about 10% MAU from our old telemetry system tells us that they are using `org.mozilla.vrbrowser.wavevr`. 10% is not a small amount though, I am afraid we have some users get FxR through unofficial channels  instead of WaveVRStore.

keianhzo:
@daoshengmu You can enable the `.dev` suffix in any build setting `simultaneousDevProduction=true` in `user.properties`. We use that so we can have both dev and release builds in all the devices as otherwise it would clash with the store build in case you have that installed.

The `wavevrStore` flavor uses the default `org.mozilla.vrbrowser` and the `wavevr` flavor uses `org.mozilla.vrbrowser.wavevr`. As the app comes preinstalled in the Vice Focus, if we didn't use the `.wavevr` suffix flavor we wouldn't be able to test staging builds there which are the ones that QA uses.

I understand your concerns but we should talk about this in a team meeting because this may have consequences for the QA process. I'll bring it up in the next one.

daoshengmu:
> I don't understand what this PR is trying to accomplish. We have to have the `.wavevr` for local development. Calling it `.dev` instead won't work because that is already used for a different purpose. It can't be helped if people are side loading the `.wavevr` build flavor. I don't see any way to have only one build flavor for WaveVR.

I am trying to avoid using `.wavevr` for local development. Technically, we should unify non-`org.mozilla.vrbrowser` local development to be `.dev` only. This will help us minimize our app id as few as possible.

Adding a `org.mozilla.vrbrowser.wavevr` for telemetry data source doesn't really matter, it can be easy to achieve. However,  I am afraid it might happen from other new platforms need us to do this similar things as WaveVR. That will make us have to provide a couple of app ids. It will be difficult to manage.

daoshengmu:
Let’s leave it as is for now.

PR #2575, opened 2020-01-03
Updated Immersive mode spinner

keianhzo:
Immersive mode spinner update


bluemarvin:
I'm not sure I like the rocking motion of the fox.

app/src/main/cpp/LoadingAnimation.cpp
```diff
@@ -39,33 +42,52 @@ struct LoadingAnimation::State {
...
+     m.globe_rotation -= max;
   }
 
   vrb::Matrix transform = vrb::Matrix::Identity();
   const float scale = 0.05f;
   transform.ScaleInPlace(vrb::Vector(scale, scale, scale));
-  transform.PreMultiplyInPlace(vrb::Matrix::Rotation(vrb::Vector(0.0f, 1.0f, 0.0f), m.rotation));
   transform.PreMultiplyInPlace(vrb::Matrix::Position(vrb::Vector(0.0f, 0.0f, -1.5f)));
-  m.spinner->SetTransform(transform);
+  m.fox->SetTransform(transform);
```
bluemarvin: Since the fox isn't moving you should only need to set this transform once when the vrb::Transform node is created.

bluemarvin:
This looks much better to me. I re-enabled multisampling for immersive mode and the image is much smoother. We probably just need to switch to immersive mode FBOs once we start receiving immersive frames. You can see what it looks like by change this line:

https://github.com/MozillaReality/FirefoxReality/blob/master/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp#L98

Change `attributes.samples = 0;` to `attributes.samples = 4;`

app/src/main/cpp/LoadingAnimation.cpp
```diff
@@ -39,33 +42,52 @@ struct LoadingAnimation::State {
...
+
+  vrb::Matrix transform1 = vrb::Matrix::Identity();
```
bluemarvin: Please use a different variable name than `transform1`

bluemarvin:
Once we land this, I think I can fix the multisampling.

app/src/main/cpp/LoadingAnimation.cpp
keianhzo: Fixed

app/src/main/cpp/LoadingAnimation.cpp
keianhzo: Fixed

bluemarvin:
We need to wait for @jvonitter to sign off before landing.

keianhzo:
Pushed to v9

MortimerGoro:
I moved this to v10 to be at the same time as the WebXR release

MortimerGoro:
There is a new animation, closing this

PR #2586, opened 2020-01-03
FxA sync fixes

### LINKED ISSUE #2409: The History and Bookmarks sync switches are not updated immediately after a sync if they were changed on another device
**[Affected Versions]:**
- Firefox Reality version 7 RC5 (0706b656)

**[Affected Platforms]:**

**[Prerequisites]:**
- Make sure to be signed into the same FxA account on the VR device as well as another non-VR device and have them be synced.

**[Steps to reproduce]:**
1. Open the “Manage Account” card from Settings and observe the History and Bookmarks sync settings.
2. On the non-VR device, change the History and Bookmarks sync settings to other states.
3. Perform a Sync on the non-VR device.
4. Perform a Sync on the VR device and observe the History and Bookmarks sync switches.

**[Expected results]:**
- The sync switches are changed to the latest set states.

**[Actual results]:**
- The sync switches are not updated to the new states.

**[Notes]:**
- Closing and reopening the “Manage Account” card makes the switches display the proper states.
- Attached is a recording of the [issue](https://drive.google.com/file/d/1J9kcrPwba3Z1A0MS4PPwsAUVhhX1QjLE/view?usp=sharing).

### LINKED ISSUE #2411: The History and Bookmarks sync settings maintain the last states set by the device across accounts and syncs
**[Affected Versions]:**
- Firefox Reality version 7 RC5 (0706b656)

**[Affected Platforms]:**
- Oculus Go - Android 7.1.2

**[Prerequisites]:**
- Make sure to be signed into an FxA account.

**[Steps to reproduce]:**
1. Open the “Manage Account” card from Settings.
2. Change the History and Bookmarks sync settings and sign out.
3. Sign in with another account for which the History and Bookmarks sync settings are different.
4. Observe the sync settings.

**[Expected results]:**
- The sync settings are the ones corresponding to the new account.

**[Actual results]:**
- The sync settings are the ones corresponding to the last set from the VR device.

**[Notes]:**
- This issue is also reproducible when only using one account when syncing the settings from another device and logging out and then logging in again.
- The issue is also reproducible if a sync is performed before singing out.
- Due to the Sync automatically performed at sign in, the settings for the other devices connected to the account used are overridden as well.
- This issue is also reproducible on Firefox Preview.
- Attached is a recording of the [issue](https://drive.google.com/open?id=1nPlbkyXO8FGT01AuEpAiaipCAxiGP_hm).

keianhzo:
Fixes #2411 Fixes #2409 We were syncing only when the user explicitly did it or after exiting the accounts dialog but this has some side effects like the issue one. I've moved this to the same way as Fenix does it, we might encounter past issues like the sync status being reverted if changing it and syncing right away but this is more similar to how Fenix works and users shouldn't be syncing that much so I guess it's better to have real time sync.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/Accounts.kt
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/Accounts.kt
@@ -25,8 +25,8 @@ import mozilla.components.service.fxa.sync.SyncStatusObserver
 import mozilla.components.service.fxa.sync.getLastSynced
 import org.mozilla.vrbrowser.R
 import org.mozilla.vrbrowser.VRBrowserApplication
-import org.mozilla.vrbrowser.utils.BitmapCache
 import org.mozilla.vrbrowser.telemetry.GleanMetricsService
+import org.mozilla.vrbrowser.utils.BitmapCache
 import org.mozilla.vrbrowser.utils.SystemUtils
 import org.mozilla.vrbrowser.utils.ViewUtils
 import java.net.URL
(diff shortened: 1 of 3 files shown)


bluemarvin:
@MortimerGoro I'm just going to land this. I guess any blocking issues can be addressed in a follow up.

PR #2604, opened 2020-01-08
Fixes #1946 Awesome bar copy url

### LINKED ISSUE #1946: I'd like to copy the URL from the awesome bar to the URL bar
When you start typing and the awesome bar starts showing results sometimes you need to copy the URL to the URL bar to edit it and remove/add something. At the moment we only offer the option of clicking on it to go to that URL.

I'd like to have a button to copy that URL to the URL bar to be able to edit it without navigating to that URL. Ideally we could add a button to the right of the awesome bar item that just copies the URL to the URL bar, closes the awesome bar and places the cursor at the end of the URL bar text so you can start editing it.

We could also add other options like, sharing the URL as a tab through FxA "Send tab service" or just copy the URL to past it somewhere else.

**Additional context**
photo5829939885937636135

keianhzo:
Fixes #1946 Adds support for action menu in the awesome bar currently only supporting copying the row's URL. A border has been added to the actions menu (approved) as the background color it's the same as the awesome bar's background color.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/SuggestionsWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/SuggestionsWidget.java
@@ -1,7 +1,12 @@
 package org.mozilla.vrbrowser.ui.widgets;
 
+import android.content.ClipData;
+import android.content.ClipboardManager;
 import android.content.Context;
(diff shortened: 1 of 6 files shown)


bluemarvin:
Please address the taskcluster failure.

keianhzo:
Should be ok now

PR #2606, opened 2020-01-09
Fix skybox reverting to non-void after exiting immersive theatre mode.

### LINKED ISSUE #2605: After changing skybox to void, skybox reverts on exiting theatre mode.
## Configuration

Firefox Reality version: 8
Firefox Reality build ID: from master 34c12f5bd4001e8985ac9b76a16d100ff5267d4a

Hardware: All supported

## Steps to Reproduce
1. Select a skybox other than void.
2. Switch skybox to void.
3. Enter theatre mode and then choose an immersive projection
4. Exit immersive theatre mode.

## Current Behavior
Previous non-void skybox is visible.

## Expected Behavior
"void" remains selected, i.e. no skybox is visible.

## Possible Solution
Currently, in BrowserWorld, if changing to "void" and a skybox has previously been loaded, the previous skybox is just hidden, rather than removed from the scene graph, making other uses of show/hide skybox produce undesired behaviour. Better to remove and delete the skybox. Fixed by #2606.

philip-lamb:
Fixes #2605

### DIFF
--- a/app/src/main/cpp/BrowserWorld.cpp
+++ b/app/src/main/cpp/BrowserWorld.cpp
@@ -1480,13 +1480,21 @@ BrowserWorld::CreateSkyBox(const std::string& aBasePath, const std::string& aExt
   ASSERT_ON_RENDER_THREAD();
   vrb::PausePerformanceMonitor pauseMonitor(*m.monitor);
   const bool empty = aBasePath == "cubemap/void";
+  if (empty) {
+    if (m.skybox) {
+      VRLayerCubePtr layer = m.skybox->GetLayer();
+      if (layer) {
+        m.device->DeleteLayer(layer);
+      }
+      m.skybox->GetRoot()->RemoveFromParents();
+      m.skybox = nullptr;
+    }
+    return;
+  }
   const std::string extension = aExtension.empty() ? ".ktx" : aExtension;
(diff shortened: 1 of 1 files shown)


app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -1480,13 +1480,17 @@ BrowserWorld::CreateSkyBox(const std::string& aBasePath, const std::string& aExt
   ASSERT_ON_RENDER_THREAD();
   vrb::PausePerformanceMonitor pauseMonitor(*m.monitor);
   const bool empty = aBasePath == "cubemap/void";
+  if (empty) {
+    if (m.skybox) {
+      m.rootOpaqueParent->RemoveNode(*m.skybox->GetRoot());
+      m.skybox = nullptr;
```
bluemarvin: I believe this will leak the cube layer on devices that support async layers. I think you need to do:
```
  VRLayerCubePtr layer = m.skybox->GetLayer();
  if (layer) {
    m.device->DeleteLayer(layer);
  }
```
@MortimerGoro is that correct?

Also you can just call:
`m.skybox->RemoveFromParents();`

MortimerGoro: yes, that's correct

PR #2619, opened 2020-01-10
Bump GV to 74.0.20200107215758

bluemarvin:


bluemarvin:
@MortimerGoro This is the same version used in v7.1 and immersive mode worked for me when I tested it.

MortimerGoro:
Tested and WebVR works in this one

PR #2633, opened 2020-01-14
Support for shaped clipped event handler

keianhzo:
Adds a delegate to handle sticky shape clipped events for UI buttons. 

This is a continuation of the work done for the honeycomb buttons input clipping based on the background shape.

You can appreciate the clipping in the URL bar buttons and the back buttons in the panels as they extend from UIBUtton. We can easily extend this to the rest of the UI buttons using the same delegate so all the input is background shape clipped.

Related to: https://github.com/MozillaReality/FirefoxReality/pull/2601

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/views/ClippedEventDelegate.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/views/ClippedEventDelegate.java
@@ -1,75 +1,82 @@
 package org.mozilla.vrbrowser.ui.views;
 
-import android.graphics.Path;
-import android.graphics.RectF;
 import android.graphics.Region;
 import android.view.MotionEvent;
 import android.view.View;
 import android.view.View.OnClickListener;
 
-import androidx.annotation.DrawableRes;
 import androidx.annotation.NonNull;
 
 import org.mozilla.vrbrowser.utils.ViewUtils;
 
-import java.util.Deque;
+public abstract class ClippedEventDelegate implements View.OnHoverListener, View.OnTouchListener {
 
-public class ClippedEventDelegate implements View.OnHoverListener, View.OnTouchListener {
-
-    private View mView;
-    private Region mRegion;
     private boolean mHovered;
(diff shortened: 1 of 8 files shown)


bluemarvin:
@keianhzo Looks like this needs to be rebased.

MortimerGoro:
@keianhzo master has also some instances of this behaviour but it got worse with this PR. Example:

- Click on the home button
- Move the pointer to back button
- Release the pointer in back button
- The home action is triggered in this branch but not in master

MortimerGoro:
Ideally we should cancel all the clicks if they are released outside the area (unless the button is small and needs some extra margin to avoid failing touches)

bluemarvin:
> Ideally we should cancel all the clicks if they are released outside the area (unless the button is small and needs some extra margin to avoid failing touches)

But isn't this how we wanted sticky buttons to work? When you pressed in a button it would capture focus until it was released?

MortimerGoro:
> But isn't this how we wanted sticky buttons to work? When you pressed in a button it would capture focus until it was released?

The problem we had is that after pressing if the pointer was moved the touch was cancelled and the click missed. For me the correct behavior is capture the focus until it's released, but if released outside of the button shape it should cancel the click. You can try the behavior in Oculus Launcher tray menu, it works that way.

bluemarvin:
Okay, to be clear we still want the keyboard keys to work the same, where even if you release outside of the key, it is still pressed. But for other bigger buttons, we don't want the button pressed if the pointer is outside the button?

keianhzo:
@MortimerGoro I've updated the PR and added support for cancelling the events when released outside the button bounds.

MortimerGoro:
@keianhzo this one needs to be rebased too, conflicts with the changes for Android 8: // Android >8 doesn't perform a click when long clicking in ImageViews even if long click is disabled

keianhzo:
I don't think this is worth landing. There seems to be a few side effect and the benefit doesn't justify the risk of regressions.

PR #2636, opened 2020-01-14
Fixes #2568 Fix issue with shift and multi tapping

### LINKED ISSUE #2568: [Keyboard] The keyboard doesn't change to lowercase if the up arrow key it's quickly clicked three times or more
**[Affected Versions]:**
- Firefox Reality version 8-33611921 (67403bf2)

**[Affected Platforms]:**

**[Prerequisites]:**

**[Steps to reproduce]:**                                        
1. Focus the Awesomebar.
2. Quickly click the up arrow key from the keyboard three times.
3. Observe the behavior.

**[Expected results]:**
- The keyboard is switched to lowercase.

**[Actual results]:**
- The keyboard remains in uppercase.

**[Notes]:**
- This issue is reproducible on all keyboard languages.
- Attached is a screen recording of the issue: [link](https://drive.google.com/open?id=1jlXyRaAcWM56awrRK0AvSk_ae9Q3ZOHL).

keianhzo:
Fixes #2568 Fix issue with shift and multi tapping

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/views/CustomKeyboardView.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/views/CustomKeyboardView.java
@@ -1507,7 +1507,10 @@ public class CustomKeyboardView extends View implements View.OnClickListener {
             }
 
         } else if (key.codes[0] == CustomKeyboard.KEYCODE_SHIFT) {
-            mInMultiTap = true;
+            if (eventTime < mLastTapTime + MULTITAP_INTERVAL
+                    && keyIndex == mLastSentIndex) {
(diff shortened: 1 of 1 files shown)


bluemarvin:
This patch doesn't work the way I expect. The time out seems to be too long. The double click gets the shift key into caps lock mode. But after that we should just reset since I don't think we need to support a triple click?

keianhzo:
@bluemarvin I've rebased and reduced the multitap timeout. Also I have fixed the keyboard refresh, I've seen that we weren't updating both shift keys when multitapping. I think it's ok if multitapping several times keeps it enabled just in case you click more than two times by mistake.

PR #2639, opened 2020-01-14
Fixes #1945 Restore library panels

### LINKED ISSUE #1945: Save/restore library panels when the app is reopened
At the moment FxR is saving the layout and restores it after the app is reopened but in only restores browser windows, not library panels.

I'd like to be able to restore also library panels so my exact layout is restored upon app reopen. See the screenshot.

Also would be great if I could save/restore specific layouts that I might use for different tasks. For example a layout that I use for reading different newspapers, a layout for developing or a layout for general browsing or watching movies.

**Additional context**
org mozilla vrbrowser dev-20191008-234253

keianhzo:
Fixes #1945 Restore library panels. I really don't like to have a flag to know if the user has initiated the interaction so if there is a better alternative I'd be happy to change this. Without the flag the panels would be closed during the restore when about:blank and home are loaded as we always close them when there is a navigation event.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/Windows.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/Windows.java
@@ -27,6 +27,7 @@ import org.mozilla.vrbrowser.ui.widgets.dialogs.UIDialog;
 import org.mozilla.vrbrowser.utils.BitmapCache;
 import org.mozilla.vrbrowser.utils.ConnectivityReceiver;
 import org.mozilla.vrbrowser.utils.SystemUtils;
+import org.mozilla.vrbrowser.utils.UrlUtils;
 
 import java.io.File;
 import java.io.FileReader;
@@ -65,6 +66,7 @@ public class Windows implements TrayListener, TopBarWidget.Delegate, TitleBarWid
         int textureHeight;
         float worldWidth;
         int tabIndex = -1;
+        PanelType panelType;
 
         public void load(WindowWidget aWindow, WindowsState aState, int aTabIndex) {
             WidgetPlacement widgetPlacement;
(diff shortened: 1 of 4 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/Windows.java
```diff
@@ -275,6 +289,14 @@ private WindowWidget addRestoredWindow(@NonNull WindowState aState, @NonNull Ses
         newWindow.getPlacement().worldWidth = aState.worldWidth;
         newWindow.setRestored(true);
         placeWindow(newWindow, aState.placement);
+        switch (aState.panelType) {
```
bluemarvin: I'm getting a crash here:
https://crash-stats.mozilla.com/report/index/ded91d9d-9f44-46ba-bc9c-18fca0200123
https://crash-stats.mozilla.com/report/index/e8e5b8d8-e61a-4b3a-87e9-9770f0200123
STR:
1) Run FxR without this patch applied and quit so windows are saved.
2) Apply patch and re-run FxR
FxR will crash on start up.

keianhzo: Good catch. I've added a check to see if the new panel information is present.

keianhzo:
This should be ready for review.

bluemarvin:
I agree this isn't pretty but I can't think of anything cleaner. @MortimerGoro last chance to comment before landing it.

MortimerGoro:
@bluemarvin @keianhzo would it be possible to use a URL based approach instead of adding the extra flags and enum? 

For example map `about://history` or `about://bookmarks` kind of URLs to the library panels

bluemarvin:
> @bluemarvin @keianhzo would it be possible to use a URL based approach instead of adding the extra flags and enum?
> 
> For example map `about://history` or `about://bookmarks` kind of URLs to the library panels

That could be workable. I'm not sure how we would make it work with the GV history system though. I like the concept, it isn't clear how we would implement it.

keianhzo:
@bluemarvin @MortimerGoro That a good idea, I've added support for it. It doesn't seem to mess up the history and it's a much cleaner solution for restoring.

PR #2702, opened 2020-01-23
update metrics.md file

bluemarvin:


daoshengmu:
These polite words are generated by Glean automatically. After applying PR of #2701, they are gone.

PR #2745, opened 2020-02-05
Support for dragging outside the widget bounds

keianhzo:
When you are press and drag outside the widget bounds the input stops. This adds support for dragging outside the widget bounds. I think it is a more natural behavior for the input.

I started this branch a while ago so I guess it's a good moment to share and check if this is something that we like.

### DIFF
--- a/app/src/main/cpp/BrowserWorld.cpp
+++ b/app/src/main/cpp/BrowserWorld.cpp
@@ -398,8 +398,8 @@ BrowserWorld::State::UpdateControllers(bool& aRelayoutWidgets) {
                             controller.lastButtonState & ControllerDelegate::BUTTON_TOUCHPAD;
 
     bool isResizing = resizingWidget && resizingWidget->IsResizingActive();
-    bool isDragging = pressed && wasPressed && controller.widget;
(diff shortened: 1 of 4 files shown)


.editorconfig
```diff
@@ -0,0 +1,26 @@
+[*]
```
MortimerGoro: If we still need this file, can we separate it to a different PR?

keianhzo:
@MortimerGoro updated

bluemarvin:
@keianhzo Looks like this conflicts with the two controller typing.

bluemarvin:
I got this crash when trying to resize the window https://crash-stats.mozilla.com/report/index/6e253a60-c37d-4cdf-8954-0d9940200403

bluemarvin:
Crash is here: https://github.com/MozillaReality/FirefoxReality/pull/2745/files#diff-96af3977b6728c8ea9f0abae3d641614R421
`widget` is `null`

bluemarvin:
I'm still seeing the crash.

keianhzo:
@bluemarvin I've pushed a fix for the resize issue

PR #2747, opened 2020-02-05
Basic instrumentation tests

keianhzo:
This adds basic support for instrumentation test for evaluation of future work. 

The currently implemented tests are:

- Open settings and check that they are closed with the back button
- Open settings and check that they are closed with the device back button
- Open settings and check that they are closed clicking on the world
- Check that no more than 3 windows can be opened
- Grab a screenshot. It's a one eye screenshot so this is more aimed at emulator tests.

We can extend this by using direct real world input instead of actual perform clicks, there is a WIP branch with work on that regard: `experimental/inverse_touch_test`

Tests can be run both on emulator or device.


daoshengmu:
Will it run through T-C? Just like what we did for junit tests.

PR #2772, opened 2020-02-10
Fix leaked JNI local reference in GeckoSurfaceTexture

MortimerGoro:
I hit a JNI reference overflow when working on exit from WebXR. This happens because exit is not working yet and the ExternalBlitter->cancelFrame() is called every frame, leading to a crash after some seconds.

The leak also happens in WebVR but in a slower pace

### DIFF
--- a/app/src/main/cpp/ExternalBlitter.cpp
+++ b/app/src/main/cpp/ExternalBlitter.cpp
@@ -160,6 +160,7 @@ ExternalBlitter::CancelFrame(const int32_t aSurfaceHandle) {
     surface = iter->second;
   } else {
     surface = GeckoSurfaceTexture::Create(aSurfaceHandle);
+    m.surfaceMap[aSurfaceHandle] = surface;
   }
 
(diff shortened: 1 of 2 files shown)


app/src/main/cpp/ExternalBlitter.cpp
```diff
@@ -160,6 +160,7 @@ ExternalBlitter::CancelFrame(const int32_t aSurfaceHandle) {
     surface = iter->second;
   } else {
     surface = GeckoSurfaceTexture::Create(aSurfaceHandle);
+    m.surfaceMap[aSurfaceHandle] = surface;
```
bluemarvin: What condition causes us to have a `aSurfaceHandle` value with out it being in the `m.surfaceMap`?

MortimerGoro: A new texture that hasn't been used before or StopPresenting() call which cleared the surfaces

PR #2781, opened 2020-02-12
Follow system language

keianhzo:
This adds support for a "Follow system language" placeholder that uses whatever the system locale is as long as it's supported by FxR, otherwise it falls back to en-US. 

Also fixes some issues with the correct language hot reload. Now we should be able to change the language either from the Settings or from the system settings and everything so be instantly localized in the new language.

Things to test:
- Change the language from the language setting: All the UI strings should be now localized in the new language
- Click home - Go to the system settings - Change the language - Go back to the app: All the UI strings should be now localized in the new language
- Reset the languages: The UI should show "Follow the system Language" and all the strings should be localized in the system language. In case of an unsupported language (ie. pt-PT), it should fall back to en-US
- Set a display language explicitly - Click home - Go to the system settings - Change the language - Go back to the app: The user chosen language should remain as the UI language.
- Check that when "Follow system language" is selected the voice search is performed in the correct language based on the system one. In case of an unsupported language (ie. pt-PT), it should fall back to en-US
- Test updating from v8. The app should retain the previous language configuration and show specific locales for all three language settings.
- All the sumo pages should be shown localized in the content language order. They were previously hardcoded to en-US.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
@@ -58,6 +58,7 @@ import org.mozilla.vrbrowser.search.SearchEngineWrapper;
 import org.mozilla.vrbrowser.telemetry.GleanMetricsService;
 import org.mozilla.vrbrowser.telemetry.TelemetryWrapper;
 import org.mozilla.vrbrowser.ui.OffscreenDisplay;
+import org.mozilla.vrbrowser.ui.adapters.Language;
 import org.mozilla.vrbrowser.ui.widgets.KeyboardWidget;
 import org.mozilla.vrbrowser.ui.widgets.NavigationBarWidget;
 import org.mozilla.vrbrowser.ui.widgets.RootWidget;
@@ -86,7 +87,6 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
-import java.util.Locale;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.Consumer;
@@ -210,7 +210,8 @@ public class VRBrowserActivity extends PlatformActivity implements WidgetManager
 
     @Override
     protected void attachBaseContext(Context base) {
-        super.attachBaseContext(LocaleUtils.setLocale(base));
+        Context newContext = LocaleUtils.init(base);
+        super.attachBaseContext(newContext);
     }
 
     @Override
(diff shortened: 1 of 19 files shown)


keianhzo:
@Softvision-GeluHaiduc You can start with the language testing after this lands.

bluemarvin:
Got this crash on first run:
```
E/GeckoCrashHandler: >>> REPORTING UNCAUGHT EXCEPTION FROM THREAD 1 ("main")
    java.lang.NullPointerException: Attempt to invoke interface method 'boolean java.util.List.isEmpty()' on a null object reference
        at org.mozilla.vrbrowser.utils.LocaleUtils.getPreferredLanguages(LocaleUtils.java:161)
        at org.mozilla.vrbrowser.utils.LocaleUtils.getPreferredLanguageTags(LocaleUtils.java:151)
        at org.mozilla.vrbrowser.VRBrowserActivity.onCreate(VRBrowserActivity.java:245)
        at android.app.Activity.performCreate(Activity.java:6724)
        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1119)
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2632)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2740)
        at android.app.ActivityThread.-wrap12(ActivityThread.java)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1477)
        at android.os.Handler.dispatchMessage(Handler.java:102)
        at android.os.Looper.loop(Looper.java:154)
        at android.app.ActivityThread.main(ActivityThread.java:6144)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)
```

bluemarvin:
I get the same crash after a clean install also.

bluemarvin:
I was testing on the Quest.

keianhzo:
Fixed.

bluemarvin:
@daoshengmu Said he would investigate the test failures.

daoshengmu:
It works with t-c now. :)

Softvision-GeluHaiduc:
I have verified these new behaviors using the latest RC, Firefox Reality version 9 RC1 (33f6c4d8), installed on the Oculus Go, Quest, HTC Vive Focus, Focus Plus and Pico Neo2 headsets.
I have observed some different behaviors in the website language list on the HTC devices when French or Italian was set as app language and the placeholder for the URL bar doesn’t change languages when the device language is changed on the Pico Neo2.
In tested all the scenarios outlined in the PR and also some extra ones.

keianhzo:
@Softvision-GeluHaiduc This two PRs should fix those issues:
- https://github.com/MozillaReality/FirefoxReality/pull/2797
- https://github.com/MozillaReality/FirefoxReality/pull/2798

PR #2797, opened 2020-02-18
System config changes fixes

keianhzo:
This fixes some issues that I've seen while testing v9 with language update from the system config. Fixes the correct update of the URL and Title bars.
Also fixes the correct update of app dialogs in case the system language is changed while a dialog is opened.


Softvision-GeluHaiduc:
I have verified these new behaviors using the latest RC, Firefox Reality version 9 RC2 (37ac4cbd), installed on the Oculus Go, Quest, HTC Vive Focus, Focus Plus and Pico Neo2 headsets. Everything seems to be working as expected.
However, on the Pico Neo2  the placeholder for the URL bar doesn’t change languages when the device language is changed to another language besides English. I will log a separate issue for this.

PR #2798, opened 2020-02-18
Update Bookmarks strings after a language change

keianhzo:
Now after a language change the bookmark titles should be updated correctly.


Softvision-GeluHaiduc:
I have verified these new behaviors using the latest RC, Firefox Reality version 9 RC2 (37ac4cbd), installed on the Oculus Go, Quest, HTC Vive Focus, Focus Plus and Pico Neo2 headsets.
Now, after a language change, the bookmark titles are updated correctly.

PR #2801, opened 2020-02-19
Adjusting zh-tw keyboard layout.

### LINKED ISSUE #2714: [Keyboard][zh-TW] The string for the Space key exits the keys boundaries
**[Affected Versions]:**
- Firefox Reality version 9-40290101 (854287c0)

**[Affected Platforms]:**

**[Prerequisites]:**

**[Steps to reproduce]:**
1. Focus the Awesomebar.
2. Click the Language select key (globe).
3. Select the third language from the top left from the sub-menu displayed.
4. Observe the string displayed in the Space key.

**[Expected results]:**
- The string is correctly displayed.

**[Actual results]:**
- The string exits the key's boundaries.

**[Notes]:**
- Attached a screen recording of the issue [here](https://drive.google.com/open?id=1Rjv26adFZlRx8dS0STjYABipOi1E27RS).

daoshengmu:
Fixes #2714.

This makes more sense based on iOS zh-TW keyboard layout.

However, I still can see we have couple of questions on `KeyboardWidget`. @keianhzo  Please take a look at this screen recording (https://drive.google.com/file/d/1m43TsivrTurHQnA19aKykkDbhLhyQ7gI/view?usp=sharing)

- I can't be easy to adjust its placement and resize the KeyboardLayout according to the parameters from `KeyboardInterface`. Although I spend an afternoon, I still feel the width of `keyboardContainer` or `keyboardLayout` is not what I expect. [0:00 - 0:23 sec in the video]
- Pinyin and Japanese keyboards didn't be placed at the the same position as others. We can see there is an offset between it and others. [0:23 - 0:41 sec in the video] I have checked their UI definitions are the same with `keyboard_qwerty.xml`, but they are still wrong. That's odd.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/keyboards/ChineseZhuyinKeyboard.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/keyboards/ChineseZhuyinKeyboard.java
@@ -44,7 +44,7 @@ public class ChineseZhuyinKeyboard extends BaseKeyboard {
     @Override
     public CustomKeyboard getAlphabeticKeyboard() {
         if (mKeyboard == null) {
-            mKeyboard = new CustomKeyboard(mContext.getApplicationContext(), R.xml.keyboard_zhuyin);
+            mKeyboard = new CustomKeyboard(mContext.getApplicationContext(), R.xml.keyboard_qwerty_zhuyin);
             loadDatabase();
         }
         return mKeyboard;
@@ -153,7 +153,7 @@ public class ChineseZhuyinKeyboard extends BaseKeyboard {
     @Override
     public String getSpaceKeyText(String aComposingText) {
         if (aComposingText == null || aComposingText.trim().isEmpty()) {
-            return StringUtils.getStringByLocale(mContext, R.string.settings_language_traditional_chinese, getLocale());
+            return "";
         } else {
             return mContext.getString(R.string.zhuyin_spacebar_selection);
         }
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/keyboards/ChineseZhuyinKeyboard.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/keyboards/ChineseZhuyinKeyboard.java
@@ -24,10 +24,17 @@ import java.util.regex.Pattern;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
+import jp.co.omronsoft.openwnn.ComposingText;
+import jp.co.omronsoft.openwnn.SymbolList;
+import jp.co.omronsoft.openwnn.WnnWord;
+
 public class ChineseZhuyinKeyboard extends BaseKeyboard {
     private static final String LOGTAG = SystemUtils.createLogtag(ChineseZhuyinKeyboard.class);
     private static final String nonZhuyinReg = "[^ㄅ-ㄩ˙ˊˇˋˉ]";
     private CustomKeyboard mKeyboard;
+    private CustomKeyboard mSymbolsKeyboard;
+    private SymbolList mSymbolsConverter;  // For Emoji characters.
+    private List<Words> mEmojiList = null;
     private DBWordHelper mWordDB;
     private DBPhraseHelper mPhraseDB;
     private HashMap<String, KeyMap> mKeymaps = new HashMap<>();
@@ -48,6 +55,17 @@ public class ChineseZhuyinKeyboard extends BaseKeyboard {
             loadDatabase();
         }
         return mKeyboard;
+    }
+
+    @Nullable
+    @Override
+    public CustomKeyboard getSymbolsKeyboard() {
+        if (mSymbolsKeyboard == null) {
+            mSymbolsKeyboard = new CustomKeyboard(mContext.getApplicationContext(), R.xml.keyboard_symbols_zhuyin);
+            // We use openwnn to provide us Emoji character although we are not using JPN keyboard.
+            mSymbolsConverter = new SymbolList(mContext, SymbolList.LANG_JA);
+        }
+        return mSymbolsKeyboard;
     }
 
     @Nullable
@@ -93,6 +111,31 @@ public class ChineseZhuyinKeyboard extends BaseKeyboard {
         return result;
     }
 
+    @Override
+    public CandidatesResult getEmojiCandidates(String aComposingText) {
+        if (mEmojiList == null) {
+            List<Words> words = new ArrayList<>();
+            ComposingText text = new ComposingText();
+            mSymbolsConverter.convert(text);
+
+            int candidates = mSymbolsConverter.predict(text, 0, -1);
+            if (candidates > 0) {
+                WnnWord word;
+                while ((word = mSymbolsConverter.getNextCandidate()) != null) {
+                    words.add(new Words(1, word.stroke, word.candidate));
+                }
+                mEmojiList = words;
+            }
+        }
+
+        CandidatesResult result = new CandidatesResult();
+        result.words = mEmojiList;
+        result.action = CandidatesResult.Action.SHOW_CANDIDATES;
+        result.composing = aComposingText;
+
+        return result;
+    }
+
     private String GetTransCode(String aText) {
         String code = aText;
         String transCode = "";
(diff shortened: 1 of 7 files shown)


keianhzo:
@daoshengmu the proposal at https://github.com/MozillaReality/FirefoxReality/issues/2714#issuecomment-587028871 was not a final design I was just trying to understand if it was a valid layout based on what I see in Android Zhuyin, we still probably need to go through Design review for this and request a new redesign with a larger space bar and maybe that also keeps the same height as the rest of the keyboards as this one (and also my proposal are 5 rows instead of 4).

Regarding the KeyboardLayout, you may need to override the `getAlphabeticKeyboardWidth` method from `BaseKeyboard` to provide a new width. I've this already sorted out in my branch, which is basically the same as this one with small layout differences.

What you see in the Pinyin and Japanese keyboards is because those have the clause composition bar on top where the composed characters are displayed, that's why thy are placed a bit further down than the others.

@emstanley who should be making keyboard layout related decisions? I think in the past Janice run it through localization contributors?

daoshengmu:
> What you see in the Pinyin and Japanese keyboards is because those have the clause composition bar on top where the composed characters are displayed, that's why thy are placed a bit further down than the others.

It makes sense now :)

I have done the adjustment for zh-tw keyboard, and It looks great. I aware zh-tw IME would be the only one keyboard which uses 5 rows for now. Therefore, I need to do some customize work for it. For examples, its symbol keyboard is still 4 rows. It makes me need to change its layout when users are doing mode changing. Anyway, this PR is ready to review. @keianhzo

keianhzo:
@daoshengmu I guess we should wait until @emstanley confirms who's the right person to review this new design and we get the approval?

daoshengmu:
> Maybe we should fix the alignment of the second/third rows and also the enter button should be a bit bigger to be aligned with the backspace button.
> 
 the second/third rows are not aligned is intentional. iOS keyboard is like that, I haven't checked it on Android, but I guess so.
> 
> I'm also not very convinced about changing to a 4 rows layout when switching to the symbols keyboard. Android Zhuyin keeps the keyboard height when switching to the symbols keyboard but only shows 4 rows, switching the keyboard height feels weird, we don't do that for any other keyboard. Let see what @thenadj says after reviewing.

iOS also keeps the same height of keyboard layout, but when switching to the symbol keyboard that only has 4 rows, so it will make the keys size be bigger.

daoshengmu:
I have fixed the alignment problem with the backspace button. I have a feeling we should consider to make the symbol keyboard of Zh-TW keyboard to be 5 rows or we need to adjust key height   and keyboard height globally.

cjyjae35:
[BUG-40 Zhuyin Keyboard](https://trello.com/c/fFVzVV7y/521-bug-40-zhuyin-keyboard)

cjyjae35:
[BUG-40 Zhuyin Keyboard](https://trello.com/c/fFVzVV7y/521-bug-40-zhuyin-keyboard)

thenadj:
@daoshengmu Please take a look at BUG-40 for updates to the keyboard. 
[BUG-40 Zhuyin Keyboard_001.pdf](https://github.com/MozillaReality/FirefoxReality/files/4264598/BUG-40.Zhuyin.Keyboard_001.pdf)

bluemarvin:
Why are we arbitrarily changing the numeric key layout on *all* keyboard in this design?

thenadj:
@bluemarvin In April of 2019 we got the feedback that we should change the numeric key layout from the phone layout to the extended layout. At that point, we updated it in all specs. But looking at the builds now, all keyboards still have the telephone layout.

daoshengmu:
Using the extended layout for the numeric key layout makes sense to me. However, based on the new design, it still wouldn't solve our original problem that is caused by the space bar is not long enough to show the full text of `中文 ( 正體 / 注音)` as the captured video below.
https://drive.google.com/file/d/1Dj6KhPSCI4J6pSnIWkU3Ih08dm96KqCL/view?usp=sharing

thenadj:
@daoshengmu We recommend not to show the name on the spacebar of the Chinese keyboard. The user can see the language the chose from the language picker menu.

daoshengmu:
@thenadj I am ok to not show the input name on the spacebar for Traditional Chinese keyboard. How about the Simplified Chinese keyboard (Pinyin)? Does it need to show the name?

Moreover, I saw the design document mention the spacebar will show `Space` in different languages. But, we didn't have it now, we just make it to be blank in spacebar. Do we need to add it back to all the keyboards?

thenadj:
@daoshengmu Show it if there is space in the pinyin keyboard and keep it blank.

daoshengmu:
@keianhzo It is available to be reviewed now :).

daoshengmu:
Screen Shot 2020-03-09 at 10 25 16 PM

@cjyjae35  @thenadj One thing I would like to discuss is the character `ㄛ` and `ㄨ`。In the spelling system of Zhuyin, 'ㄨ' is always come before `ㄛ`. But, in this new design, 'ㄨ' is put at the below of 'ㄛ'. It is not a normal way that I experienced before. It is not too bad, users might be able to endure it. I just wanna share my thought to see if we still have space to enhance it in the layout.

bluemarvin:
> Screen Shot 2020-03-09 at 10 25 16 PM
> 
> @cjyjae35 @thenadj One thing I would like to discuss is the character `ㄛ` and `ㄨ`。In the spelling system of Zhuyin, 'ㄨ' is always come before `ㄛ`. But, in this new design, 'ㄨ' is put at the below of 'ㄛ'. It is not an usual way that I experienced before. It is not too bad, users might be able to endure it. I just wanna share my thought to see if we still have space to enhance it in the layout.

@daoshengmu Is the solution to just swap the two characters? If so, please just do it. I would like to get this landed.

daoshengmu:
> > @cjyjae35 @thenadj One thing I would like to discuss is the character `ㄛ` and `ㄨ`。In the spelling system of Zhuyin, 'ㄨ' is always come before `ㄛ`. But, in this new design, 'ㄨ' is put at the below of 'ㄛ'. It is not an usual way that I experienced before. It is not too bad, users might be able to endure it. I just wanna share my thought to see if we still have space to enhance it in the layout.
> 
> @daoshengmu Is the solution to just swap the two characters? If so, please just do it. I would like to get this landed.

@bluemarvin Things would not be so simple. Every character has their order in the keyboard by columns. If we wanna land it soon, we can open a follow up issue to address it, this problem is minor, honestly.

bluemarvin:
@daoshengmu Let get it right, we have time. Please propose a solution.

thenadj:
I agree, please get the order of the characters corrected. We are not native Chinese speakers and trust your feedback.

daoshengmu:
org mozilla vrbrowser-20200312-124744
org mozilla vrbrowser-20200312-124756

I think it looks good enough to me now. I took a small number of space from `Enter` button and give a margin at the second row.

bluemarvin:
If it is important to have the keys in a grid layout, would it make sense to have the enter key be the same width as delete and space keys?

daoshengmu:
> If it is important to have the keys in a grid layout, would it make sense to have the enter key be the same width as delete and space keys?

I don't have a strong opinion on this. One advantage of making Enter key be board might be the Enter key size would look like similar with other keyboards and if we still need to show `Enter` text on the key, the width can't be to be too thin.

bluemarvin:
So, maybe instead of showing `Enter` maybe we should show the enter symbol. This is what I see done on other mobile keyboards.

image

daoshengmu:
Showing symbols to present `Enter` and `Space` looks ok to me. Keys in grid layout should be unnecessary. I feel these functional keys should have different size from other keys.

keianhzo:
@daoshengmu Is this ready to land or we are going to change the enter key?

daoshengmu:
> @daoshengmu Is this ready to land or we are going to change the enter key?

My preference is landing this PR first and address the Enter key concern at another issue. Then, discuss it with our design team.

cjyjae35:
> > @daoshengmu Is this ready to land or we are going to change the enter key?
> 
> My preference is landing this PR first and address the Enter key concern at another issue. Then, discuss it with our design team.

`Enter`should be displayed on the key if the width is enough. If not, please use the symbols instead.

bluemarvin:
@daoshengmu Do you have enough information to finish this up? Would like to include in v10.

daoshengmu:
> @daoshengmu Do you have enough information to finish this up? Would like to include in v10.

@bluemarvin I have checked `Enter` key on all keyboards, all of them have enough width to show the text. So, I assume at this moment we still don't need to do any change for symbols.

daoshengmu:
> @daoshengmu Is this ready to land or we are going to change the enter key?

It is ready to land. Please help review once you're available.

daoshengmu:
> @daoshengmu Maybe we should also add the emoji symbols to the Simplified Chinese kb as a follow up? I know it was popular in JP but if it's popular in TW maybe it is popular too in Mainland China?

Yep. We will work on this in #2974.

PR #2802, opened 2020-02-19
Remove the back handler for non focused views

keianhzo:
Remove the back handler for non focused views.

STRs:
- Open two windows
- Open History/Bookmarks in one of them
- Focus the other window
- Click on the back button

Expected result: We go back in history in the currently focused window

Actual result: The History/Bookmark panel is closed in the non focused window.


Softvision-GeluHaiduc:
I have verified this issue with the latest Firefox Reality version 9 RC3 (18aba8be) installed on the Oculus Go, Quest, HTC Vive Focus, Focus Plus and Pico Neo2 headsets.
In the case of the Oculus Quest and Pico Neo2 the issue is still reproducible intermittently.
On the other devices tested, Oculus Go, HTC Vive Focus and Focus Plus, the issue seems to be fixed.

PR #2820, opened 2020-02-20
Update PlatformActivity.java

J-Tech-Foundation:
will need try catch IOException statements, but just for easier debugging this should be added.


bluemarvin:
The global exception handler should already catch this. I don't understand why this change is necessary.

PR #2825, opened 2020-02-21
coil support

anselm:
This is support for coil. It adds coil to the list of plugins and then also I've pasted the latest rev of coil itself into the web plugins folder.


bluemarvin:
What are the performance implications for adding this web extension to every page load?

app/src/main/assets/web_extensions/coil/static/goSettings.js
```diff
@@ -0,0 +1,2 @@
+const url = 'https:/coil.com/settings'
```
jonathansampson: Shouldn't that be `http://` (note, two forward slashes)?

anselm: wow good catch - i will forward to that team

anselm: https://github.com/coilhq/web-monetization-projects/blob/master/packages/coil-extension/static/goSettings.js was the original location - am advising them

PR #2838, opened 2020-02-21
Proxify Hamburger Menu. Fixes #2833

### LINKED ISSUE #2833: [Oculus] The hover effect of the inactive window is displayed over the hamburger menu
**[Affected Versions]:**
- Firefox Reality version 9 RC2 (37ac4cbd)

**[Affected Platforms]:**
- Oculus Go - Android 7.1.2

**[Prerequisites]:**

**[Steps to reproduce]:**
1. Click the new window button from the menu tray.
2. Focus the window to the left and click the Hamburger menu.
3. Hover the other window and observe the behaviour.

**[Expected results]:**
- The hover effect of the window is displayed behind the Hamburger menu.

**[Actual results]:**
- The hover effect of the window is displayed over the Hamburger menu.

**[Notes]:**
- The issue is also reproducible in production, FxR V8.
- The issue is not reproducible on the HTC or Pico devices.
- Attached a screen recording of the issue [here](https://drive.google.com/open?id=1gtcB80sH1j255DHQDxIVb6E4JePG-rBu).

bluemarvin:


bluemarvin:
> Should we proxify only when needed? we are doing that on the keyboard (only when more than one window is visible)

I could do that, I just figured it wasn't too big an issue since the menu is visible often or for very long. So I figured just proxifying while visible would be sufficient. I can look into check for multiple windows.

PR #2867, opened 2020-02-26
Use the default appservices megazord, not the deprecated fenix megazord.

rfk:
The custom "fenix" megazord contains all the same code as appears in
the default appservices megazord, so consumers who are using it are
adding complexity to their build process for no gain. The appservices
team intend to stop building the fenix megazord and instead encourage
consumers to use the default full megazord. See additional discussion
in https://github.com/mozilla/application-services/issues/1466.

This PR removes the megazord substitution rules, meaning FirefoxReality
will get the full megazord by default. Fenix is making a corresponding
change over in https://github.com/mozilla-mobile/fenix/pull/8742.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -485,18 +485,6 @@ dependencies {
     implementation deps.android_components.glean
     implementation deps.app_services.rustlog
 
-    // For production builds, the native code for all `org.mozilla.appservices`
-    // dependencies gets compiled together into a single "megazord" build, and
-    // different megazords are published for different subsets of features. Ref
-    // https://mozilla.github.io/application-services/docs/applications/consuming-megazord-libraries.html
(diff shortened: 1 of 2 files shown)


rfk:
I'm not really set up to QA whether this change breaks anything, but an currently running `./gradlew :test` locally as a basic smoketest. I'm happy to take advise on futher testing that should be done.

/cc @grigoryk

keianhzo:
Seems to work for me too. Merging and tagging QA for further testing.

Softvision-GeluHaiduc:
I’ve performed exploratory testing on the FxA, Bookmarks and History features using the latest master, Firefox Reality version 10-40720037 (dbb8ed3a), installed on the Oculus Quest headset. 
I haven’t encountered any new issues during testing.

PR #2879, opened 2020-02-26
Ensure WindowWidget runnables are called in onFirstComposite

bluemarvin:


bluemarvin:
This doesn't seem to address the issue.

PR #2888, opened 2020-02-27
Adds support for the home long press to recenter event

keianhzo:
Adds support for the home long press to recenter event

### DIFF
--- a/app/src/picovr/cpp/DeviceDelegatePicoVR.cpp
+++ b/app/src/picovr/cpp/DeviceDelegatePicoVR.cpp
@@ -4,6 +4,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 #include "DeviceDelegatePicoVR.h"
+#include "DeviceUtils.h"
 #include "ElbowModel.h"
 #include "VRBrowserPico.h"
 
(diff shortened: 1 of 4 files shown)


bluemarvin:
I should say, this was on the G2. I think the issue is that Pico recenter doesn't just recenter the heading, they recenter the pitch of the head too. I'll test on the Neo 2.

bluemarvin:
It does work as expected on the Neo2. I guess I should compare how it works on the Go.

bluemarvin:
The Go works like I expected where only the heading is re-centered. So re-centered the pitch is something unique to the pico G2. I'm not sure what the correct solution for this is.

app/src/picovr/cpp/DeviceDelegatePicoVR.cpp
```diff
@@ -286,9 +288,8 @@ DeviceDelegatePicoVR::GetReorientTransform() const {
 
 void
 DeviceDelegatePicoVR::SetReorientTransform(const vrb::Matrix& aMatrix) {
-  // Until we can get the new heading don't reset it on the Neo 2
-  if (m.type != kTypeNeo2) {
-    // m.reorientMatrix = aMatrix;
+  if (m.type == kTypeNeo2) {
+    m.reorientMatrix = aMatrix;
   }
```
bluemarvin: I just disabled it for the G2 and I think it all works as expected.

PR #2909, opened 2020-03-03
Add support for Oculus Focus Awareness

bluemarvin:
From the docs:
To enable overlay for testing purposes, enter the following commands in a terminal:

adb shell setprop debug.oculus.dbg_enable_overlay 1
adb shell am force-stop com.oculus.vrshell

After entering these commands, if an app has overlay support enabled in its app manifest, pressing the Oculus button from inside the app will cause the Universal Menu to be displayed as an overlay.

To disable overlay support, enter the following commands:

adb shell setprop debug.oculus.dbg_enable_overlay 0
adb shell am force-stop com.oculus.vrshell


bluemarvin:
You will need to pull the latest 3rd party in order to build this.

bluemarvin:
Okay, guess it doesn’t work yet. There isn’t any thing I do to make it visible as oculus handles the drawing. I guess we can report this to oculus.

bluemarvin:
I'm going to close this until it actually works. Don't want this being turned on by oculus and then breaking FxR.

PR #2915, opened 2020-03-04
Load IP based URIs instead of doing a google search

### LINKED ISSUE #2914: Load local IP URLs instead of doing a google search
e.g. If you type 192.168.0.1 or localhost in the URL bar it does a Google search instead of trying to load the IP address. This is really annoying for devs, who are used to not to type the whole protocol when loading a local url.

MortimerGoro:
Fixes #2914

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/views/NavigationURLBar.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/views/NavigationURLBar.java
@@ -340,6 +340,10 @@ public class NavigationURLBar extends FrameLayout {
             if (!hasProtocol && !urlText.contains(" ") && UrlUtils.isDomain(urlText)) {
                 urlText = "https://" + urlText;
                 hasProtocol = true;
+            } else if (!hasProtocol && !urlText.contains(" ") && UrlUtils.isIPUri(urlText)) {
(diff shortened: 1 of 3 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/views/NavigationURLBar.java
```diff
@@ -340,6 +340,10 @@ public void handleURLEdit(String text) {
             if (!hasProtocol && !urlText.contains(" ") && UrlUtils.isDomain(urlText)) {
                 urlText = "https://" + urlText;
                 hasProtocol = true;
+            } else if (!hasProtocol && !urlText.contains(" ") && UrlUtils.isIPUri(urlText)) {
+                String protocol = UrlUtils.isLocalIP(urlText) ? "http://" : "https://";
+                urlText = protocol + urlText;
+                hasProtocol = true;
```
bluemarvin: I was thinking about ways we could degrade from https:// to http:// in FxR so we always try secure first.

MortimerGoro: We can address that as a follow-up, we also need to handle that situation for normal URLs. Localhost and most of the local IPs are http, that's why I defaulted that way for now.

PR #2922, opened 2020-03-04
Add long press reload that bypasses cache

bluemarvin:


bluemarvin:
@MortimerGoro @keianhzo I'm not convinced the developer switch is necessary. I would okay with removing it and just having long press reload. Also, there is a bypass proxy flag. Should we support that too?

MortimerGoro:
@bluemarvin I'd keep the developer switch, it can help with discoverability and if you have to test a page tons of times it's quicker to use normal click than reload.

bluemarvin:
@MortimerGoro @keianhzo I think this is ready to land. I think we should go with out changing the color or tooltip for now. Opinions?

bluemarvin:
> In Pico devices (I guess related to Android 8) I get also the click event after the logn click.

Is there an Android 8 way to handle this?

bluemarvin:
@keianhzo Okay, I think it is fixed.

Softvision-GeluHaiduc:
I’ve performed exploratory testing on this new feature using the Oculus Go, HTC Vive Focus Plus, Pico Neo2 and Pico G2 4K headsets, logged all the issues found and also written test cases to cover the new functionality.

Issue #1080, opened 2019-04-13
Consider supporting a no-controller mode


bluemarvin:
## Hardware

Pico G2
## Context

The Pico G2 has buttons on the head set that makes a no-controller option possible. I don't believe it would be much work to support this mode on the device.

bluemarvin:
![IMG_0652](https://user-images.githubusercontent.com/3066/56072164-ff136380-5d48-11e9-8ff3-20f5c9beafc9.jpg)
Top: app button
Middle: confirm button
Bottom: home button

RealMG:
Also, have this trouble.
This might be caused by the limitation of VivePort, you could meet the same experience inside HTC VivePort

jvonitter:
duplicated by #1578

cjyjae35:
![](https://github.trello.services/images/mini-trello-icon.png) [DE-80 Gaze Mode](https://trello.com/c/vKHoXnV6/488-de-80-gaze-mode)

bluemarvin:
Fixed by #2959

PR #2929, opened 2020-03-06
Fix for the Pico headset buttons

keianhzo:
Related to #1080 

This PR fixes the issue with the headset button in Pico devices. Also re-enables the back button for quitting the app in Pico.

For some reason before the first back headset click the system `Window` has the `OffscreenPresentation` as the `Window.Callback` target. As we are not currently overriding `dispatchKeyEvent` in the `OffscreenPresentation`, that event gets lost. After that click it seems that the callback target is set to the `VRBBrowserActivity` and the following events are correctly forwarded.
Still unanswered why that first headset button press sets the `Activity` as the `Window.Callback`.

### DIFF
--- a/app/src/picovr/java/org/mozilla/vrbrowser/PlatformActivity.java
+++ b/app/src/picovr/java/org/mozilla/vrbrowser/PlatformActivity.java
@@ -122,7 +122,7 @@ public class PlatformActivity extends VRActivity implements RenderInterface, CVC
 
     @Override
     public void onBackPressed() {
-        super.onBackPressed();
+        // Eat the back button.
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


app/src/picovr/java/org/mozilla/vrbrowser/PlatformActivity.java
```diff
@@ -122,7 +122,7 @@ protected void onDestroy() {
 
     @Override
     public void onBackPressed() {
-        // Eat the back button.
+        super.onBackPressed();
```
bluemarvin: Do we want this in this commit?

keianhzo:
@bluemarvin I've reverted the back button behavior.

PR #2930, opened 2020-03-06
Update grade & build tools

MortimerGoro:


bluemarvin:
This needs to wait until I get the docker image updated. Currently blocked on a slow curl upload to autograph.

daoshengmu:
The failure is because `robolectric` needs to use SDK 28, otherwise, you can choose to upgrade it to `4.3.1`. However, it needs Java 9.

keianhzo:
@daoshengmu It seems that we can force Robolectric to use SDK 28. I've pushed a fix.

daoshengmu:
> @daoshengmu It seems that we can force Robolectric to use SDK 28. I've pushed a fix.

It works properly for me. GJ!

PR #2936, opened 2020-03-10
Dual controller typing

### LINKED ISSUE #2645: Support both hand controllers for keyboard input
From store review.

image

requesting both hand controllers support for keyboard input (for smoother text editing although currently FxR seems to support only right hand controller).

keianhzo:
Closes #2645 Adds support for dual controller typing. 

Android doesn't support multiple device input in a clean way. The viewgroup clears the hovers from the other hovered items when a new child is hovered which causes a hover fight when more than one device is hovering elements from the same layout. When the view drawing is handled this is easier to workaround but in our case that would mean to rewrite quite a lot of code at this point. 

This implementation add support for dual hovering for the keyboard views (characters, symbols and numbers) and for the rest of the views in the keyboard widget (domains, languages and autocompletion) the first device entering them has preference.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/KeyboardWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/KeyboardWidget.java
@@ -63,9 +63,11 @@ import org.mozilla.vrbrowser.ui.views.CustomKeyboardView;
 import org.mozilla.vrbrowser.ui.views.KeyboardSelectorView;
 import org.mozilla.vrbrowser.ui.widgets.dialogs.VoiceSearchWidget;
 import org.mozilla.vrbrowser.utils.StringUtils;
+import org.mozilla.vrbrowser.utils.ViewUtils;
 
 import java.util.ArrayList;
 import java.util.Locale;
+import java.util.concurrent.atomic.AtomicInteger;
 
 
 public class KeyboardWidget extends UIWidget implements CustomKeyboardView.OnKeyboardActionListener, AutoCompletionView.Delegate,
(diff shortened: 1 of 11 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/KeyboardWidget.java
```diff
@@ -109,6 +114,11 @@
     private boolean mInternalDeleteHint = false;
     private Session mSession;
     private boolean mInputRestarted = false;
+    public AtomicInteger mPopUpHoverDeviceId = new AtomicInteger(-1);
+    public AtomicInteger mLanguageHoverDeviceId = new AtomicInteger(-1);
+    public AtomicInteger mDomainHoverDeviceId = new AtomicInteger(-1);
+    public AtomicInteger mAutoCompleteHoverDeviceId = new AtomicInteger(-1);
+    public AtomicInteger mControlHoverDeviceId = new AtomicInteger(-1);
```
bluemarvin: Are these accessed from different threads? If not, they don't need to be atomic I think.

keianhzo: I only did that to be able to pass the value by reference but it doesn't make much sense, I've updated it.

bluemarvin: I think if you do need to pass something that doesn't need to be accessed from multiple thread, the `Integer` type will work. Atomic adds some overhead. We have some code in the `onLoadRequest()` using `AtomicInteger`s that we should probably change.

keianhzo:
@MortimerGoro fixed. Regarding:

> Long press a letter to show the extra letters menu: you can't use the other controller to select a letter from that menu

Android doesn't support multiple device input in a clean way. The viewgroup clears the hovers from the other hovered items when a new child is hovered which causes a hover fight when more than one device is hovering elements from the same layout. When the view drawing is handled this is easier to workaround but in our case that would mean to rewrite quite a lot of code at this point.

Softvision-RemusDranca:
We tested and logged multiple issue regarding this feature. We also finished creating the test cases for it. Considering this I will remove the QA label.

PR #2950, opened 2020-03-12
Cleanup warnings in noapi PlatformActivity

bluemarvin:


app/src/noapi/java/org/mozilla/vrbrowser/PlatformActivity.java
```diff
@@ -252,6 +241,7 @@ private void setupUI() {
             }
             return false;
         });
+        setImmersiveSticky();
```
keianhzo: Why is this required now? Android 10?

bluemarvin: It seems to lose it some times on my phone so I added it a few more place to try and reset it for when it gets unset.

PR #2951, opened 2020-03-12
Clean up warnings from VRBrowserActivity

bluemarvin:


app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
```diff
@@ -1468,7 +1471,7 @@ public boolean canOpenNewWindow() {
     }
 
     @Override
-    public void openNewWindow(String uri) {
+    public void openNewWindow(@NonNull String uri) {
```
keianhzo: This is being called from https://github.com/MozillaReality/FirefoxReality/blob/2516a84dcc5b425a9178dc3f441938ce57e8d2b6/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/menus/ContextMenuWidget.java#L84 with a `Nullable `string.

app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
```diff
@@ -1227,36 +1230,36 @@ public void updateVisibleWidgets() {
...
     @Override
-    public void startWidgetResize(final Widget aWidget, float aMaxWidth, float aMaxHeight, float minWidth, float minHeight) {
+    public void startWidgetResize(@NonNull final Widget aWidget, float aMaxWidth, float aMaxHeight, float minWidth, float minHeight) {
         mWindows.enterResizeMode();
         queueRunnable(() -> startWidgetResizeNative(aWidget.getHandle(), aMaxWidth, aMaxHeight, minWidth, minHeight));
     }
 
     @Override
-    public void finishWidgetResize(final Widget aWidget) {
+    public void finishWidgetResize(@NonNull final Widget aWidget) {
```
keianhzo: Should never be null but could potentially be null when called from: https://github.com/MozillaReality/FirefoxReality/blob/2516a84dcc5b425a9178dc3f441938ce57e8d2b6/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/NavigationBarWidget.java#L985

app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
```diff
@@ -1227,36 +1230,36 @@ public void updateVisibleWidgets() {
     }
 
     @Override
-    public void startWidgetResize(final Widget aWidget, float aMaxWidth, float aMaxHeight, float minWidth, float minHeight) {
+    public void startWidgetResize(@NonNull final Widget aWidget, float aMaxWidth, float aMaxHeight, float minWidth, float minHeight) {
```
keianhzo: Should never be null but could potentially be null when called from:
https://github.com/MozillaReality/FirefoxReality/blob/2516a84dcc5b425a9178dc3f441938ce57e8d2b6/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/NavigationBarWidget.java#L985

app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
```diff
@@ -1169,7 +1172,7 @@ public void addWidget(Widget aWidget) {
     }
 
     @Override
-    public void updateWidget(final Widget aWidget) {
+    public void updateWidget(@NonNull final Widget aWidget) {
```
keianhzo: I don't think we can guarantee that `aWidget `will never be null.

app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
bluemarvin: All of the `@NonNull` additions were because the base class has it set. So we either need to fix the base class or fix our usage of it.

app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
bluemarvin: Again, base class has it set to `@NonNull`

app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
bluemarvin: Then we need to fix the base class.

app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
bluemarvin: We need need to fix the base class the.

keianhzo:
I'd update the base class too and add the null checks. For what I've seen in all those cases there are chances that the parameter is null.

bluemarvin:
Okay, I try and look at this when I have more time.

PR #2959, opened 2020-03-13
Pico Gaze Mode support

keianhzo:
Related to #1080 This PR adds support for Pico devices Gaze mode.

Design pending, it will be addressed as a follow up when ready.


app/src/main/cpp/ExternalVR.cpp
```diff
@@ -417,9 +417,12 @@ ExternalVR::PushFramePoses(const vrb::Matrix& aHeadTransform, const std::vector<
...
       immersiveController.triggerValue[i] = controller.immersiveTriggerValues[i];
     }
+    for (int j = 0; j< controller.numButtons; ++j) {
+      immersiveController.triggerValue[j] = controller.immersiveTriggerValues[j];
+    }
     immersiveController.numAxes = controller.numAxes;
-    for (int i = 0; i< controller.numAxes; ++i) {
-      immersiveController.axisValue[i] = controller.immersiveAxes[i];
+    for (int j = 0; j< controller.numAxes; ++j) {
+      immersiveController.axisValue[j] = controller.immersiveAxes[j];
```
bluemarvin: I'm not sure what is going on here. It looks like trigger value is being set twice and in the last for loop `i` was replaced with `j`? I think this can be reverted? You could add a space between `i<` though.

app/src/picovr/cpp/DeviceDelegatePicoVR.cpp
```diff
@@ -227,6 +241,10 @@ struct DeviceDelegatePicoVR::State {
       }
     }
   }
+
+  int32_t GazeModeIndex() {
+    return type == kTypeNeo2 ? 2 : 1;
+  }
```
bluemarvin: Is this right? The G2 actually has two controllers, one for left and one for right and only one is ever visible. I would think the gaze index would be 2 for both G2 and Neo2.

app/src/picovr/java/org/mozilla/vrbrowser/PlatformActivity.java
```diff
@@ -146,7 +147,20 @@ protected void onDestroy() {
 
     @Override
     public void onBackPressed() {
-        // Eat the back button.
+        super.onBackPressed();
```
bluemarvin: I don't think we should do this until we have UI to prevent accidental exit. The home button always works.

bluemarvin:
I've confirmed that if you switch controller to left hand on G2, you get the gaze controller instead in FxR.

app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -189,6 +189,7 @@ struct BrowserWorld::State {
   std::unordered_map<vrb::Node*, std::pair<Widget*, float>> depthSorting;
   std::function<void(device::Eye)> drawHandler;
   std::function<void()> frameEndHandler;
+  bool wasInGazeMode;
```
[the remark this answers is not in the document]
keianhzo: Fixed

app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -714,6 +735,7 @@ BrowserWorld::RegisterDeviceDelegate(DeviceDelegatePtr aDelegate) {
     m.leftCamera = m.device->GetCamera(device::Eye::Left);
     m.rightCamera = m.device->GetCamera(device::Eye::Right);
     ControllerDelegatePtr delegate = m.controllers;
+    delegate->SetGazeModeIndex(m.device->GazeModeIndex());
```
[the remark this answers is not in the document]
keianhzo: Don't get this, what do you mean exactly?

app/src/main/cpp/ControllerContainer.cpp
```diff
@@ -35,6 +35,7 @@ struct ControllerContainer::State {
   GeometryPtr beamModel;
   bool visible = false;
   vrb::Color pointerColor;
+  int gazeIndex;
```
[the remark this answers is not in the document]
keianhzo: Fixed

app/src/main/cpp/ExternalVR.cpp
keianhzo: I just changed the name of the for variable as we were using the same name as in the outer for.

app/src/picovr/cpp/DeviceDelegatePicoVR.cpp
keianhzo: I wasn't aware of this, makes sense now seeing the controller settings where you can change to left/right mode. I've updated to not override the second controller.

app/src/picovr/java/org/mozilla/vrbrowser/PlatformActivity.java
keianhzo: Reverted

keianhzo: I guess we want a v10 task to add this UI, right?

keianhzo:
Right, wasn't aware of the second controller for G2 but makes sense seeing the controller settings and the fact that you can switch between left/right modes. I've fixed it.

app/src/main/cpp/ControllerContainer.cpp
```diff
@@ -200,33 +201,36 @@ ControllerContainer::CreateController(const int32_t aControllerIndex, const int3
...
-      beamTransform->SetTransform(aBeamTransform);
-      controller.beamParent = beamTransform;
-      controller.beamToggle->AddNode(beamTransform);
-    }
-    controller.transform->AddNode(controller.beamToggle);
-    controller.beamToggle->ToggleAll(false);
-    if (m.beamModel && controller.beamParent) {
-      controller.beamParent->AddNode(m.beamModel);
+      VRB_ERROR("Failed to add controller model");
     }
```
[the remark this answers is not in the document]
keianhzo: That's where it is, right?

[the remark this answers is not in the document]
bluemarvin: Let me look at the code. Diffs can be confusing sometimes.

app/src/main/cpp/BrowserWorld.cpp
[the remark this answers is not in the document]
bluemarvin: `m.controllers->SetGaceModeIndex(m.device->GazeModeIndex());`

app/src/main/cpp/ControllerContainer.cpp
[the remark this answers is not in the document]
bluemarvin: Looking at the code, this is fine, I must have been confused by the diff.

app/src/main/cpp/BrowserWorld.cpp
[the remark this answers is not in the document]
bluemarvin: It's not a big deal, I see we use the delegate later in the code.

app/src/picovr/java/org/mozilla/vrbrowser/PlatformActivity.java
bluemarvin: I'm not sure if we can get it into v10? Depends on how quick we can get a design. Although I would be okay with creating a place holder. I guess we could always just exit which is what android apps do but I've always hated that about android.

app/src/main/cpp/BrowserWorld.cpp
bluemarvin: I would prefer: `bool wasInGazeMode = false;`

app/src/main/cpp/ControllerContainer.cpp
```diff
@@ -35,12 +35,14 @@ struct ControllerContainer::State {
   GeometryPtr beamModel;
   bool visible = false;
   vrb::Color pointerColor;
+  int gazeIndex;
```
bluemarvin: Same here, `int gazeIndex = -1`
Bellow it is initialized to `0` which I think not what we want?

app/src/main/cpp/ExternalVR.cpp
```diff
@@ -414,12 +414,12 @@ ExternalVR::PushFramePoses(const vrb::Matrix& aHeadTransform, const std::vector<
...
-    for (int i = 0; i< controller.numButtons; ++i) {
-      immersiveController.triggerValue[i] = controller.immersiveTriggerValues[i];
+    for (int j = 0; j< controller.numButtons; ++j) {
+      immersiveController.triggerValue[j] = controller.immersiveTriggerValues[j];
     }
     immersiveController.numAxes = controller.numAxes;
-    for (int i = 0; i< controller.numAxes; ++i) {
-      immersiveController.axisValue[i] = controller.immersiveAxes[i];
+    for (int j = 0; j< controller.numAxes; ++j) {
+      immersiveController.axisValue[j] = controller.immersiveAxes[j];
```
bluemarvin: If you are going to update this, please add a space `j<` -> `j <`

app/src/picovr/cpp/DeviceDelegatePicoVR.cpp
```diff
@@ -103,14 +106,18 @@ struct DeviceDelegatePicoVR::State {
...
       controllers[index].index = index;
       if (index == 0) {
         controllers[index].hand = ElbowModel::HandEnum::Right;
       } else {
         controllers[index].hand = ElbowModel::HandEnum::Left;
       }
-      controllers[index].is6DoF = true;
+      if (index != gazeIndex) {
+        controllers[index].is6DoF = true;
+      }
```
bluemarvin: Not sure if it matters but I guess technically, the Neo2 gaze controller is 6DoF.

app/src/picovr/cpp/DeviceDelegatePicoVR.cpp
```diff
@@ -373,6 +401,23 @@ DeviceDelegatePicoVR::StartFrame() {
 
   m.cameras[0]->SetHeadTransform(head);
   m.cameras[1]->SetHeadTransform(head);
+
+
+  // Update te gaze mode state based on controllers availability
+  m.isInGazeMode = true;
+  for (int32_t i = 0; i < m.controllers.size(); ++i) {
+    if (i != m.gazeIndex) {
+      m.isInGazeMode &= !m.controllers[i].enabled;
```
bluemarvin: I'm not sure I like using a bitwise and operator with booleans. Maybe:
```
  if (!m.controllers[i].enabled) {
    m.isInGazeMode = false;
    break;
  }
```

app/src/picovr/cpp/DeviceDelegatePicoVR.cpp
```diff
@@ -373,6 +401,23 @@ DeviceDelegatePicoVR::StartFrame() {
...
+    }
+  }
+
+  if (m.isInGazeMode) {
+    for (int i=0; i<m.controllers.size(); i++) {
+      m.controllers[i].enabled = (i == m.gazeIndex);
+    }
+    m.controllers[m.gazeIndex].transform = GetHeadTransform();
+  }
```
bluemarvin: Maybe instead:
```
  if (m.isInGazeMode) {
    m.controllers[m.gazeIndex].enabled = m.isInGazeMode;
    m.controllers[m.gazeIndex].transform = GetHeadTransform();
  }
```

app/src/main/cpp/BrowserWorld.cpp
keianhzo: Fixed

app/src/main/cpp/ExternalVR.cpp
keianhzo: Fixed

app/src/picovr/cpp/DeviceDelegatePicoVR.cpp
keianhzo: Updated

app/src/picovr/cpp/DeviceDelegatePicoVR.cpp
keianhzo: Updated

app/src/picovr/cpp/DeviceDelegatePicoVR.cpp
keianhzo: Updated

app/src/picovr/cpp/DeviceDelegatePicoVR.cpp
keianhzo: Updated

keianhzo:
@bluemarvin Updated

app/src/main/cpp/ControllerContainer.cpp
keianhzo: Fixed

bluemarvin:
@MortimerGoro I'm going to land this so QA has time to test it. Any issues can be filed as a follow-up.

Softvision-GeluHaiduc:
I’ve performed exploratory testing on this new feature using the Pico G2 4K and Pico Neo2 headsets, logged all the issues found and also written test cases to cover the new functionality.

PR #2961, opened 2020-03-13
Prevent duplicated callSurfaceChanged call when exiting immersive mode

### LINKED ISSUE #2957: [Pico] The open windows ca no longer be scrolled after exiting immersive mode
**[Affected Versions]:**
- Firefox Reality version 10-40730015 (b958efbf)

**[Affected Platforms]:**
- Pico Neo2 - Android 8.1.0

**[Prerequisites]:**
- Make sure to have 3 windows opened.

**[Steps to reproduce]:**
1. In one of the windows navigate to the following website: “https://spiderman.webvr.link/”
2. Click the “VR” button from the corner.
3. Press the Back button on the controller or on the headset.
4. Try to scroll the other windows.

**[Expected results]:**
- The windows can be scrolled.

**[Actual results]:**
- The windows can no longer be scrolled.

**[Notes]:**
- The issue is not present in the new windows created.
- Sometimes the windows are completely frozen.
- If a new website is loaded in the window the issue is still present. 
- Attached a screen recording of the issue [here](https://drive.google.com/open?id=1C7NG6A4cuNAQVBUEgpqZN7Nsxkb_g2Fk).

MortimerGoro:
Fixes #2957

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WindowWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WindowWidget.java
@@ -314,7 +314,7 @@ public class WindowWidget extends UIWidget implements SessionChangeListener,
         super.onResume();
         if (isVisible() || mIsInVRVideoMode) {
             mSession.setActive(true);
-            if (!SettingsStore.getInstance(getContext()).getLayersEnabled()) {
+            if (!SettingsStore.getInstance(getContext()).getLayersEnabled() && !mSession.hasDisplay()) {
                 // Ensure the Gecko Display is correctly recreated.
                 // See: https://github.com/MozillaReality/FirefoxReality/issues/2880
                 callSurfaceChanged();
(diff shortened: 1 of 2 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WindowWidget.java
```diff
@@ -314,7 +314,7 @@ public void onResume() {
         super.onResume();
         if (isVisible() || mIsInVRVideoMode) {
             mSession.setActive(true);
-            if (!SettingsStore.getInstance(getContext()).getLayersEnabled()) {
+            if (!SettingsStore.getInstance(getContext()).getLayersEnabled() && !mSession.hasDisplay()) {
```
bluemarvin: I wonder if we can just check if the mSession is active and then only call `setActive(true)` and `callSurfaceChanged()` if it isn't?

MortimerGoro: I think is safer to keep this check. My worry is that other place of code calls SetActive before this method and in that case we would miss the surfaceChanged call

PR #2964, opened 2020-03-16
Make sure domains without protocol are force to https

keianhzo:
We get a lot of http->https forwarded history entries because the InlineAutocompleEditText return domains without protocol. This PR forces all domains without protocol to https so we don't have many http->https forward history entries.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/search/suggestions/SuggestionsProvider.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/search/suggestions/SuggestionsProvider.java
@@ -60,7 +60,7 @@ public class SuggestionsProvider {
 
     private String getSearchURLOrDomain(String text) {
         if (UrlUtils.isDomain(text)) {
-            return text;
+            return UrlUtils.ensureHttps(text);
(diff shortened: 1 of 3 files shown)


bluemarvin:
We made this change on purpose so that gecko would figure it out for us. If we need something in a-c to not show the redirect we should do that instead.

bluemarvin:
@MortimerGoro && @keianhzo Is this something we want to pursue? I'm certain we do not want to force https.

keianhzo:
@bluemarvin The load flow for URL without domain in GV is:

- onLoadRequest - google.com
- onLoadRequest - http://www.google.com
- onLoadRequest - https://www.google.com/?gws_rd=ssl

Which leads to at least two history entries. It's the same for Fenix. I guess the best solution would be for GV to use htts by default in case there is no protocol, this is just a workaround for that.

bluemarvin:
I don’t see a problem with two history entries. I was asked to change the code so that gecko can decide what protocol to use and to not force it to be https. There will be an extension that handles that at some point users can install.

keianhzo:
Closing this as it's not the right approach.

PR #2970, opened 2020-03-17
Reduce WebVR idle times

MortimerGoro:


bluemarvin:
When going into immersive mode I see a one frame of immersive, then one frame of the spinner an then the immersive mode starts. Is there any way to fix this?

PR #2980, opened 2020-03-18
Handle version code conversion in debug builds

keianhzo:
Handle version code conversion in debug builds


bluemarvin:
Did my change break this?

PR #2986, opened 2020-03-18
Clean up warning from VRBrowserActivity

bluemarvin:


bluemarvin:
@keianhzo I think I addressed your concerns from the previous PR. I thought it would reopen it, but I guess github just creates a new PR.

PR #2987, opened 2020-03-18
Upgrade Docker image to latest build tools

bluemarvin:


bluemarvin:
@MortimerGoro This is the docker file I used to create the current image we are using with the gradle upgrades.

PR #2998, opened 2020-03-19
Add support for Oculus App Linking

MortimerGoro:
I tried to test this using this code from a different app id:
```
ovrApplicationOptionsHandle options = ovr_ApplicationOptions_Create();
ovrID id = 2180252408763702;
ovr_ApplicationOptions_SetDeeplinkMessage(options, "https://www.mozilla.org");
ovr_Application_LaunchOtherApp(id, options);
```
But it  shows the Oculus Store page with a update button.


bluemarvin:
I think to test you might need to upload a build to the alpha channel?

bluemarvin:
Also, I wanted to use JSON for the message body so we could easily add things in the future. Or even allow multiple URLs to be sent at once.

MortimerGoro:
Ok, I'll add a json format and then we can test with alpha channel

bluemarvin:
Great, just let me know when you are ready to test it.

MortimerGoro:
@bluemarvin ready for testing

PR #3000, opened 2020-03-19
Fix crash if onLoadError uri parameter is null

MortimerGoro:
Crash report:
https://crash-stats.mozilla.com/report/index/edff4a20-a2d4-41a4-9dd5-773d10200319

I reproduced it sending invalid uris in intents

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/utils/InternalPages.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/utils/InternalPages.java
@@ -3,6 +3,8 @@ package org.mozilla.vrbrowser.utils;
 import android.content.Context;
 import android.util.Base64;
 
+import androidx.annotation.Nullable;
+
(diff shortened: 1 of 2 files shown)


app/src/common/shared/org/mozilla/vrbrowser/utils/InternalPages.java
```diff
@@ -134,9 +136,10 @@ public static String createErrorPageDataURI(Context context,
                 showSSLAdvanced = false;
         }
 
-        html = html
-                .replace("%url%", uri)
-                .replace("%advancedSSLStyle%", showSSLAdvanced ? "block" : "none");
+        if (uri != null) {
+            html = html.replace("%url%", uri);
+        }
```
bluemarvin: I wonder if it would be better to do:
`html = html.replace("%url%", uri != null ? uri : "NULL");`

PR #3019, opened 2020-03-20
Add WebXR settings and permission flow

### LINKED ISSUE #2062: Settings for WebXR

MortimerGoro:
Fixes #2062


bluemarvin:
> LGTM although I notice we need to do mulit-lang support for WebXR permission page. We can open another bug to work on it.

What do you mean by multi-lang? The localization happens automatically after the PR lands.

daoshengmu:
> 
> 
> > LGTM although I notice we need to do mulit-lang support for WebXR permission page. We can open another bug to work on it.
> 
> What do you mean by multi-lang? The localization happens automatically after the PR lands.

Yep. That's what I mean. It would be great to have the localization automatically.

bluemarvin:
Hiding the icon for the feed can probably be a follow up.

PR #3022, opened 2020-03-20
Developer switch for enabling experimental multi-e10s

bluemarvin:


bluemarvin:
Hm, I'm not sure if we should land this yet. Maybe make it only available in debug builds?

flodolo:
> Hm, I'm not sure if we should land this yet. Maybe make it only available in debug builds?

Ping for the question here, especially given the last comment in the PR ;-)
https://github.com/mozilla-l10n/android-l10n/pull/186#discussion_r396394237

bluemarvin:
The problem is there isn't a good name for it. Multiprocessing is already enabled by default. Multi-multiprocessing isn't any better. It's for developers to test the feature so I'm not sure what else we can call it, it will go away once it is enabled by default.

PR #3028, opened 2020-03-23
Do not show WebXR icon in the content feed

MortimerGoro:


app/src/common/shared/org/mozilla/vrbrowser/ui/viewmodel/WindowViewModel.java
```diff
@@ -285,6 +289,16 @@ public void onChanged(Spannable aUrl) {
         }
     };
 
+    private Observer<Spannable> mIsContentFeedObserver = new Observer<Spannable>() {
+        @Override
+        public void onChanged(Spannable o) {
+            String current = url.getValue().toString();
```
bluemarvin: Where does `url` come from?

bluemarvin: Okay, I think I see. The variable naming in this file is all over the place. We have a member variable name `url` but also local variables called `url`. This file really needs to be cleaned up.

app/src/common/shared/org/mozilla/vrbrowser/ui/viewmodel/WindowViewModel.java
```diff
@@ -285,6 +289,16 @@ public void onChanged(Spannable aUrl) {
...
+            String feed = getApplication().getString(R.string.homepage_url);
```
keianhzo: Can we move this to UrlUtils? We already have some helper methods there for identifying specific URLs.

MortimerGoro: Done

PR #3037, opened 2020-03-24
Fixes #2424 Enhanced Tracking Protection

### LINKED ISSUE #2424: Update ETP
Our tracking protection is out of sync with the rest of the Firefoxen. What are the options for us to update/moderninze it to be more in line with the rest of the family?

Note that our current implementation also blocks GA tracking, particularly on the content feed, which is something we'd like to enable.

keianhzo:
Fixes #2424 Enhanced Tracking Protection


keianhzo:
There are a few common issues with the WebXR settings PR:
- Dialog not closed when switching windows
- Dialog not closed when clicking on the button again
- Hide buttons in the content feed.

I'll fix those when merging the pending PRs that already address those.

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/EngineProvider.kt
```diff
@@ -25,7 +25,11 @@ object EngineProvider {
 
             builder.crashHandler(CrashReporterService::class.java)
             builder.contentBlocking(ContentBlocking.Settings.Builder()
-                    .antiTracking(ContentBlocking.AntiTracking.AD or ContentBlocking.AntiTracking.SOCIAL or ContentBlocking.AntiTracking.ANALYTIC)
+                    .antiTracking(
+                            ContentBlocking.AntiTracking.AD or
+                            ContentBlocking.AntiTracking.SOCIAL or
+                            ContentBlocking.AntiTracking.ANALYTIC)
+                    .enhancedTrackingProtectionLevel(SettingsStore.getInstance(context).trackingProtectionLevel)
```
bluemarvin: I think we only want one or the other set at one time?

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/Session.java
```diff
@@ -1239,6 +1231,25 @@ public void onContentBlocked(@NonNull final GeckoSession session, @NonNull final
...
+        }
+
+        if ((event.getAntiTrackingCategory() & ContentBlocking.AntiTracking.CONTENT) != 0) {
+            Log.i(LOGTAG, "Loading Content: " + event.uri);
+        }
+
+        if ((event.getAntiTrackingCategory() & ContentBlocking.AntiTracking.SOCIAL) != 0) {
+            Log.i(LOGTAG, "Loading Social: " + event.uri);
+        }
+    }
```
bluemarvin: These need to be `Log.d` so they get stripped out of release builds or we will get rejected from HTC store.

app/src/common/shared/org/mozilla/vrbrowser/browser/tracking/TrackingProtectionPolicy.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.Collections;
+import java.util.List;
+
+public class TrackingProtectionPolicy {
+
+    @IntDef(value = { ETP_NONE, ETP_DEFAULT, ETP_STRICT})
+    public @interface ETP_Level {}
+    public static final int ETP_NONE = ContentBlocking.EtpLevel.NONE;
+    public static final int ETP_DEFAULT = ContentBlocking.EtpLevel.DEFAULT;
+    public static final int ETP_STRICT = ContentBlocking.EtpLevel.STRICT;
```
bluemarvin: I'm not sure why we need to do this.

app/src/common/shared/org/mozilla/vrbrowser/browser/tracking/TrackingProtectionPolicy.java
```diff
@@ -0,0 +1,133 @@
...
+        CONTENT(1 << 4),
+        TEST(1 << 5),
+        CRYPTOMINING(1 << 6),
+        FINGERPRINTING(1 << 7),
+        MOZILLA_SOCIAL(1 << 8),
+        SCRIPTS_AND_SUB_RESOURCES(1 << 9999),
+        RECOMMENDED(AD.value + ANALYTICS.value + SOCIAL.value +
+                TEST.value + MOZILLA_SOCIAL.value + CRYPTOMINING.value),
+        STRICT(RECOMMENDED.value + SCRIPTS_AND_SUB_RESOURCES.value +
+                FINGERPRINTING.value);
```
bluemarvin: And this, should we just use the values in GeckoView so if they change we don't need to keep them synchronized?

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/EngineProvider.kt
droeh: These are independent, actually. `antiTracking` controls what *content* gets blocked when `GeckoSessionSettings.useTrackingProtection` is true, while `enhancedTrackingProtectionLevel` controls what *cookies* get blocked when our cookie behavior is `ACCEPT_NON_TRACKERS`.

app/src/common/shared/org/mozilla/vrbrowser/browser/tracking/TrackingProtectionPolicy.java
```diff
@@ -0,0 +1,133 @@
...
+        NONE(0),
+        AD(1 << 1),
+        ANALYTICS(1 << 2),
+        SOCIAL(1 << 3),
+        CONTENT(1 << 4),
+        TEST(1 << 5),
+        CRYPTOMINING(1 << 6),
+        FINGERPRINTING(1 << 7),
+        MOZILLA_SOCIAL(1 << 8),
+        SCRIPTS_AND_SUB_RESOURCES(1 << 9999),
```
droeh: `1 << 9999` is very alarming unless Java does something cute here that I'm not aware of.

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/EngineProvider.kt
bluemarvin: The way I understood what was explained to me is that we only want one active at a time. If we are blocking cookies we don't want to be blocking content as that is the purpose of ETP, it just blocks the tracking cookies and the content is all loaded which reduce breakage, where content blocking is more strict but also causes more breakage.

app/src/common/shared/org/mozilla/vrbrowser/browser/tracking/TrackingProtectionPolicy.java
bluemarvin: yikes, that is bad, I assume it is just zero which is probably not what they intended. I'm not sure we even want all this.

bluemarvin:
@keianhzo droeh says this is the site to use for testing: https://senglehardt.com/test/trackingprotection/test_pages/

keianhzo:
@bluemarvin @droeh I've updated and removed all redundant code. I've been testing with the site above and I think the results are good, also are the sames as Fenix.

bluemarvin:
Seems to work my tests. It would be nice if we could do a v2 where the shield changed colors when it actually blocks some thing and then clicking on that would give more details on what was blocked.

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/EngineProvider.kt
```diff
@@ -25,7 +25,11 @@ object EngineProvider {
 
             builder.crashHandler(CrashReporterService::class.java)
             builder.contentBlocking(ContentBlocking.Settings.Builder()
-                    .antiTracking(ContentBlocking.AntiTracking.AD or ContentBlocking.AntiTracking.SOCIAL or ContentBlocking.AntiTracking.ANALYTIC)
+                    .antiTracking(
+                            ContentBlocking.AntiTracking.AD or
+                            ContentBlocking.AntiTracking.SOCIAL or
+                            ContentBlocking.AntiTracking.ANALYTIC)
```
bluemarvin: I think we should be using the level from the preferences here. I assume it gets over-written later but we should be starting with the correct anti-tracking values I think.

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/EngineProvider.kt
bluemarvin: > These are independent, actually. `antiTracking` controls what _content_ gets blocked when `GeckoSessionSettings.useTrackingProtection` is true, while `enhancedTrackingProtectionLevel` controls what _cookies_ get blocked when our cookie behavior is `ACCEPT_NON_TRACKERS`.

@droeh I guess I'm still confused, I don't see `GeckoSessionSettings.useTrackingProtection()` except when the session is created. Does this mean it never gets turned off?

bluemarvin:
I wonder if we should even show the shield if tracking protection is turned off.

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/EngineProvider.kt
keianhzo: Makes sense, updated

keianhzo:
@bluemarvin Regarding:

> But it says tracking protection is still on and cookies are still getting blocked:

I see the same behavior in Fenix, maybe @droeh can clarify.

And regarding:

> I wonder if we should even show the shield if tracking protection is turned off.

In Desktop/Fenix you can never completely turn ETP off so the icon is always visible. Maybe if ETP is **Off** we could show the disabled icon, change the bubble text to something like "ETP is disabled. Go to setting to turn it on" and make the button to take you to the ETP settings. It's just weird that in desktop and Fenix is always visible and we hide it. Also it is faster to go to the setting from there if you want to re-enable again.

keianhzo:
@bluemarvin updated with hiding the tracking protection icon if disabled.

bluemarvin:
https://github.com/MozillaReality/FirefoxReality/issues/3053

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/EngineProvider.kt
droeh: Sorry for the slow reponse, I'm not used to getting work mail from github!
There are cases where you might want to leave ETP and TP both enabled, for example you may want to enable TP only for one category and ETP for the rest. In general they are only exposed in UI mutually exclusively, though (eg as "Standard" and "Strict" in desktop).

And yeah, I missed that, but you definitely need to update `GeckoSessionSettings.useTrackingProtection()` to control TP, and do it for every session. (Yes, that sucks, it's on the list of things to change in the API.)

I'll keep an eye on github mail and respond quicker if you have any more questions.

PR #3059, opened 2020-03-26
Fix glean test for Oculus channel name.

daoshengmu:
Don't merge. It is for capturing logs.


bluemarvin:
@daoshengmu I assume we can close this now?

daoshengmu:
@bluemarvin I think we can land this small fix for future use cases in Oculus,

PR #3065, opened 2020-03-27
Set up WebXR Room Scale sizes

MortimerGoro:


bluemarvin:
> I think we need to set `Cap_StageParameters` if we are passing in stage size?

I see we are already setting it.

PR #3070, opened 2020-03-30
Fix WebXR quick permission regression. Unify permission messages.

MortimerGoro:


app/src/main/res/values/strings.xml
```diff
@@ -1453,7 +1453,7 @@ the Select` button. When clicked it closes all the previously selected tabs -->
 
     <!-- This string is displayed in the title of the quick WebXR blocking dialog, accessed from the VR icon in the URL bar.
      '%1$s' will be replaced at runtime with the app's name. -->
-    <string name="webxr_block_dialog_message">‘%1$s’ wants to access WebXR API</string>
+    <string name="webxr_permission_dialog_message">WebXR is %1$s for this site. (&lt;a href="%2$s">Learn More&lt;/a>)</string>
```
flodolo: Two questions:
1) What does `WebXR is Firefox for this site.` mean? Because that's what the comment implies
2) Is the HTML supposed to look like this?

MortimerGoro: Ups, the comment is wrong. The real text will be: WebXR is "on/off" for this site

flodolo: And where do "ON" and "OFF" come from? Note that they're adjectives, so they need to be adapted to the name, and you can't reuse generic ones.

flodolo: Any update on this? We can't expose for localization any strings until this is fixed.

PR #3072, opened 2020-03-30
Closes #672 Downloads support

### LINKED ISSUE #672: Implement downloads to device
## Hardware
Oculus Go

## Steps to Reproduce
1. Attempt to download a file in Firefox Reality
2. Notice nothing happens

## Current Behavior
A file is not downloaded.

## Expected Behavior
A file will be downloaded to `/sdcard/Android/data/org.mozilla.vrbrowser/files/Download`.

## Possible Solution
Design wires and conversations are in **issue #635**.

Technically: When a user navigates to a URL with a response header of `Content-Disposition: attachment`, a mime-type that Gecko thinks is downloadable, or a user clicks on a link with a `download` attribute (e.g., `<a href="img_1590128.png" download="screenshot.png"></a>`).

There's a Downloads GeckoView Android Component for this now: https://github.com/mozilla-mobile/android-components/blob/master/components/feature/downloads/README.md

There may be additional work to be done, but let's double check that we are correctly using the Safe Browsing API so that URLs that are known to be threats cannot be downloaded.

keianhzo:
Closes #672 This PR adds Downloads support. Some refactoring has been done to take all the Library panels logic out from the Window class.

Waiting for final design.


bluemarvin:
Seems pretty good. Not sure I like the badge. I also find the save location of internal or external confusing. I have no idea where that would be if I went looking for the files on the device if I were to mount it.

keianhzo:
Updated with final design: https://trello.com/c/iBLotKN4

bluemarvin:
Looks like the download icon is still missing?
org mozilla vrbrowser-20200407-170240

bluemarvin:
The spec also has the download progress like on desktop with a blue badge after download is complete.

keianhzo:
Not sure why you don't see them but both things you mentioned were already there in the last push. I've forced push again and fixed and issue I was seeing with clipping.

keianhzo:
@bluemarvin seems to be related to the firstpaint issue.

bluemarvin:
> @bluemarvin seems to be related to the firstpaint issue.

Which first paint issue? Has it been fixed in master because I rebased and am still seeing the issue.

keianhzo:
@bluemarvin I didn't have time to fix the firstpaint issue today but what you see here it's related to that. You can also reproduce it if you open a new tab in the background from the context menu and switch to that tab.

PR #3091, opened 2020-04-02
Fixes #3087 Use App executors instead of ThreadUtils

### LINKED ISSUE #3087: Use App executors instead of ThreadUtils background thread.
## Configuration

Firefox Reality version: 10
Firefox Reality build ID: 

Hardware: all

## Issue
PR #3076 added a background thread that is used to upload crash reports. Since FxR already has existing worker background threads, the crash reporter should use that instead.

keianhzo:
Fixes #3087 Use App executors instead of ThreadUtils. Removed ThreadUtils as in most cases we can just use the view main looper.


bluemarvin:
TC failure is the unit test failing to download components again.

PR #3106, opened 2020-04-03
Add application notes to crash report

bluemarvin:


bluemarvin:
This depends on https://bugzilla.mozilla.org/show_bug.cgi?id=1626979
Is there any other application data we should include?

PR #3119, opened 2020-04-06
Fix halo artifact on Oculus

### LINKED ISSUE #3109: [Oculus] A Transparent background is wrongly displayed for the menus, prompts and buttons
**[Affected Versions]:**
- Firefox Reality version 10-140941948 (b539410).

**[Affected Platforms]:**
- Oculus Go - Android 7.1.2

**[Prerequisites]:**
- The headset is turned.

**[Steps to reproduce]:**                                        
1. Open the Firefox Reality app.
2. Open the History library so that the background gets dimmed.
3. Observe the area of the navigation buttons.

**[Expected results]:**
- The UI is correctly displayed.

**[Actual results]:**
- A transparent background is visible.

**[Additional Notes]:**
- This issue is also visible for the prompts and menus available.
- The issue is highly visible when the background is dimmed.
- Attached is a screen recording of the issue: [link](https://drive.google.com/open?id=1ZEDE08G6N7E2B0QPjfcQM3eNkvoiV3gq).

bluemarvin:
Fixes #3109

### DIFF
--- a/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
+++ b/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
@@ -146,11 +146,6 @@ struct DeviceDelegateOculusVR::State {
       return;
     }
     initialized = true;
-    std::string version = vrapi_GetVersionString();
-    if (version.find("1.1.32.0") != std::string::npos) {
-      VRB_ERROR("Force layer clip due to driver bug. VRAPI Runtime Version: %s",vrapi_GetVersionString());
-      OculusLayer::sForceClip = true;
-    }
 
     layersEnabled = VRBrowser::AreLayersEnabled();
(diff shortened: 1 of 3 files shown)


bluemarvin:
@MortimerGoro I don't remember why we were using `VRAPI_FRAME_LAYER_BLEND_ONE` previously, but `VRAPI_FRAME_LAYER_BLEND_SRC_ALPHA` seems to work and fixes the halo issue.

MortimerGoro:
We switched to BLEND_ONE in this PR: https://github.com/MozillaReality/FirefoxReality/pull/814
It was to fix a border artifact on the bean.

 I have tested it and it looks good. My only concerns is the behavior might be different in a old runtime? I can’t test it not that I have devices updated

bluemarvin:
I think most people will have newer runtimes. I think it is safe to land.

PR #3096, opened 2020-04-02
Update to new usage of Glean Gradle plugin

mdboom:
Fixes #2697 

This modifies the build to use the Glean Gradle plugin that ships with android-components, rather than directly from Glean.

This means that updating android-components should be much smoother going forward.  Right now, you have to make sure that the version of Glean forUnitTests is the correct one for a particular version of android-components.  With this change, only the version of android-components needs to be updated, and the correct version of other Glean dependencies will be determined automatically.

I'm submitting this as a draft because it does require upgrading android-components to get this feature, and there seems to be plenty of other unrelated breakage that this upgrade causes.

When you are ready to devote effort to updating android-components, this PR should be ready to use then.

Cc: @Dexterp37

PR #3130, opened 2020-04-07
Call Glean initialize prior to set startup metrtics.

### LINKED ISSUE #3129: Distribution channel name in metrics is null after Glean updates.
## Configuration

Firefox Reality version: v.10
Firefox Reality build ID: 

Hardware: 
All devices

## Steps to Reproduce

1. Watch the telemetry query from https://sql.telemetry.mozilla.org/queries/69899/source#176133,
we can see the `distribution_channel_name` in metrics would be null sometimes.

## Current Behavior

After FxR v.9, the `distribution_channel_name` would be null from some situations. I think it causes by the recent Glean updates.

## Expected Behavior

`distribution_channel_name` should be according to `BuildConfig.FLAVOR_platform`.

## Possible Solution

After discussion with Glean team, they suggest us setting metrics after `Glean.Initialize()`. Otherwise, those pings and operations will be deleted in some cases.

```
```

</details>

daoshengmu:
Fixes #3129 

Based on the recent Glean updates ( https://github.com/mozilla/glean/blob/master/CHANGELOG.md#v2300-2020-01-07).

GleanMetricsService.start() need to be called after Glean system initialization.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/telemetry/GleanMetricsService.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/telemetry/GleanMetricsService.java
@@ -46,13 +46,14 @@ public class GleanMetricsService {
         initialized = true;
 
         final boolean telemetryEnabled = SettingsStore.getInstance(aContext).isTelemetryEnabled();
+        Configuration config = new Configuration(Configuration.DEFAULT_TELEMETRY_ENDPOINT, BuildConfig.BUILD_TYPE);
+        Glean.INSTANCE.initialize(aContext, telemetryEnabled, config);
+
         if (telemetryEnabled) {
             GleanMetricsService.start();
         } else {
             GleanMetricsService.stop();
         }
-        Configuration config = new Configuration(Configuration.DEFAULT_TELEMETRY_ENDPOINT, BuildConfig.BUILD_TYPE);
-        Glean.INSTANCE.initialize(aContext, true, config);
     }
 
     // It would be called when users turn on/off the setting of telemetry.


daoshengmu:
@Dexterp37 Please help review this change if it makes in Glean. Thanks.

app/src/common/shared/org/mozilla/vrbrowser/telemetry/GleanMetricsService.java
```diff
@@ -46,13 +46,14 @@ public static void init(Context aContext) {
         initialized = true;
 
         final boolean telemetryEnabled = SettingsStore.getInstance(aContext).isTelemetryEnabled();
+        Configuration config = new Configuration(Configuration.DEFAULT_TELEMETRY_ENDPOINT, BuildConfig.BUILD_TYPE);
```
Dexterp37, line 49: @daoshengmu This change is not require nor useful: any metric set before glean initialized is batched and recorded after glean initializes, automatically (up to 100 batched actions). Why did you move this?

daoshengmu: Based on the conversation with @Travis in Matrix

`
It should still work, as Glean will cache up to 100 operations before you call initialize, but if there are several things going on at that time then it could be a factor in why you are seeing nulls.
If GleanMetricsService.start() performs some recording, then I would try and call initialize prior to that, if possible.
If telemetry is disabled, then Glean.initialize will do the right things in that case, so I'm also a little curious why you aren't passing your telemetryEnabled into Glean.initialize
`

Otherwise, we have no other choice can do, and our code freeze is on Apr. 10th.

Dexterp37: That conversation with @travis79 is correct, and it's easy to verify without making this change, since Glean reports [a metric about this problem](https://searchfox.org/glean/rev/ee9383ae6eca3821036d254d142eeeda7a585061/glean-core/metrics.yaml#307). The conversation also assumes you're on a recent version of Glean SDK, but you're not.

You should always try to update to the latest version of the Glean SDK to get the latest bugfixes and improvements. This is also required for things such as #3096.

Please note that the problem you're seeing is most probably not your product dropping data due to the pre-init queue overflowing, but [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1628057) that was recently fixed.

Moreover, the reason why you're seeing missing `distribution_name` field sometimes, is probably due to the fact that the application is crashing and not able to close properly. If you were on a more recent version of the SDK, then you could tell by examining the "ping reasons".

My tl;dr - please revert this change, as on the version you are on this might cause dataloss, update to a more recent version of glean as soon as it fits your schedule.

PR #2521, opened 2019-12-20
WebXR gamepad module support.

daoshengmu:
Fixes #2237 and #2095.

I have checked it works properly in my local but I am still waiting for WebXRSession and WebXRInputSource implementation from GeckoView. We might can merge it first? Or we need to wait for WebXR works in GV?

I will be away for a couple of weeks. If there are some WebXR progress from GV, maybe someone from our team can continue to work on this.

PR #3136, opened 2020-04-08
Implement WebXR Interstitial

### LINKED ISSUE #1966: help users figure out how to exit immersive mode
many users never figure out that they have to click the left-controller menu button to exit immersive mode. we should figure out how to teach that gesture. we should also explore ways we might consolidate all gestures on to a single controller.

Most likely way to do this is to have an interstitial screen as the user enters immersive mode. it could show the controller model with the correct button/gesture.

The interstitial should also include:
"[URL] is requesting information that can uniquely identify you, including your height and size of your room"
"You can control whether sites are allowed to enter immersive mode in Settings"

### LINKED ISSUE #3123: [Pico Neo2] The spinner from the Voice Search prompt is not animated
**[Affected Versions]:**
- Firefox Reality version 10-140982354 (bce60f9).

**[Affected Platforms]:**

**[Prerequisites]:**
- The Voice search permissions are granted.

**[Steps to reproduce]:**                                        
1. Click the Microphone button from the Awesomebar.
2. Perform a voice search and observe the spinner from the prompt.

**[Expected results]:**
- The spinner is animated.

**[Actual results]:**
- The spinner is not animated.

**[Additional Notes]:**
- Attached is a screen recording of the issue: [link](https://drive.google.com/open?id=12HcXb3WqgIj8Pyrgw-joxqKtE0rSNBK3).

MortimerGoro:
Fixes  #1966 Fixes #3123


MortimerGoro:
The spinner animation is not the final one, @keianhzo is working on the shapeshifter exported version

daoshengmu:
There is a build fail on this merge:

```
/opt/FirefoxReality/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WebXRInterstitialController.java:15: error: cannot find symbol
    private WebxrInterstitialControllerBinding mBinding;
```

MortimerGoro:
> There is a build fail on this merge:
> 
> ```
> /opt/FirefoxReality/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WebXRInterstitialController.java:15: error: cannot find symbol
>     private WebxrInterstitialControllerBinding mBinding;
> ```

Fixed

daoshengmu:
I notice the controller icons only be shown when entering the immersive mode at the 1st time. Is it what we want?

MortimerGoro:
For now it’s displayed for 3 seconds the first time you enter immersive (on each app restart.
I don’t like forcing it every EnterVR.

After trying it I think it would be good to wait for a user click to dismiss the how-to information, and remove the forced time. IMO:
- The first time you go to WebXR show the how-to-exit information. Dismiss after a click, not time based. We don’t know how long takes for a user to process the info and find the button.
- Do not show the spinner at the same time, it distracts from reading the how-to
- Show the spinner only in the next EnterVR for the whole session

Thoughts?

bluemarvin:
> For now it’s displayed for 3 seconds the first time you enter immersive (on each app restart)
> I don’t like forcing it every EnterVR.
> 
> After trying it I think it would be good to wait for a user click to dismiss the how-to information, and remove the forced time. IMO:
> 
>     * The first time you go to WebXR show the how-to-exit information. Dismiss after a click, not time based. We don’t know how long takes for a user to process the info and find the button.
> 
>     * Do not show the spinner at the same time, it distracts from reading the how-to
> 
>     * Show the spinner only in the next EnterVR for the whole session
> 
> 
> Thoughts?

I think this is a good approach. I think the controller images could be a little bigger maybe?

daoshengmu:
> For now it’s displayed for 3 seconds the first time you enter immersive (on each app restart)
> I don’t like forcing it every EnterVR.
> 
> After trying it I think it would be good to wait for a user click to dismiss the how-to information, and remove the forced time. IMO:
> 
> * The first time you go to WebXR show the how-to-exit information. Dismiss after a click, not time based. We don’t know how long takes for a user to process the info and find the button.
> * Do not show the spinner at the same time, it distracts from reading the how-to
> * Show the spinner only in the next EnterVR for the whole session
> 
> Thoughts?

I feel we can remove the spinner if we are showing the how-to-exit information, the spinner sometimes catch my attention than the how-to-exit. However, if users end the information icon, WebXR pages are still loading. We should still show the spinner instead.

Make the controller image bigger can let me see the words more easy.  I also feel if we already allow users to press a button to end the how-to-exit information, maybe longer time like 5 secs would be great for user have more time to understand it.

app/src/common/shared/org/mozilla/vrbrowser/utils/DeviceType.java
```diff
@@ -2,25 +2,48 @@
...
+        switch (aType) {
+            case OculusGo:
+                name = "Oculus Go";
+                break;
+            case OculusQuest:
+                name = "Oculus Quest";
+                break;
+            case ViveFocusPlus:
+                name = "Vive Focus Plus";
+                break;
```
bluemarvin, line 33: Do we need a `ViveFocus` also?

app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -1481,20 +1494,22 @@ BrowserWorld::DrawImmersive(device::Eye aEye) {
 }
 
 void
-BrowserWorld::TickLoadingAnimation() {
-  m.loadingAnimation->Update();
+BrowserWorld::TickWebXRInterstitial() {
+  m.rootWebXRInterstitial->SetTransform(m.device->GetReorientTransform());
   m.drawHandler = [=](device::Eye eye) {
-    DrawLoadingAnimation(eye);
+      DrawWebXRInterstitial(eye);
```
bluemarvin, line 1500: nit: Looks like there are two extra spaces added?

app/src/common/shared/org/mozilla/vrbrowser/utils/DeviceType.java
```diff
@@ -2,25 +2,48 @@
 
 import android.util.Log;
 
+import androidx.annotation.IntDef;
+
 import org.mozilla.vrbrowser.BuildConfig;
 
 public class DeviceType {
     // These values need to match those in Device.h
+    @IntDef(value = { Unknown, OculusGo, OculusQuest, ViveFocusPlus, PicoNeo2, PicoG2 })
```
bluemarvin, line 11: It looks like `Device.h` needs to be updated?

daoshengmu:
Do we have an alpha blending problem on the spinner image? I can see its rectangle background when we are going to end the animation.

MortimerGoro:
@daoshengmu @bluemarvin I implemented the flow we talked about. Let me know what you think. Feel free to change the new text

daoshengmu:
> @daoshengmu @bluemarvin I implemented the flow we talked about. Let me know what you think. Feel free to change the new text

It seems like pressing squeeze btns on Quest can't exit the how-to-exit page.

app/src/common/shared/org/mozilla/vrbrowser/utils/DeviceType.java
daoshengmu: We need ViveFocus as well

MortimerGoro:
> > @daoshengmu @bluemarvin I implemented the flow we talked about. Let me know what you think. Feel free to change the new text
> 
> It seems like pressing squeeze btns on Quest can't exit the how-to-exit page.

Fixed

MortimerGoro:
> Do we have an alpha blending problem on the spinner image? I can see its rectangle background when we are going to end the animation.

I noticed that too but it seems a driver bug. It only happens with the first layer displayed on a black background. If I change the order the alpha blending moves to the first layer rendered. I'll look into it next week, we can address that as a follow-up

app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -483,10 +483,6 @@ struct DeviceDelegateOculusVR::State {
         } else {
           VRB_WARN("Undefined hand type in DeviceDelegateOculusVR.");
         }
-
-        // This is always false in Oculus Browser.
-        const bool thumbRest = false;
-        controller->SetButtonState(controllerState.index, ControllerDelegate::BUTTON_OTHERS, 5, thumbRest, thumbRest);
       } else {
```
daoshengmu, line 486: It looks odd to me. This is thumbRest instead of the grip button, and thumbRest is always true. If `thumbRest` will cause problems to you, Neo2 will have it too because it also has `thumbRest`.

daoshengmu: You can remove this commit now. I already gave a fix at #2521 by defining a new btn state `BUTTON_SQUEEZE`. That makes sense for us to use it in app.

PR #3137, opened 2020-04-08
Bump GV and enable WebXR

MortimerGoro:


MortimerGoro:
Better to land this tomorrow, @daoshengmu's controller patches are landing in Gecko

PR #3164, opened 2020-04-15
Fix panel restore on restart.

bluemarvin:


keianhzo:
The panels stopped working after a session without firstcontentfulpaint is opened (download starts in new session)

bluemarvin:
> The panels stopped working after a session without firstcontentfulpaint is opened (download starts in new session)

It works for Pico and HTC, I forgot to handle the non SurfaceTexture case on the Quest.

bluemarvin:
@keianhzo I've updated the PR to work on Oculus also.

PR #3166, opened 2020-04-15
Handle content panels drawing for sessions without firstcontentfulpaint

keianhzo:
Handle content panels drawing for session without  `firstcontentfulpaint`. 
The second commit is a refactor  to avoid keeping track of the setviewcalls to fix the issues we had when restoring windows with library panels enabled. It introduces a little lag when a the panel is hidden as we take some time to get the `firstcontentfulpaint `call, but I think is manageable.


keianhzo:
Already addressed in another PR

PR #3174, opened 2020-04-16
l10n string updates

keianhzo:
l10n string updates. Related to https://github.com/mozilla-l10n/android-l10n/pull/192

### DIFF
--- a/app/src/main/res/values-da/strings.xml
+++ b/app/src/main/res/values-da/strings.xml
@@ -108,7 +108,7 @@
 
     <!-- This string is used as the title of the pop-ups dialog used to remove sites from the currently
          allowed pop-up sites. -->
-    <string name="settings_privacy_policy_popups_title_v1">Pop op-tilladelser</string>
+    <string name="settings_privacy_policy_popups_title">Pop op-tilladelser</string>
 
     <!-- This string is displayed under the title the pop-ups dialog used to remove sites from the currently
          allowed pop-up sites. -->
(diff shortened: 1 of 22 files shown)


keianhzo:
@flodolo please help review.

app/src/main/res/values-da/strings.xml
```diff
@@ -108,7 +108,7 @@
 
     <!-- This string is used as the title of the pop-ups dialog used to remove sites from the currently
          allowed pop-up sites. -->
-    <string name="settings_privacy_policy_popups_title">Pop op-tilladelser</string>
+    <string name="settings_privacy_policy_popups_title_v1">Pop op-tilladelser</string>
```
flodolo, line 111: As a general rule, you shouldn't modify localized files. But this is definitely wrong:
- You're adding a new string in en-US, `settings_privacy_policy_popups_title_v1` with text `Exceptions for pop-ups`. The old one was `settings_privacy_policy_popups_title`, `Pop-ups permissions`.
- By doing these changes, all locales will have this new string "pre-translated", but with the wrong reference (the old one, `Pop-ups permissions`.

Please revert all changes to the localized files, only add `settings_privacy_policy_popups_title_v1` to the English/source file as you already did.

keianhzo: Thanks, fixed.

keianhzo:
@flodolo I've updated the PR. I've added an issue to move html encoded entities to CDATA if that plays better with the localization flow: https://github.com/MozillaReality/FirefoxReality/issues/3185

keianhzo:
@bluemarvin This is ready to be landed.

Issue #2948, opened 2020-03-12
The volume slider from immersive mode remains displayed even after moving the pointer away from its active controller area


heres-maria:
**[Affected Versions]:**
- Firefox Reality version 10-40720037 (dbb8ed3a)
- Firefox Reality version 8.1 (a0fcc8a3)
- Firefox Reality version 8 (ad776185)

**[Affected Platforms]:**
- HTC Vive Focus Plus - Android 7.1.2
- Oculus Quest - Android 7.1.1
- Pico Neo 2 - Android - 8.1.0

**[Prerequisites]:**
- The headset is turned on, the browser is installed and open.

**[Steps to reproduce]:**                                        
1. Go to a 360 video (e.g. https://youtu.be/hNAbQYU0wpg).
2. Play it and enter immersive mode.
3. Hover the volume button.
4. Point the pointer away from the volume slider and observe the behavior.

**[Expected results]:**
- The volume slider is dismissed.

**[Actual results]:**
- The volume slider remains displayed. 

**[Notes]:**
- The slider remains displayed across different videos.
- The issue is also reproducible for the other types of videos (180, normal, etc) when in immersive mode.
- Moving the pointer over the actionable area several times sometimes dismisses the slider.
- This issue is reproducible in production.
- Design [document](https://trello-attachments.s3.amazonaws.com/5a260c5d480978f87ffad6aa/5bd0ec2f89896a62d7978eca/ff3e0803782a1d13ccc3327d0bdcea33/UF-32_Volume_Controls_in_Transport_Controls_001.pdf) used as reference.
- Attached is a [recording](https://drive.google.com/file/d/1q8G19DHwP1M9byWGNZf3M9eh27O-gygR/view?usp=sharing) of this issue.

keianhzo:
asdsadasdasdsadasd
asdsadads

daron-walters:
[Link To Fix Video](https://drive.google.com/open?id=1uuoN0MKQSaLQl7GCkkwi3AEl91aisW2e[](url))

daron-walters:
here is a [video](https://drive.google.com/open?id=18eSqabJA-Agde0bfFpdFIVxgdK4SnJwz) of the fix

Softvision-GeluHaiduc:
I have verified this issue with the latest Firefox Reality version 10 RC1 (6db2f60a) installed on the Oculus Go, HTC Vive Focus Plus, Pico Neo2 and Pico G2 4K headsets.
[...]

PR #3181, opened 2020-04-16
The volume slider from immersive mode remains displayed even after mo…

daron-walters:
…ving the pointer away - #2948

1. updated code to use the x location of the volume button as the starting point for hover to show/hide slider.  Per design doc the slider should show from the beginning of the volume button to the end of the slider

2. added code to handle hide the slider when the x location is outside of the media controls.  On oculus quest the x coord is always negetive when pointer is outside control area

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/MediaControlsWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/MediaControlsWidget.java
@@ -8,12 +8,14 @@ package org.mozilla.vrbrowser.ui.widgets;
 import android.content.Context;
 import android.graphics.Rect;
 import android.graphics.drawable.Drawable;
+import android.os.Handler;
 import android.util.AttributeSet;
 import android.view.MotionEvent;
 import android.view.View;
 import android.widget.FrameLayout;
 import android.widget.TextView;
 
+import org.mozilla.telemetry.schedule.jobscheduler.TelemetryJobService;
 import org.mozilla.vrbrowser.R;
 import org.mozilla.geckoview.MediaElement;
 import org.mozilla.vrbrowser.browser.Media;
@@ -42,6 +44,10 @@ public class MediaControlsWidget extends UIWidget implements MediaElement.Delega
     private boolean mPlayOnSeekEnd;
     private Rect mOffsetViewBounds;
     private VideoProjectionMenuWidget mProjectionMenu;
+    static long VOLUME_SLIDER_CHECK_DELAY = 1000;
+    private Handler mVolumeCtrlHandler = new Handler();
+    private boolean mHideVolumeSlider = false;
+    private Runnable mVolumeCtrlRunnable;
 
     public MediaControlsWidget(Context aContext) {
         super(aContext);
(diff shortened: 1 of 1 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/MediaControlsWidget.java
```diff
@@ -197,8 +197,13 @@ public void onSeekPreview(String aText, double aRatio) {
             if (mMedia == null) {
                 return false;
             }
+            if(event.getX()<0)
+            {
+                mVolumeControl.setVisibility(View.GONE);
+                return false;
+            }
```
bluemarvin, line 204: A few formatting nits:
```
if(event.getX()<0)
{
}
```
Should be:
```
if (event.getX() < 0) {
}
```
Space after `if`. Space on each side of `<`.  Curly brackets `{}` should be K&R style.

bluemarvin:
@keianhzo I sort of wonder if the better solution is to show the volume bar when the speaker button is hovered and then start a timer after the pointer leaves the volume button that hides the volume slider if the pointer does not enter the volume slider after a certain time. If the pointer does enter the volume slider, the timer should be reset and started again once the pointer leaves the slider. Thoughts?

keianhzo:
@bluemarvin @daron-walters That's a good option. 
Other options:
- We show the volume slider when the pointer leaves the button through the right side of the volume button. In any other case we hide it. We can do this handling the threshold both the widget hover listener and in the button hover listener.
- Or we just always show the volume bar, I don't see any good reason (other than the spec) why it shouldn't be always visible and we save the user all this hovering thing.

daron-walters:
All three are good options, I will default to you guys on the final decision.  Just let me know and I can start on implementation

keianhzo:
@daron-walters I think @bluemarvin idea is probably the most simple-effective one, I'd go ahead with that one.

daron-walters:
@keianhzo will do

keianhzo:
@daron-walters when exiting the volume button through the bottom side, the volume is not hidden. 

I think we could do something like:
- Show the volume slider when the hover enters the volume button 
- Schedule a hide whenever the hover exits the volume button or the volume slider
- Cancel the scheduled hide when the hover enters the volume slider.

I think that would simplify the logic and we can remove the widget level hover and the threshold related code.

daron-walters:
@keianhzo I made the change where it is handled by the buttons and not the widget as a whole.  I will reach out tomorrow to make sure I full understand all scenarios before doing another pull request

keianhzo:
@daron-walters Looks good, I think the only thing missing is to cancel the hide task when we enter the volume slider otherwise it gets dismissed while you are trying to change the volume. Squash the commit before pushing again so we can more clearly see the commit changes.

daron-walters:
never did a squash on rebase before until now so let me know if it is correct.  Added the change Manuel informed me of as far as turning off timer when hover enters volume slider

bluemarvin:
You should should only have one commit remaining after you squash your branch. Typically you would do this by running `git rebase -i origin/master` and then flattening all the commits but the first. But for some reason what ever you've done to your branch, it causes conflicts with it self when trying to flatten. I'm trying to figure out how to untangle it.

bluemarvin:
I've flattened your commit and pushed it to this branch: https://github.com/MozillaReality/FirefoxReality/tree/v10/volume_slider_immersive_mode_flattened
Take a look. If it appears correct you can push it to this branch as follows:
`git push -f origin v10/volume_slider_immersive_mode_flattened:v10/volume_slider_immersive_mode`
Let me know if you have any questions.

bluemarvin:
@daron-walters You unsquashed your commit. You need to start with the flattened branch that I created and cherry-pick your new commit on top of the flattened branch locally. Then push that to GitHub.

daron-walters:
ok will do

PR #3223, opened 2020-04-23
Support for opening elements in new tabs

keianhzo:
Until now we only supported opening links in new windows/tabs. This PR adds support for also opening elements. ie. you want to open an image element (without a link) in a new window/tab


bluemarvin:
@keianhzo This needs a rebase to land.

keianhzo:
@bluemarvin Rebase issues fixed. I've updated the PR to work more similar to what desktop does and show "View image/video/audio" instead of opening the src in new tab/windows.

PR #3225, opened 2020-04-23
Replace ellipisis unicode for string

keianhzo:
Fixes the build error:
```
line 1:87 token recognition error at: ';'
line 1:88 token recognition error at: '''
```


bluemarvin:
Looks like taskcluster failed from github being down. Re-ran task and it passed.

PR #3232, opened 2020-04-23
Add support for lay-down mode in VR videos

MortimerGoro:


bluemarvin:
I think when exiting immersive mode, the heading adjustment code removes the pitch that was set before entering immersive mode.

PR #3237, opened 2020-04-23
Fixes #3229 Avoid restoring/clearing exceptions list when inserting

### LINKED ISSUE #3229: Tracking protection is wrongly enabled in Private Browsing when disabling/enabling it in normal browsing from the pop-up
**[Affected Versions]:**
- Firefox Reality version 10 RC1 (6db2f60a)

**[Affected Platforms]:**
- Pico Neo2 - Android 8.1.0

**[Prerequisites]:**

**[Steps to reproduce]:**
1. Open a Private Browsing session and navigate to a website.
2. Click the TP shield icon an disable Tracking Protection.
3. Exit Private Browsing and open a website in Normal Browsing.
4. Click the TP shield icon an disable Tracking Protection.
5. Enter in Private Browsing again and observe the Tracking Protection shield icon.

**[Expected results]:**
- The icon is strikethrough, the Tracking Protection is disabled.

**[Actual results]:**
- The icon is not strikethrough, the Tracking Protection is enabled.

**[Notes]:**
- Attached a screen recording of the issue: [link](https://drive.google.com/open?id=1IKTt3fkrrOnQyLjY4wAshLc3SBUQZsuT).

keianhzo:
Fixes #3229 Avoid restoring/clearing exceptions list when inserting

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/content/TrackingProtectionStore.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/content/TrackingProtectionStore.java
@@ -4,7 +4,6 @@ import android.app.Application;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.preference.PreferenceManager;
-import android.util.Log;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
@@ -12,8 +11,6 @@ import androidx.lifecycle.DefaultLifecycleObserver;
 import androidx.lifecycle.Lifecycle;
 import androidx.lifecycle.LifecycleOwner;
 import androidx.lifecycle.Observer;
-import androidx.recyclerview.widget.DiffUtil;
-import androidx.recyclerview.widget.ListUpdateCallback;
 
 import org.json.JSONException;
 import org.json.JSONObject;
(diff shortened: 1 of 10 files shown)


app/src/common/shared/org/mozilla/vrbrowser/browser/content/TrackingProtectionStore.java
```diff
@@ -112,18 +131,10 @@ public void onInserted(int position, int count) {
                 mListeners.forEach(listener -> listener.onExcludedTrackingProtectionChange(
                         UrlUtils.getHost(exception.uri),
                         true));
+                List<Session> sessions = SessionStore.get().getSessionsByHost(UrlUtils.getHost(permission.url));
+                sessions.forEach(session -> mContentBlockingController.addException(session.getGeckoSession()));
```
bluemarvin, line 135: Why is this necessary? If we are inserting it into the model, hasn't it already been added by a session?

keianhzo: Yes, that was unnecessary. Fixed

keianhzo:
@bluemarvin I've updated the PR to handle the exceptions persistence in a more straightforward way. 

The whole point of using the lists diffs was being able to reuse the `SitePermissionsOptionsView `to handle all exceptions in the same way but it was making the tracking exceptions handling code more complicated than it should. I've added a specific `TrackingPermissionsOptionsView` so we always update the tracking exceptions using the same methods and update the storage afterwards instead of needing to listen to the storage updates events. I think this makes the code simpler and easier to understand.

PR #3253, opened 2020-04-27
Do not display popup icons when a popup is launched on page leave

### LINKED ISSUE #3250: The "Pop-UP Blocked" notification and icon are wrongly redisplayed on the next loaded website
**[Affected Versions]:**
- Firefox Reality version 10 RC2 (17a313b).

**[Affected Platforms]:**
- HTC Vive Plus - Android 7.1.2

**[Prerequisites]:**

**[Steps to reproduce]:**
1. Navigate to a website with pop-ups ex: http://www.popuptest.com/popuptest1.html.
2. Focus the Awesomebar and navigate to a website with no pop-ups.
3. Observe if the "Pop-UP Blocked" notification and icon are displayed.

**[Expected results]:**
- They are not displayed.

**[Actual results]:**
- They are displayed even if the website doesn't has pop-ups.

**[Additional Notes]:**
- This issue is also reproducible if navigating Back and Forward using the navigation buttons.
- If one more website is navigated the "Pop-UP Blocked" notification and icon are no longer displayed.
- Attached is a screen recording of the issue: [link](https://drive.google.com/open?id=1MLjuOQBK09DgSlKq_oWXqRRp5VWQc6sj).

MortimerGoro:
Fixes  #3250

I checked we match the Firefox Desktop behavior:

- Open http://popuptest.com/popuptest2.html
- Navigate to other place. Popup website triggers another popup on page leave.
- The popup is rejected and the notification/icon are not visible in the awesome bar

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/PromptDelegate.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/PromptDelegate.java
@@ -251,16 +251,11 @@ public class PromptDelegate implements
                 final String uri = UrlUtils.getHost(session.getCurrentUri());
                 SitePermission site = mAllowedPopUpSites.stream().filter((item) -> item.url.equals(uri)).findFirst().orElse(null);
                 if (site != null) {
-                    mAttachedWindow.postDelayed(() -> {
-                        result.complete(popupPrompt.confirm(AllowOrDeny.ALLOW));
-                        session.setPopUpState(SessionState.POPUP_ALLOWED);
-                    }, 500);
-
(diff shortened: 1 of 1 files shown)


keianhzo:
@MortimerGoro It still happens whenever you navigate to another page in the same host probably because we are delaying the state updates. 
We can probably remove the `postDelayed `code from here:
https://github.com/MozillaReality/FirefoxReality/blob/12726e0796080855a2b632357a29f7ff39c72dca/app/src/common/shared/org/mozilla/vrbrowser/browser/PromptDelegate.java#L254-L263
and that would fix it and we wouldn't need the host check.

PR #3262, opened 2020-04-28
Checking external storage permission before loading files.

### LINKED ISSUE #3257: The downloaded file is still opened in a new tab if the app's permissions to access the external storage is denied
**[Affected Versions]:**
- Firefox Reality version 10 RC2 (17a313b).

**[Affected Platforms]:**

**[Prerequisites]:**
- Have an available file in the Downloads library.
- The Permissions to access External Storage is disabled.

**[Steps to reproduce]:**
1. Click the Downloads button from the tray.
2. Hover over a file and click the 3 dot menu button.
3. Click the "Open in a Tab" button.
4. Deny the permissions from the prompt displayed.
5. Observe the behavior.

**[Expected results]:**
- The file is not opened.

**[Actual results]:**
- The file is opened even if the access was denied.

**[Additional Notes]:**
- If the file is opened New Window using the 3 dot menu, the file is not displayed when denning permissions to access the external storage, but a new window is still created and the placeholder is displayed.
- On the HTC devices the "File location not found" page is displayed in the new tab when opening the file in a new tab.
- Attached is a screen recording of the issue: [link](https://drive.google.com/open?id=1bWoVba4OMDpIdvuhrX7J0lzOhKGpsGJI).

daoshengmu:
Fixes #3257

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/Session.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/Session.java
@@ -8,7 +8,9 @@ package org.mozilla.vrbrowser.browser.engine;
 import android.app.Activity;
 import android.content.Context;
 import android.content.SharedPreferences;
+import android.content.pm.PackageManager;
 import android.graphics.Bitmap;
+import android.net.Uri;
 import android.preference.PreferenceManager;
 import android.util.Log;
 import android.view.Surface;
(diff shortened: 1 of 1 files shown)


keianhzo:
@daoshengmu I was looking into this and it seems that the root cause for this is that the session restore happens before the session listeners are set in the session. That makes the `LoadRequest` from `WindowWidget` listener (and any other listeners) not to be called when the `loadUri `from restore is executed.

Although this fixes the issue I think it would be better to fix that issue. I've opened a PR for that: https://github.com/MozillaReality/FirefoxReality/pull/3265

bluemarvin:
Should be fixed by #3265

PR #3279, opened 2020-04-28
Ensure mUri is not null in Window save

bluemarvin:
Speculative fix for https://crash-stats.mozilla.org/report/index/6b13ce7f-7f54-48f3-acf4-ebb540200428


bluemarvin:
> Seeing at the crash, looks like it must be that. Can we make it in a single line? like:
> 
> ```
> .filter(sessionState -> SAVE_BLACKLIST.stream().noneMatch(uri ->
>                             sessionState.mUri == null || sessionState.mUri.startsWith(uri)))
> ```

Shouldn't it be `sessionState.mUri != null && sessionState.mUri.startsWith(uri)))`

PR #3297, opened 2020-04-30
Closes #3277 Improved setSession code

### LINKED ISSUE #3277: Make setupListener calls safer
See: 
https://github.com/MozillaReality/FirefoxReality/pull/3265#pullrequestreview-402123804

keianhzo:
Closes #3277 Refactor of the setSession code to set the previous session active state and setup the listeners.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WindowWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WindowWidget.java
@@ -94,6 +94,10 @@ public class WindowWidget extends UIWidget implements SessionChangeListener,
     public static final int SESSION_RELEASE_DISPLAY = 0;
     public static final int SESSION_DO_NOT_RELEASE_DISPLAY = 1;
 
+    @IntDef(value = { DEACTIVATE_CURRENT_SESSION, LEAVE_CURRENT_SESSION_ACTIVE})
+    public @interface SetSessionActiveState {}
+    public static final int DEACTIVATE_CURRENT_SESSION = 0;
+    public static final int LEAVE_CURRENT_SESSION_ACTIVE = 1;
 
     private Surface mSurface;
     private int mWidth;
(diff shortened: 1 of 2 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WindowWidget.java
```diff
@@ -1095,21 +1100,35 @@ public void draw(Canvas aCanvas) {
...
+                }
                 if (aDisplayAction == SESSION_RELEASE_DISPLAY) {
                     oldSession.releaseDisplay();
                 }
             }
 
             mSession = aSession;
+
+            setupListeners(mSession);
+            SessionStore.get().setActiveSession(mSession);
```
MortimerGoro, line 1130: SessionStore active session is set here regardless if the session itself is set to DEACTIVE. Is this correct?

keianhzo: The **setSession** target session is always activated, what the **aSessionState** parameter determines is what we do with the old session.

keianhzo:
@bluemarvin I've rebased and I had to resolve a conflict affecting your latest patch https://github.com/MozillaReality/FirefoxReality/pull/3350. I've tested and seems to be good but maybe you want to take a look.

app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WindowWidget.java
```diff
@@ -94,6 +94,11 @@
     public static final int SESSION_RELEASE_DISPLAY = 0;
     public static final int SESSION_DO_NOT_RELEASE_DISPLAY = 1;
 
+    @IntDef(value = { ACTIVATE, DEACTIVATE, DO_NOT_CHANGE})
+    public @interface SetSessionActiveState {}
+    public static final int ACTIVATE = 0;
```
bluemarvin, line 99: I don't see `ACTIVATE` used anywhere and I'm not sure I understand the use case.

bluemarvin: I guess I find it confusing because the flag is actually for the current session that becomes the old session and not the session being set. Maybe the flags should be `DEACTIVATE_CURRENT_SESSION` and `LEAVE_CURRENT_SESSION_ACTIVE`?

keianhzo: I left that just in case we needed it in the future but I've removed it and now we just pass a flag, I guess if we need it sometime we can go back to the IntDef.

app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WindowWidget.java
```diff
@@ -1107,21 +1106,28 @@ public void draw(Canvas aCanvas) {
...
     }
 
-    public void setSession(@NonNull Session aSession) {
-        setSession(aSession, SESSION_RELEASE_DISPLAY);
+    public void setSession(@NonNull Session aSession, boolean aDeactivateCurrentSession) {
+        setSession(aSession, SESSION_RELEASE_DISPLAY, aDeactivateCurrentSession);
     }
 
-    public void setSession(@NonNull Session aSession, @OldSessionDisplayAction int aDisplayAction) {
+    public void setSession(@NonNull Session aSession, @OldSessionDisplayAction int aDisplayAction, boolean aDeactivateCurrentSession) {
```
bluemarvin, line 1113: I would prefer something other than a boolean be used for `aDeactivateCurrentSession` since it makes the call sites harder to read.

keianhzo: Reverted to use an IntDef

app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/Windows.java
```diff
@@ -1342,10 +1326,7 @@ public void onTabsReceived(@NonNull List<TabData> aTabs) {
 
             if (i == 0 && !fullscreen) {
                 // Set the first received tab of the list the current one.
-                SessionStore.get().setActiveSession(session);
-                targetWindow.setupListeners(session);
-                targetWindow.getSession().setActive(false);
-                targetWindow.setSession(session);
+                targetWindow.setSession(session, WindowWidget.LEAVE_CURRENT_SESSION_ACTIVE);
```
bluemarvin, line 1329: The removed code seems to be `DEACTIVATE_CURRENT_SESSION` and not `LEAVE_CURRENT_SESSION_ACTIVE`. Am I reading it wrong?

keianhzo: Right, I've updated.

PR #3308, opened 2020-05-01
Adjust Oculus Go controller transform based on WebXR input profile.

daoshengmu:
We need to adjust it as we did for Quest and Focus Plus. Luckily, for Focus controllers, we don't need it.

It needs to wait for [Bug 1634808](https://bugzilla.mozilla.org/show_bug.cgi?id=1634808) be merged to GV first.


daoshengmu:
Bug 1634808 is landed. We can start to review now.

PR #3311, opened 2020-05-02
Update generated user.js to use user_pref

### LINKED ISSUE #3310: Unable to disable WebRender starting in GV 77.0.20200430082621
## Configuration

Firefox Reality version: 10
Firefox Reality build ID: 

Hardware: All

## Steps to Reproduce

1. Launch FxR

## Current Behavior

WebRender is enabled and windows are rendered incorrectly.

## Expected Behavior

WebRender should be disabled.

## Possible Solution
This is a regression in Gecko see: https://bugzilla.mozilla.org/show_bug.cgi?id=1634814

bluemarvin:
Fixes #3310
Starting with GeckoView Nightly 77.0.20200430082621 user.js
needs to use user_pref() instead of pref() to set prefs.


bluemarvin:
We will need this PR if we are going to update the GV version in v10.

PR #3334, opened 2020-05-05
Fix v2 to v3 Database migration

### LINKED ISSUE #3329: The pop-up exceptions list is not saved when updating FxR from V9 to V10
**[Affected Versions]:**
- Firefox Reality version 10 RC4 (29991298)

**[Affected Platforms]:**
- Pico G2 4K - Android 8.1.0

**[Prerequisites]:**
- The production version of the app is installed. (FxR V9)
- Make sure to have at least one entry in the pop-up permissions list in FxR V9.

**[Steps to reproduce]:**
1. Update the app to the latest RC, FxR V10 RC4.
2. Open the "Privacy & Security" card from Settings and click the “Advanced” button from the "Block Pop-up Windows" setting.
3. Observe the list.

**[Expected results]:**
- The list is persisted across an update.

**[Actual results]:**
- The list is not persisted across an update.

**[Notes]:**
- The issue is not reproducible while updating from V10 RC1 to RC4.
- Attached a screenshot of the issue.
Tue05-05-2020_14-33-20 95

MortimerGoro:
Fixes #3329

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/db/AppDatabase.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/db/AppDatabase.java
@@ -38,7 +38,7 @@ public abstract class AppDatabase extends RoomDatabase {
     @NonNull
     private static AppDatabase buildDatabase(final @NonNull Context appContext, final @NonNull AppExecutors executors) {
         return Room.databaseBuilder(appContext, AppDatabase.class, DATABASE_NAME)
-                .addMigrations(MIGRATION_1_2)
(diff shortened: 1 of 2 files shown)


bluemarvin:
This didn't work for me. FxR crashes when trying to run v10 for the first time after using v9 to create a popup exception.

MortimerGoro:
Ready for testing again. There was a conflict between different types and the `allowed` column. Unfortunately SQLite doesn't support column remove without the mess of deleting a recreating the whole table, so it's easier to keep it. We can rename it in the future if we add a boolean field

PR #3342, opened 2020-05-06
Fix XRRigidTransform orientation is inverted.

daoshengmu:
Bump to GV 78.0.20200506093716

Test case: https://playcanv.as/e/p/TAYVQgU2/.
The controllers were inverted in comparison with Oculus Browser. This is because our Gecko [Bug 1635363](https://bugzilla.mozilla.org/show_bug.cgi?id=1635363) when setting a quaternion from a rotation matrix.

### DIFF
--- a/versions.gradle
+++ b/versions.gradle
@@ -24,7 +24,7 @@ ext.deps = [:]
 def versions = [:]
 // GeckoView versions can be found here:
 // https://maven.mozilla.org/?prefix=maven2/org/mozilla/geckoview/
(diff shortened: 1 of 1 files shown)


bluemarvin:
I think tomorrows GV will have the DRM fix so maybe we should wait to bump for tomorrow GV?

daoshengmu:
yep. It makes to me.

bluemarvin:
Fixed by #3347

PR #3344, opened 2020-05-07
Update to A-C 39.0.1

grigoryk:
This appears to work well, I'm still testing it out (but, only in an emulator since I don't have a VR device handy). 

cc @pocmo

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/Accounts.kt
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/Accounts.kt
@@ -296,7 +296,7 @@ class Accounts constructor(val context: Context) {
 
     fun pollForEventsAsync(): CompletableFuture<Boolean?>? {
         return CoroutineScope(Dispatchers.Main).future {
-            services.accountManager.authenticatedAccount()?.deviceConstellation()?.pollForEventsAsync()?.await()
+            services.accountManager.authenticatedAccount()?.deviceConstellation()?.pollForCommandsAsync()?.await()
         }
     }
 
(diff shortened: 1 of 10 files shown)


pocmo:
Thank you, Grisha! I'll give it a try on my Quest too. :)

pocmo:
Did some tests and I was able to login, sync and send tabs to other devices. However if I sent tabs to the device itself then they never showed up. But I am not sure whether that is caused by the version bump.

pocmo:
I switched back and forth between `master` and a branch with your patch applied. Only on `master` I was able to get a tab sent from my desktop to show up after manually synchronizing. The tabs I've sent previously (with the patched version) do not suddenly show up though - so I wonder if they have been received but dropped on the floor?

pocmo:
Looking at the log we receive one device event for the synced tab. But the problem seems to be around here:

```
val filteredEvents = events.filterIsInstance(DeviceCommandIncoming.TabReceived::class.java)
```

This instance check is never true. We receive an `AccountEvent.DeviceCommandIncoming`. But this is not the same as the `DeviceCommandIncoming` class `TabReceived` is defined on. 🤔 

https://github.com/mozilla-mobile/android-components/blob/b1f8371fcdf7a4b1501c31b7753ca263088caea2/components/concept/sync/src/main/java/mozilla/components/concept/sync/AccountEvent.kt#L38

vs.

https://github.com/mozilla-mobile/android-components/blob/b1f8371fcdf7a4b1501c31b7753ca263088caea2/components/concept/sync/src/main/java/mozilla/components/concept/sync/AccountEvent.kt#L22

pocmo:
It is fixed after unpacking the event first:

```Kotlin
val filteredEvents = events
    .filterIsInstance(AccountEvent.DeviceCommandIncoming::class.java)
    .map { it.command }
    .filterIsInstance(DeviceCommandIncoming.TabReceived::class.java)
```

keianhzo:
There is a Draft PR with the Glean and AC updates waiting to be landed when we are done with the current release: https://github.com/MozillaReality/FirefoxReality/pull/3096

I might compare and update that one with whatever we have missed from this one.

grigoryk:
Ah, thanks for the catch @pocmo !

@keianhzo PR in #3096 looks good to me - well, it should go to a higher A-C version now.

keianhzo:
Thanks for the review @grigoryk. I've updated AC in the other PR, I guess we can close this one.

grigoryk:
Yup!

PR #3350, opened 2020-05-08
Prevent listeners from getting added more than once

bluemarvin:
STR:

1) Visit [https://yts.mx/movie/theeb-2014](https://yts.mx/movie/theeb-2014)
2) Click on page to spawn ad
3) Close ad with tab view and select https://yts.mx/movie/theeb-2014 tab
4) Reload page and click on page to spawn new add

Actual:
Spawned ad tab fails to load

Expected:
Spawned ad tab loads

Cause:
The Window.java `SessionChangeListener` was getting added multiple times so when a new tab was added it got called twice and on the second call, it was disabling the new session since it had already been set.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/Session.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/Session.java
@@ -286,7 +286,7 @@ public class Session implements ContentBlocking.Delegate, GeckoSession.Navigatio
     }
 
     public void addNavigationListener(GeckoSession.NavigationDelegate aListener) {
-        mNavigationListeners.add(aListener);
+        mNavigationListeners.addIfAbsent(aListener);
         dumpState(aListener);
     }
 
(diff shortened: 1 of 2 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WindowWidget.java
```diff
@@ -1158,12 +1158,18 @@ public void onCurrentSessionChange(GeckoSession aOldSession, GeckoSession aSessi
...
+            return;
+        }
         // e.g. tab opened via window.open()
         aSession.updateLastUse();
         Session current = mSession;
         setupListeners(aSession);
         setSession(aSession);
         SessionStore.get().setActiveSession(aSession);
+        aSession.setActive(true);
+        //current.setActive(false);
```
keianhzo, line 1172: Do we need to keep this?

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/Session.java
```diff
@@ -286,7 +286,9 @@ public void setHistoryDelegate(GeckoSession.HistoryDelegate aDelegate) {
     }
 
     public void addNavigationListener(GeckoSession.NavigationDelegate aListener) {
-        mNavigationListeners.add(aListener);
+        if (!mNavigationListeners.contains(aListener)) {
+            mNavigationListeners.add(aListener);
+        }
```
keianhzo, line 291: We could avoid this using `addIfAbsent`.

PR #3352, opened 2020-05-08
Check for available apps when opening a file

keianhzo:
Check for available apps when opening a file otherwise we just show that the file is unsupported. It's not easy to test because of the lack of utilities in most platforms at the moment. It can be tested with:
- 3GP, TIFF in Oculus
- 3GP in Pico

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WindowWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/WindowWidget.java
@@ -9,6 +9,7 @@ import android.content.ActivityNotFoundException;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.content.Intent;
+import android.content.pm.PackageManager;
 import android.content.res.Configuration;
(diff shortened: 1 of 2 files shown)


bluemarvin:
Looks like this depends on https://github.com/MozillaReality/FirefoxReality/pull/3297 and needs to be rebased?

bluemarvin:
I wasn't able to ever get the built in app to play any of the video I downloaded on the Pico G2. Additionally, all video's I downloaded from https://www.appsloveworld.com/download-sample-3gp-video-file-for-testing/ failed to download yet the download works on desktop. Downloaded videos that didn't play were from https://filesamples.com/formats/3gp

keianhzo:
@bluemarvin Fixed rebase issues. 

Regarding the system app playback errors, I think there not much we can do about it other than never allowing to open with an external app.

The download failure, the Android DownloadManager fails with `Stop requested with status TOO_MANY_REDIRECTS: Too many redirects` but I couldn't find much info about it, might be system limitation.

Issue #3348, opened 2020-05-07
Remove `dom.vr.require-gesture: false` on release


fernandojsg:
## Configuration

Firefox Reality version: 10 (master)
Firefox Reality build ID: ea7a9e07

Hardware: Oculus Quest

## Steps to Reproduce

1. Visit https://autoenter-xr.glitch.me
2. It will "auto-enter" XR

## Current Behavior
It successfully trigger the XR interstitial to enter XR without requiring user gesture

## Expected Behavior
You should not be able to start a XR session without triggering an user gesture (Unless you have the `dom.vr.require-gesture` set to false.

## Possible Solution
Set `dom.vr.require-gesture` set to true by default on releases.

bluemarvin:
Something must have changed in Gecko because we only set it if it is in the extras of the intent.

fernandojsg:
I believe @MortimerGoro could have more info on that

bluemarvin:
We can't ship with this enabled, I'll look into today.

PR #3353, opened 2020-05-08
Only allow prefs to be set by extras in debug builds

bluemarvin:


bluemarvin:
Fixes #3348

fernandojsg:
I agree on the point of disabling all the prefs by default letting the user to set them if they want, but why can you also set them when running the app through an adb command, it could be useful to have access to that or at least having just a pref that you could accept other prefs by command line?

bluemarvin:
> I agree on the point of disabling all the prefs by default letting the user to set them if they want, but why can you also set them when running the app through an adb command, it could be useful to have access to that or at least having just a pref that you could accept other prefs by command line?

We need a quick fix for v10. I filed this as follow up https://github.com/MozillaReality/FirefoxReality/issues/3354

fernandojsg:
OK, it makes sense, thanks!

PR #3364, opened 2020-05-12
Fixes #2883 Ensure webexts are loaded when restoring sessions

### LINKED ISSUE #2883: The YouTube layout changes after closing and reopening/restarting the app
**[Affected Versions]:**
- Firefox Reality version 9 RC4 (c64365e3)

**[Affected Platforms]:**
- Pico Neo 2 - Android 8.1.0

**[Prerequisites]:**

**[Steps to reproduce]:**
1. Go to YouTube.
2. Change a setting that causes a restart prompt (e.g. Display -> DPI).
3. Choose the “Restart Now” option.
4. Observe the YouTube layout.

**[Expected results]:**
- The layout remains the same after restarting.

**[Actual results]:**
- The layout completely changes after restarting.

[Notes]:
- Trying to use the website in the new layout will return an “Invalid response received” error.
- The issue is also reproducible when closing the app and subsequently using the “force-stop” adb command before reopening it.
- Reloading the page fixes this issue.
- Related to #2557.
- Attached is a [recording ](https://drive.google.com/file/d/1Dqe4J0rI2c-w6bDo-Lbtc-Ms38J1z950/view?usp=sharing)of the issue.

keianhzo:
Fixes #2883 Web extensions are not being loaded for restored sessions after a restart.

STRs:
- Open a youtube video
- Change the MSAA setting and restart
- See the video URL (it has `app=desktop` appended)
- Change the MSAA setting and restart again

**Actual result**: The loaded url after the first restart has `app=desktop` appended, the webextension hasn't been applied, after the second restart Youtube has an incorrect layout.

**Expected result**: The correct Youtube layout should be used for restored sessions.

This would happen with any extension.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
@@ -93,9 +93,11 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.Consumer;
+import java.util.function.Function;
 
 import static org.mozilla.vrbrowser.ui.widgets.UIWidget.REMOVE_WIDGET;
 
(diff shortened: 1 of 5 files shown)


app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
```diff
@@ -392,6 +394,19 @@ public void onWindowVideoAvailabilityChanged(@NonNull WindowWidget aWindow) {
...
+        EngineProvider.INSTANCE.loadExtensions()
+                .thenAcceptAsync(aVoid -> {
+                    Log.d(LOGTAG, "WebExtensions loaded");
+                    mWindows.restoreSessions();
+                }, getServicesProvider().getExecutors().mainThread())
+                .exceptionally(throwable -> {
+                    String msg = throwable.getLocalizedMessage();
+                    if (msg != null) {
+                        Log.e(LOGTAG, "Extensions load error: " + msg);
+                    }
```
bluemarvin, line 407: Do we not want to try and restore sessions even if there is an error loading the extensions.

keianhzo: I've updated to restore sessions even if the extensions load fails. I guess it would be nice to also notify the user, otherwise we might be loading sessions which won't work properly due the the lack of extensions. I'll open a follow up, we might need a less intrusive notification system than an alert for these cases, ideally a window bottom-up timed sliding notification.

keianhzo:
Opened https://github.com/MozillaReality/FirefoxReality/issues/3379 add a better way of notifying these type of events.

PR #3376, opened 2020-05-14
Prevent duplicated setActive(false) calls

### LINKED ISSUE #3375: IInactive window and nfinite WebXR spinner after showing the Oculus menu in multiwindow
STR:
- Open https://webvr.info/samples/03-vr-presentation.html in two different windows
- Enter VR in one of them
- Show the Oculus menu and press continue
- The other vr sample is not resumed, if you click enter VR you get a infinite spinner

MortimerGoro:
It seems that calling setActiva(false) twice causes some issues. When we enter immersive mode we call setActive(false) on the non focused windows and we call surfaceDestroyed. If we show the Oculus menu later, setActive(false) is called again

Fixes #3375

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/Session.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/Session.java
@@ -767,6 +767,12 @@ public class Session implements ContentBlocking.Delegate, GeckoSession.Navigatio
     }
 
     public void setActive(boolean aActive) {
+        if (!aActive && mState.mSession != null && !mState.isActive()) {
+            // Prevent duplicated setActive(false) calls. There is a GV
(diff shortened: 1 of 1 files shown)


app/src/common/shared/org/mozilla/vrbrowser/browser/engine/Session.java
```diff
@@ -767,6 +767,9 @@ public void goForward() {
     }
 
     public void setActive(boolean aActive) {
+        if (mState.mSession != null && mState.isActive() == aActive) {
+            return;
+        }
```
bluemarvin, line 772: I think `setActive(true);` is okay and since we can get out of sync, I think we should just guard against `setActive(false)` twice?

PR #3380, opened 2020-05-15
Picovr suffix

keianhzo:
Pico now has a ROM baked version of FxR. We need to provide a suffix so QA can test staging build without colliding with the production appId.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -168,7 +168,7 @@ android {
                     arguments "-DVR_SDK_LIB=wavevr-lib", "-DWAVEVR=ON"
                 }
             }
-            applicationIdSuffix ".wavevr"
+            applicationIdSuffix ".internal"
         }
 
         wavevrStore {
@@ -184,6 +184,17 @@ android {
         }
 
         picovr {
+            dimension "platform"
+            externalNativeBuild {
+                cmake {
+                    cppFlags " -DPICOVR"
+                    arguments "-DPICOVR=ON"
+                }
+            }
+            applicationIdSuffix ".internal"
+        }
+
+        picovrStore {
             dimension "platform"
             externalNativeBuild {
                 cmake {
@@ -229,6 +240,8 @@ android {
                 'oculusvr3dofStoreArm64Release',
                 'picovrArm64Debug',
                 'picovrArm64Release',
+                'picovrStoreArm64Debug',
+                'picovrStoreArm64Release',
                 'wavevrArm64Debug',
                 'wavevrArm64Release',
                 'wavevrStoreArm64Debug',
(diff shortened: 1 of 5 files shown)


bluemarvin:
@keianhzo Maybe we should unify the suffix between WaveVR and pico? It would make scripting easier. Maybe `.internal`?

keianhzo:
@bluemarvin Updated.

bluemarvin:
Also the build targets need to be updated for taskcluster:
```
diff --git a/tools/taskcluster/build_targets.py b/tools/taskcluster/build_targets.py
index d14d65dc..cdc1b4f4 100644
--- a/tools/taskcluster/build_targets.py
+++ b/tools/taskcluster/build_targets.py
@@ -39,6 +39,7 @@ platforms = {
    'wavevr': ['arm64'],
    'wavevrStore': ['arm64'],
    'picovr': ['arm64'],
+   'picovrStore': ['arm64'],
    'noapi': ['arm64', 'x86_64'],
 }
```

bluemarvin:
@daoshengmu Should this:
https://github.com/MozillaReality/FirefoxReality/blob/bda5ecfa408dc2f8baa21460fd581c32cdf1c725/app/src/common/shared/org/mozilla/vrbrowser/telemetry/GleanMetricsService.java#L160

be update to:
```
private static void setStartupMetrics() {
    String type = BuildConfig.FLAVOR_platform;
    if (DeviceType.isOculusBuild()) {
        type = "oculusvr";
    } else if (DeviceType.isPicoVR()) {
        type = "picovr";
    }
    Distribution.INSTANCE.getChannelName().set(type);
}
```

daoshengmu:
> 
> 
> @daoshengmu Should this:
> 
> https://github.com/MozillaReality/FirefoxReality/blob/bda5ecfa408dc2f8baa21460fd581c32cdf1c725/app/src/common/shared/org/mozilla/vrbrowser/telemetry/GleanMetricsService.java#L160
> 
> be update to:
> 
> ```
> private static void setStartupMetrics() {
>     String type = BuildConfig.FLAVOR_platform;
>     if (DeviceType.isOculusBuild()) {
>         type = "oculusvr";
>     } else if (DeviceType.isPicoVR()) {
>         type = "picovr";
>     }
>     Distribution.INSTANCE.getChannelName().set(type);
> }
> ```

Yes. We can do this to keep our telemetry be matched. We might be able to consider to make `DeviceType.isWaveBuild())` and type = 'wavevrStore'. In that way, we can collect our data from different kinds of users. In our telemetry, we can use `channel==release` to filter developers.

daoshengmu:
@keianhzo We need to do the same change to here too, https://github.com/MozillaReality/FirefoxReality/blob/master/app/src/common/shared/org/mozilla/vrbrowser/telemetry/TelemetryWrapper.java#L148

keianhzo:
@daoshengmu @bluemarvin This should be ready.

PR #3381, opened 2020-05-15
Null check for the notification parent view

keianhzo:
I've come across a NullPointerException when showing multiple concurrent notifications in the Tray bar during testing.


bluemarvin:
@keianhzo Does this affect v10?

Issue #3324, opened 2020-05-04
Add legacy client id in deletion request ping in Glean


daoshengmu:
**Is your feature request related to a problem? Please describe.**
Per [the request](https://bugzilla.mozilla.org/show_bug.cgi?id=1634744) from Glean team, we need to add this delete request ping for the transition of moving out our legacy telemetry.

**Describe the solution you'd like**
[The ping](https://mozilla.github.io/glean/book/user/pings/deletion_request.html) is like the document mentioned. Add it in the `metrics.yaml` and call legacy_client_id.set(TelemetryHolder.get().getClientId() ) before Glean.Initialize().

PR #3384, opened 2020-05-15
Resolve label metrics are _other_ and add legacy client id

### LINKED ISSUE #3212: Tab opened and FxA received_tab metric labels are  `_other_` in the table
## Configuration

Firefox Reality version: 
Firefox Reality build ID: v.10

Hardware: 
All the devices

## Steps to Reproduce

1. Visit https://sql.telemetry.mozilla.org/queries/70315/source

## Current Behavior

We are using TabSource.name() and DeviceType.name(), it will return uppercase letters. According to Glean doc [1], it is not allowed.

[1] https://mozilla.github.io/glean/book/user/metrics/labeled_counters.html#limits

## Expected Behavior

The fields in our metric table should have keys and values instead of a `_other_`.

## Possible Solution

We should `toLowerCase()` it and give unit tests to prevent this kind of issue be happened again.

```
```

</details>

daoshengmu:
Fixes #3212 and #3324.

In #3212, our labels should be converted to lower case, and add legacy client id in deletion request ping for #3324.

### DIFF
--- a/app/metrics.yaml
+++ b/app/metrics.yaml
@@ -80,7 +80,8 @@ searches:
   counts:
     type: labeled_counter
     description: >
-      Counting how many searches are queried in a specific search engine. The search engine `identifier`s are used as keys for this metric.
+      Counting how many searches are queried in a specific search engine.
+      The search engine `identifier`s are used as keys for this metric.
     bugs:
       - https://github.com/MozillaReality/FirefoxReality/issues/2230
     data_reviews:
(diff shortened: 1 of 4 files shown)


daoshengmu:
@Dexterp37 Please help review. I am also interested if our #3129 issue would be resolved after we did a Glean update. Is there a way to verify or add a test for it?

app/metrics.yaml
```diff
@@ -254,3 +254,19 @@ control:
       - fxr-telemetry@mozilla.com
       - dmu@mozilla.com
     expires: "2020-11-01"
+
+legacy_telemetry:
+  client_id:
+    type: uuid
+    description: >
+      A UUID uniquely identifying the legacy client.
```
Dexterp37, line 262:
```suggestion
      A UUID uniquely identifying the legacy telemetry client id.
      This is used for supporting legacy telemetry in the `deletion-request` ping.
```

app/metrics.yaml
```diff
@@ -254,3 +254,19 @@ control:
...
+    send_in_pings:
+      - session-end
```
Dexterp37, line 264:
```suggestion
      - deletion-request
```

Dexterp37:
> 
> 
> @Dexterp37 Please help review. I am also interested if our #3129 issue would be resolved after we did a Glean update. Is there a way to verify or add a test for it?

Yes, it received some fixes after v25 (see [the changelog](https://github.com/mozilla/glean/blob/master/CHANGELOG.md#v2500-2020-02-17)).

Dexterp37:
@daoshengmu are you planning on merging this soon?

daoshengmu:
Fixes #3129

PR #3390, opened 2020-05-18
Adjust Oculus Go WebXR exit arrow to the back button instead of the system button.

daoshengmu:
The arrow currently points to the system button instead of the back button that we exactly wanna users to use. We should do this change.


Softvision-GeluHaiduc:
I have verified this with the latest Firefox Reality version 11-141402320 (eda7d955) installed on the Oculus Go headset.
Now, the WebXR exit arrow is pointed to the correct button.
org mozilla vrbrowser-20200520-111113

PR #3428, opened 2020-05-27
OpenXR implementation

MortimerGoro:
Known issues (already reported to Oculus):

- Washed out colors: Seems related to having SRGB by default or not applying VRAPI_FRAME_LAYER_FLAG_INHIBIT_SRGB_FRAMEBUFFER
- Quad and Cube layers work correctly when using FBO swapchains. But when redenring compositor layers with AndroiSurface based swapChain I get a `XR_ERROR_LAYER_INVALID` error when calling xrEndFrame. AndroidSurface layers are disabled until the issue is fixed.
- Reusing the instances of input XrActions with different subactionPaths for each controller on the Quest doesn't work well: the right controller reported the same pose and buttons as the left controller. I found a workaround. It works well creating different instance XrActions for each subactionPath/Controller. I'll switch back to single XrAction instances once the problem is fixed (seems a OpenXR SDK bug)

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,13 @@ npm install
 npm run compress
 ```
 
+## Enable OpenXR builds
+You can enable OpenXR API for Oculus by adding this property to your `user.properties` file:
+
+```ini
+openxr=true
+```
+
 ## Development troubleshooting
 
 ### `Device supports , but APK only supports armeabi-v7a[...]`
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -82,7 +82,34 @@ target_link_libraries(native-lib native_app_glue)
 set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -u ANativeActivity_onCreate")
 endif()
 
-if(OCULUSVR)
+
+if(OPENXR)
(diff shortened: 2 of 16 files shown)


bluemarvin:
I wonder how much effort it would take to make a noapi version using the OpenXR device. :thinking:

MortimerGoro:
> I wonder how much effort it would take to make a noapi version using the OpenXR device. 🤔

We would need some loader or OpenXR runtime for noapi. I wonder if something like https://monado.freedesktop.org/ could do the job.

> I think we should enable this to build on master tires to prevent bit rot.

Is `user.properties` path easy to support in TC?

bluemarvin:
> > I wonder how much effort it would take to make a noapi version using the OpenXR device. thinking
> 
> We would need some loader or OpenXR runtime for noapi. I wonder if something like https://monado.freedesktop.org/ could do the job.
> 

I'll take a look.

> > I think we should enable this to build on master tires to prevent bit rot.
> 
> Is `user.properties` path easy to support in TC?

I probably could support it but it would be easier if it was just another platform since I would need to run the regular builds, then set the property and then run the build just for the OpenXR build.

bluemarvin:
Opened this issue https://github.com/MozillaReality/FirefoxReality/issues/3447

I don't want this to bitrot from lack of building on taskcluster.

PR #3441, opened 2020-05-29
What's new implementation

keianhzo:
Based on: https://trello.com/c/dzS3HOql/559-uis-97-whats-new-experienceimplementation


bluemarvin:
org mozilla vrbrowser-20200529-123739
It looks like `Report an Issue` is slightly offset to the right.

bluemarvin:
How do I get the what's new button to show in the navigation bar?

keianhzo:
@bluemarvin I've updated the PR. You should see the what's new icon in the URL bar every time you run FxR from AS as it generates a new version code.

bluemarvin:
So it looks like the first time FxR is run and the setting `settings_key_app_version` doesn't exist, the what's new button in the nav bar is not shown. To reproduce:

1) Quit FxR
2) `adb shell pm clear org.mozilla.vrbrowser`
3) Launch FxR

Actual:
No `What's New` button

Expected:
`What's New` button is in the nav bar.

keianhzo:
@bluemarvin fixed

keianhzo:
@bluemarvin I thought we wanted what's new for every release. If we don't we might need to do some manual update based what's new like what you proposed but we will need to remember to always update the URL whenever we want to button to be shown.

bluemarvin:
> @bluemarvin I thought we wanted what's new for every release. If we don't we might need to do some manual update based what's new like what you proposed but we will need to remember to always update the URL whenever we want to button to be shown.

My concern is we aren't great at updating what is new. So would rather we only show the button when we actually have something new to show and I don't really want to show the button for dot releases since bug fixes really aren't worth the users time. I guess ideally it would dynamically check but that probably too much effort for this issue.

keianhzo:
@bluemarvin An alternative would be to only show the button for major releases and leave out the dot releases. Would that be a good compromise?

keianhzo:
@bluemarvin @Softvision-RemusDranca I've updated the PR with the new flow. 

Now we user a romote JSON to sho/hide the What's new button. The remote properties JSON lives in https://mixedreality.mozilla.org/FirefoxReality/props.json (at least until we have a app specific backend for FxR). This JSON would contain any remote properties that we would want to sync with the app in a dictionary by version. At the moment the only property is the What's New URL. The app will sync the remote properties at startup, if the remote properties cannot be retrieved the cached ones are used (if existing), otherwise the remote ones will be persisted locally. In any case when the remote properties consumers would be notified.

What's new should be displayed?
- The first app start after install
- If the what's new button (from settings/navigation bar) hasn't been clicked before && the remote properties JSON has an entry for that version.
- If the what's new button (from settings/navigation bar) has been clicked before && the app version is different from the one when the what's new button was last clicked && the remote properties JSON has an entry for that version.
- Otherwise the button should remain hidden

Softvision-GeluHaiduc:
@keianhzo I have some follow-up questions regarding the behavior:
-  After clicking the “What’s New” button from settings/navigation bar does the button disappear from the settings menu as well or just from the navigation bar? Or does only the blue dot from the button in settings disappear?
-  Is it possible to have multiple What's New URL on a single version of the app?
-  How are dot releases treated in the JSON file? Is the What's New URL from V11 going to work in V11.1, for example, or will there be a new version added in the JSON file for the dot release?

keianhzo:
>     After clicking the “What’s New” button from settings/navigation bar does the button disappear from the settings menu as well or just from the navigation bar? Or does only the blue dot from the button in settings disappear?

Only the navigation bar should dissappear, the Settings one should stay but without the badge.

>     Is it possible to have multiple What's New URL on a single version of the app?

No, there is only one URL per version.

>     How are dot releases treated in the JSON file? Is the What's New URL from V11 going to work in V11.1, for example, or will there be a new version added in the JSON file for the dot release?

We can show What's New for any version the only condition is that the app version matches one in the JSON file, it can be a major or a dot.

Softvision-GeluHaiduc:
I have verified this feature with multiple builds installed on the Oculus Go, HTC Vive Focus Plus, Pico Neo2 and Pico G2 4K headsets. Every issue found was logged, fixed and verified.
Given this I will remove the QA label.

PR #3443, opened 2020-05-30
Wait to load URI from intent until after session restore.

### LINKED ISSUE #3431: [Pico] Cannot send url when launch browser activity if FxR is not running
## Configuration

Firefox Reality version: 10-rc7
Firefox Reality build ID: bda5ecf

Hardware: Pico Neo2 - Android 8.1.0 - PUI 3.11.3

## Steps to Reproduce

1. Kill Firefox Reality application to make sure it's not running on background
2. Use adb command: ``adb shell am start -a android.intent.action.VIEW -d "https://aframe.io/" org.mozilla.vrbrowser/org.mozilla.vrbrowser.VRBrowserActivity`` to launch Firefox Reality and send url meanwhile
3. Observe the behavior.

## Current Behavior

Firefox Reality launches with success but it couldn't direct to the url specified in adb command

## Expected Behavior

Firefox Reality launches and direct to the specified url meanwhile.

## Context

1. The use case is to open up a link in Firefox Reality from another app. 
2. Oculus version seems to behave the same way.
3. Url can be sent correctly if FxR is running on background.

<details open>

```
```

</details>

bluemarvin:
Fixes #3431

This patch changes the behavior of launching or opening a URI
from an intent. Previously the URI would be opened in the focuses window.
With this patch it matches Fenix and desktop and creates a new tab in the
focused window.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserActivity.java
@@ -93,11 +93,9 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
-import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.Consumer;
-import java.util.function.Function;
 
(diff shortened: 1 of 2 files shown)


bluemarvin:
We may want to make this part of the v10.2 release. Regressed by #3364

bluemarvin:
@keianhzo Updated.

PR #3451, opened 2020-06-02
Fix for l10n popup strings

keianhzo:
Fixes https://github.com/mozilla-l10n/android-l10n/issues/99


keianhzo:
@flodolo this should fix mozilla-l10n/android-l10n#99

PR #3394, opened 2020-05-19
Fixes #2919 Fixes #2673 Fixes #2641 Hide Youtube fixes

keianhzo:
Fixes #2919 Fixes #2947 Fixes #2673 Hide YouTube overlays in immersive

PR #3466, opened 2020-06-04
Update video projection for currently playing video

keianhzo:
We just set the video projection when we enter full screen so when videos are playing in a list or next video is played, the previous video projection remains set. This PR should address that issue and always set the currently playing video projection. If the user changes the projection for the currently playing video, that would be used until the next video is played or full-screen mode is exited.

STRs:
- Create a playlist with 360 and flat videos
- Play the first video fullscreen
- Click the next button to play the next one or let the video end.

The next video should be played and the correct projection should be used.

**Note**: This will only work when in immersive/full-screen.

Related to https://github.com/MozillaReality/FirefoxReality/pull/3394#issuecomment-631127335

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/NavigationBarWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/NavigationBarWidget.java
@@ -15,7 +15,6 @@ import android.util.AttributeSet;
 import android.util.Log;
 import android.util.Pair;
 import android.view.LayoutInflater;
-import android.view.MotionEvent;
 import android.view.View;
 import android.webkit.URLUtil;
 import android.widget.EditText;
@@ -29,6 +28,7 @@ import androidx.lifecycle.ViewModelProvider;
 
 import org.mozilla.geckoview.GeckoSession;
 import org.mozilla.geckoview.GeckoSessionSettings;
+import org.mozilla.geckoview.MediaElement;
 import org.mozilla.vrbrowser.R;
 import org.mozilla.vrbrowser.VRBrowserActivity;
 import org.mozilla.vrbrowser.VRBrowserApplication;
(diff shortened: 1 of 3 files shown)


bluemarvin:
One thing I noticed is if I start from a video in the feed, since it sets the projection in the URL it always uses that even when a video does not have a projection. In my testing where I didn't start from the feed so it didn't have the extra data in the URL, it would work as expected for an ad and drop out of immersive mode, but if the content after the add was 360, it would stay in fullscreen and never return to immersive mode.

keianhzo:
@bluemarvin We set the auto projection based on the video title/description and those doesn't change when an ad is playing so it's not easy to detect to update the projection. In your case it is probably exiting immersive mode when an ad plays because the next video is flat. Also it's quite hard for some reason to get ads, I guess it must be because of the UA?

bluemarvin:
I don't know if there is a good solution. I just wasn't sure if what I was seeing was expected. I think I would rather it not drop out of immersive mode for ads if we can't go back into immersive mode when the next video starts.

keianhzo:
That should be the behavior of this PR. If the next video is 3D, it will stay in immersive mode regardless of whether or not there is an ad before the video.

keianhzo:
@bluemarvin do you think that this is worth landing?

bluemarvin:
@keianhzo Maybe a few other people should try it and see what they think. I don't know if there is a right answer without meta data for the video. I'll try it again.

bluemarvin:
I guess I still don't understand the expected behavior. From the STR it seems like it should be able to switch between flat fullscreen and immersive mode as the videos change but I only ever see it drop out of immersive mode and it never goes back. Is that the expected behavior?

keianhzo:
@bluemarvin I see it switching from flat to 360 as long as the next video is 360. Can you share the playlist that you are using?

bluemarvin:
I think we should move this to v12. I'm not sure I understand how it is supposed to work yet.

PR #3474, opened 2020-06-05
Merge console and debug logging settings

bluemarvin:
Due to the way the settings work in gecko, having two switches
doesn't make sense. Having one switch to enable/disable Gecko
logging to logcat makes it easier to manage.


bluemarvin:
`GeckoRuntimeSettings.builder.debugLogging()` is a super set of `GeckoRuntimeSettings.builder.consoleOutput()` so setting debugLogging to false would also disable consleOutput. So it doesn't make sense to keep the consoleOutput switch since it is controlled by debugLogging. Now in FxR debug logging is enabled in debug builds and disabled by default in release builds and there is only one switch to rule them all.

PR #3493, opened 2020-06-11
Update FxR codebase to use inclusive language. Fixes #3492

### LINKED ISSUE #3492: Remove instances of non-inclusive language from the Firefox Reality code base
## Configuration

Firefox Reality version: All

## Current Behavior

Mozilla is removing non-inclusive language from the code base. See https://bugzilla.mozilla.org/show_bug.cgi?id=1644823 for reference.

I did a quick search of FxR code base and found a few instances we should fix. The biggest issue is the main github repository branch is called `master`. Renaming that will be a larger effort as there are taskcluster process that unfortunately depend on this name.

## Expected Behavior

FxR should use only inclusive language in the code.

## Possible Solution

Rename problematic language.

bluemarvin:


bluemarvin:
I will create a separate issue for renaming the main github repo from master to something more inclusive as that will require some taskcluster updates.

app/src/common/shared/org/mozilla/vrbrowser/audio/AudioEngine.java
```diff
@@ -124,12 +124,12 @@ public void playSound(Sound aSound) {
     }
 
     public void setMasterVolume(float aVolume) {
```
keianhzo, line 126: @bluemarvin We probably want to rename that one too

PR #3498, opened 2020-06-12
Closes #2267 AC WebExtensions + WebCompat + FxAWebChannels

### LINKED ISSUE #2267: Migrate FxA implemetation to WebChannels instead of interceptor

### LINKED ISSUE #3339: Use Gecko as HTTP client for Glean
I saw that we are initializing Glean with the default client - which means it will be using `HttpUrlConnection`.

In AC we have an adapter that wraps GeckoView and will use Gecko instead of `HttpUrlConnection`. This is what Fenix is using as well. I assume that is something Firefox Reality would want too.

There are two options:
* Use the `Client` implementation from AC. However since Firefox Reality and AC are not synchronized on GeckoView versions this could cause breakage if the API in GeckoView changes in a breaking way.
* Add a simple `Client` implementation wrapping GeckoView directly to Firefox Reality. Even in AC the code is minimal: https://github.com/mozilla-mobile/android-components/blob/master/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/fetch/GeckoViewFetchClient.kt

keianhzo:
Closes #2267 Closes #3339 This PR adds support for the following:
- AC 44.0.0
- AC Web Extensions bridge. We are using a `ComponentsAdapter` as a bridge between our `SessionStore` and AC's `BrowserStore` which is required by the AC WebExtensions/Features lifecycle.
- Moves our current builtin extensions (YouTube and Vimeo) to AC `WebExtensionsRuntime`.
- Adds support for the `WebCompat` extension.
- The `SessionStore` is now the `WebExtensionDelegate`
- Session life cycle code refactor to establish the `SessionStore` as the single point of communications with the `BrowserStore`.

There are some Glean related updates, mainly initialization changes and lint errors in .md files, could you please @daoshengmu verify that everything is working as expected?

@bluemarvin @MortimerGoro I've have replaced the appcompat aar for the maven version so it stays aligned with the rest of the androidx components. I couldn't find any speech related classes in that module.

**QA**: 
This should affect mainly to the WebExtensions. Things that we should keep an eye on:
- Youtube and Vimeo WebExtensions lifecycle (correct load of the extensions in restored session)
- FxA login: This shouldn't change, we are now using WebChannels but the flow is still the same.
- Would be also worth keeping an eye on sessions state: creation, recreation, shutdown and restore.


keianhzo:
@pocmo As you were involved with the `ComponentsAdapter` implementation, could you please help review? Thanks!

daoshengmu:
Please rebase your patch with master first. I was helping you to rebase it in local, but you were working on your own remote, so I can't help you give a commit.

After my investigation, your patch will skip calling `VRBrowserApplication::onActivityCreate()` when running unit tests. That actually breaks our unit test mechanism. I need do more investigation to understand why.

daoshengmu:
If we move back
```
        TelemetryWrapper.init(this);
         GleanMetricsService.init(this);
```
to `onCreate()`, the unit tests can be executed. In addition, `Configuration` in Glean seems to be unable to access directly when using Java.

daoshengmu:
Btw, we are better not to use the default client uploader. We should use GeckoView client uploader instead. I can work on it at another PR.

https://github.com/MozillaReality/FirefoxReality/issues/3339

keianhzo:
@daoshengmu I've fixed the rebase issues.

> BTW, we are better not to use the default client uploader. We should use GeckoView client uploader instead. I can work on it at another PR.

This PR already uses the `GeckoWebExecutor` in the Configuration but it seems that you reverted that in your branch.

The main difference between this PR and master is that it uses the `GeckoWebExecutor` in the Glean configuration which forces a Runtime initialization in the Application's `onCreate` while before we weren't creating the runtime until the Activity's `onCreate`. 

That seems to break Gecko somehow as nothing is renderer anymore and it also breaks the tests with:
> java.lang.IllegalArgumentException: Crash handler service must run in a separate process

As it seems that the `getSystemService` call that happens in the `GeckoRuntime` initialization is not returning the correct process for the Crash handler service. I guess the testing context is not loading the correct manifest information when using the `Application` context?

The weird thing is that the GeckoRuntime docs states:
`Create a new runtime with default settings and attach it to the given context. Create will throw if there is already an active Gecko instance running, to prevent that, bind the runtime to the process lifetime instead of the activity lifetime.` So it should work.

app/src/common/shared/org/mozilla/vrbrowser/browser/components/GeckoEngineSession.kt
```diff
@@ -0,0 +1,40 @@
...
+import mozilla.components.concept.engine.Settings
+import org.json.JSONObject
+import org.mozilla.geckoview.GeckoSession
+import org.mozilla.vrbrowser.browser.engine.SessionStore
+
+class GeckoEngineSession(
+        val geckoSession: GeckoSession
+): EngineSession() {
+    constructor() :
+        this(SessionStore.get().createSession(false, false).geckoSession)
```
pocmo, line 14: Is this empty constructor, that explicitly creates a session, needed? Looks like that would cause problems because we never want to create one and always use the one you already created in FxR code?

app/src/common/shared/org/mozilla/vrbrowser/browser/adapter/ComponentsAdapter.kt
```diff
@@ -0,0 +1,68 @@
+package org.mozilla.vrbrowser.browser.adapter
```
pocmo, line 1: nit: Do you use license headers? I probably didn't add any in my prototype.

app/src/common/shared/org/mozilla/vrbrowser/browser/components/GeckoEngineSession.kt
```diff
@@ -0,0 +1,40 @@
+package org.mozilla.vrbrowser.browser.components
```
pocmo, line 1: nit: license?

app/src/common/shared/org/mozilla/vrbrowser/browser/adapter/ComponentsAdapter.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.mozilla.vrbrowser.browser.components.GeckoEngineSession
+import org.mozilla.vrbrowser.browser.engine.Session
+
+class ComponentsAdapter private constructor(
+        val store: BrowserStore = BrowserStore()
+) {
+    companion object {
+        private val instance: ComponentsAdapter = ComponentsAdapter()
+
+        fun get(): ComponentsAdapter = instance
```
pocmo, line 17: nit: You could consider using `@JvmStatic` here so that you can use `ComponentsAdapter.get()` directly from Java.

app/src/common/shared/org/mozilla/vrbrowser/browser/components/GeckoEngineSession.kt
keianhzo: It seems that in some cases the an extension needs to create a new tab and uses that constructor to do so?

https://github.com/mozilla-mobile/android-components/blob/d642612386e8d9880ff83a78737e065b8ac870c7/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt#L80-L82

        override fun onToggleActionPopup(extension: WebExtension, action: Action): EngineSession? {
            return webExtensionDelegate?.onToggleActionPopup(extension, GeckoEngineSession(runtime), action)
        }

pocmo: Oh, I see! Then I guess this makes sense if that will then let you know about this tab too. :)

keianhzo:
@daoshengmu I've fixed the GV issue but I still can see issues with test, especially thee issue with the crashservice process not being correctly determined from the manifest in test. I'll have look into that but if you have any ideas, they are welcom.

app/src/common/shared/org/mozilla/vrbrowser/browser/components/GeckoEngineSession.kt
keianhzo: Yes, calling `SessionStore.get().createSession` should both add the session to the SessionStore sessions list and to the BrowserStore.

app/src/common/shared/org/mozilla/vrbrowser/browser/components/GeckoWebExtensionRuntime.kt
```diff
@@ -0,0 +1,281 @@
...
+
+            val privateBrowsingMigrated = sharedPref.getBoolean(MOZAC_HAS_RESET_WEBEXT_PRIVATE_BROWSING_DEFAULT, false)
+            extensions.forEach { extension ->
+                // As a workaround for https://bugzilla.mozilla.org/show_bug.cgi?id=1621385,
+                // we set all installed extensions to be allowed in private browsing mode.
+                // We need to revert back to false which is now the default.
+                if (!extension.isBuiltIn() && extension.isAllowedInPrivateBrowsing() && !privateBrowsingMigrated) {
+                    setAllowedInPrivateBrowsing(extension, false)
+                }
+
```
bluemarvin, line 224: Do we need this? Looks like some compat code that won't affect us.

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/Session.java
```diff
@@ -521,19 +517,24 @@ private GeckoSession createGeckoSession(@NonNull SessionSettings aSettings) {
...
         boolean wasFullScreen = mState.mFullScreen;
 
         SessionState previous = mState;
         mState = mState.recreate();
 
-        restore();
+        mSessionChangeListeners.forEach(listener -> {
+            listener.onSessionClosed(this);
+            listener.onSessionRemoved(this);
+        });
```
bluemarvin, line 537: Why are we closing the session and removing it after it is recreated?

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStore.java
```diff
@@ -89,15 +112,29 @@ public void onTrackingProtectionLevelUpdated(int level) {
...
+
+        mServices = ((VRBrowserApplication)context.getApplicationContext()).getServices();
 
-    public void initializeStores(Context context) {
         mBookmarksStore = new BookmarksStore(context);
         mHistoryStore = new HistoryStore(context);
+
+        // Web Extensions initialization
+        YoutubeExtension.install(mWebExtensionRuntime);
+        VimeoExtension.install(mWebExtensionRuntime);
```
bluemarvin, line 126: This pattern doesn't scale. If the WebExtenstionRuntime API changes we have to update every class. I especially dislike it because the only difference between the two classes is the meta data.

app/src/common/shared/org/mozilla/vrbrowser/browser/SettingsStore.java
```diff
@@ -738,5 +736,15 @@ public void setDownloadsSortingOrder(@SortingContextMenuWidget.Order int order)
...
+
+    public void setFxAWebChannelsEnabled(boolean isEnabled) {
+        SharedPreferences.Editor editor = mPrefs.edit();
+        editor.putBoolean(mContext.getString(R.string.settings_key_webchannels_enabled), isEnabled);
+        editor.commit();
+    }
+
+    public boolean isFxAWebChannelsEnabled() {
+        return mPrefs.getBoolean(mContext.getString(R.string.settings_key_webchannels_enabled), FXA_WEBCHANNELS_ENABLED);
+    }
```
bluemarvin, line 748: Is there a reason we want to be able to turn WebChannels off? For QA and testing?

daoshengmu:
> @daoshengmu I've fixed the GV issue but I still can see issues with test, especially thee issue with the crashservice process not being correctly determined from the manifest in test. I'll have look into that but if you have any ideas, they are welcomed.

You can run `./gradlew app:testNoapiArm64DebugUnitTest` to execute the test. Or just pick one `testXXX ` from `./gradlew tasks`.

So far, I notice there is no one instances `VRBrowserActivity`, but it do enter `VRBrowserApplication::onCreate()`.

keianhzo:
@daoshengmu I have fixed the Telemetry tests. The GV team recommended to move the GeckoRuntime startup code to the Activity onCreate so I moved all the initialization there and initialized the Telemetry services for tests in a setup method. Seems to be working fine but I needed to add a flag to avoid the legacy LegacyTelemetry initialization.

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/Session.java
keianhzo: Yep, good catch, we don't need to do that, just close the old one and open the new one. I've updated.

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStore.java
keianhzo: We can add a builtin extensions manager that initializes all of them based on metadata as we had before. I don't expect that we have many builtin extensions though mot of them are going to come from AC or the Addon Manager in the future, so I'm not sure if it'll be that useful.

app/src/common/shared/org/mozilla/vrbrowser/browser/SettingsStore.java
keianhzo: I just left it just in case we want to enable/disable easily in the future. I can remove it, is not really used right now.

app/src/common/shared/org/mozilla/vrbrowser/browser/components/GeckoWebExtensionRuntime.kt
keianhzo: Not really. Removed.

app/src/common/shared/org/mozilla/vrbrowser/VRBrowserApplication.java
```diff
@@ -50,6 +47,16 @@ protected void onActivityCreate() {
...
         mSpeechService = new SpeechService(this);
+        mAppExecutors = new AppExecutors();
+        mBitmapCache = new BitmapCache(this, mAppExecutors.diskIO(), mAppExecutors.mainThread());
+
+        TelemetryWrapper.init(this,
+                EngineProvider.INSTANCE.getDefaultClient(this)
+        );
+        GleanMetricsService.init(this,
+                EngineProvider.INSTANCE.getDefaultClient(this),
+                true
```
daoshengmu, line 58: Remove `true` flag

app/src/common/shared/org/mozilla/vrbrowser/browser/Services.kt
```diff
@@ -112,6 +120,7 @@ class Services(val context: Context, places: Places): GeckoSession.NavigationDel
 
     ).also {
         it.registerForAccountEvents(deviceEventObserver, ProcessLifecycleOwner.get(), true)
+
```
daoshengmu, line 123: Redundant newline?

app/src/common/shared/org/mozilla/vrbrowser/telemetry/GleanMetricsService.java
```diff
@@ -68,8 +72,14 @@ public static void init(Context aContext) {
             GleanMetricsService.stop();
         }
 
-        LegacyTelemetry.INSTANCE.clientId().set(UUID.fromString(TelemetryHolder.get().getClientId()));
-        Configuration config = new Configuration(Configuration.DEFAULT_TELEMETRY_ENDPOINT, BuildConfig.BUILD_TYPE);
+        if (initializeLegacy) {
+            LegacyTelemetry.INSTANCE.clientId().set(UUID.fromString(TelemetryHolder.get().getClientId()));
+        }
```
daoshengmu, line 77:
```suggestion
        LegacyTelemetry.INSTANCE.clientId().set(UUID.fromString(TelemetryHolder.get().getClientId()));
```

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStore.java
bluemarvin: Or just have a single class that takes the meta data in the constructor. I guess it doesn't matter.

app/src/common/shared/org/mozilla/vrbrowser/browser/SettingsStore.java
bluemarvin: I wasn't sure if there was a reason. I would remove it if we don't want to support the previous version.

app/src/common/shared/org/mozilla/vrbrowser/telemetry/GleanMetricsService.java
```diff
@@ -54,7 +58,7 @@
     private static GleanTimerId openPrivateWindowTimerId[] = new GleanTimerId[MAX_WINDOWS];
 
     // We should call this at the application initial stage.
-    public static void init(Context aContext) {
+    public static void init(@NonNull Context aContext, @NonNull Client client, boolean initializeLegacy) {
```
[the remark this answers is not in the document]
keianhzo: ok, I wasn't sure if that test was important. I've removed it and got rid of the legacy initialization flag.

app/src/common/shared/org/mozilla/vrbrowser/browser/SettingsStore.java
keianhzo: Removed

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStore.java
keianhzo: Fixed, I've added a BuiltinExtension class that handles the initialization for our builtin extensions.

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStore.java
```diff
@@ -359,4 +443,139 @@ public void removePermissionException(@NonNull String uri, @SitePermission.Categ
...
+
+    @Override
+    public void onUpdatePermissionRequest(@NonNull WebExtension webExtension, @NonNull WebExtension webExtension1, @NonNull List<String> list, @NonNull Function1<? super Boolean, Unit> function1) {
+
+    }
+
+    @Override
+    public void onExtensionListUpdated() {
+
+    }
```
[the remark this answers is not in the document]
keianhzo: They don't have default implementations for that interface. We don't use them right now but we will when we add support for the addon manager.

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/Session.java
```diff
@@ -454,7 +448,7 @@ private void loadDefaultPage() {
         }
     }
 
-    private void restore() {
+    private void restore(boolean isRecreation) {
```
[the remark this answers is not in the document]
keianhzo: It seems that the SessionState is null when is not a recreation. I've removed the flag.

PR #3507, opened 2020-06-16
Avoid showing Zh-TW and Zh-CN keyboard candidate view when typing symbols.

### LINKED ISSUE #3506: [Keyboard: zh-TW] The bar displayed at the top of the keyboard is auto dismissed when inserting special characters in search fields
**[Affected Versions]:**
- Firefox Reality version 11-141670446 (6b20069).

**[Affected Platforms]:** 
- Pico Neo2 - Android 8.1.0

**[Prerequisites]:**
- The keyboard language is set to zh-TW.

**[Steps to reproduce]:**
1. Navigate to a website that has a search bar (Ex Youtube).
2. Focus the search bar so that the keyboard is displayed.
3. Set the keyboard to display special characters.
4. Click the keys and observe the bar displayed at the top of the keyboard.

**[Expected results]:**
- The bar is displayed and remains displayed.
OR
- The bar is not displayed at all.

**[Actual results]:**
- The bar is briefly displayed an then auto dismissed.

**[Notes]:**
- The issue is not reproducible with the Awesomebar.
- When clicking some of the keys, the bar is dismissed very quickly and it's not caught on video.
	- Screen Recording: [link](https://drive.google.com/file/d/11EdkKzWZsTFeCa5ENX4g6t__4Ry5IPNK/view?usp=sharing).

daoshengmu:
Fixes #3506.

When typing symbols, we can just use `AUTO_COMPOSE` mode.

I was trying to make Zh-CN and JPN keyboard have the same behavior when typing symbols, using `AUTO_COMPOSE` instead of `SHOW_CANDIDATES`. But, JPN keyboard will show `．` when inserting `/`。The only way possible is fixing their code book to avoid it, but it is not worth to do. IMO, We can still rely on their code book to feed our data, iOS JPN keyboard uses the same behavior either.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/keyboards/ChinesePinyinKeyboard.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/keyboards/ChinesePinyinKeyboard.java
@@ -36,9 +36,6 @@ public class ChinesePinyinKeyboard extends BaseKeyboard {
     private DBHelper mDB;
     private HashMap<String, KeyMap> mKeymaps = new HashMap<>();
     private HashMap<String, KeyMap> mExtraKeymaps = new HashMap<>();
-    private List<Character> mAutocompleteEndings = Arrays.asList(
-            ' ', '，', '。','!','?','ー'
-    );
 
     public ChinesePinyinKeyboard(Context aContext) {
         super(aContext);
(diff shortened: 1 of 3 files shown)


daoshengmu:
It should be solved. I have made Zh-CN and Zh-TW have consistent behaviors.

PR #3527, opened 2020-06-22
Fixes #2442 Perform widgets parenting sorting only in case of equal depth

### LINKED ISSUE #2442: The awesomebar suggestions are wrongly displayed over the on screen keyboard when the awesomebar is hovered
**[Affected Versions]:**
-  Firefox Reality version 7 RC6 (b01367aa)

**[Affected Platforms]:**

**[Prerequisites]:**

**[Steps to reproduce]:**
1. Click inside the awesomebar.
2. Move the on screen keyboard above the awesomebar.
3. Write some words using the on screen keyboard.
4. Hover the awesomebar and observe the behaviour.

**[Expected results]:**
- The awesomebar suggestions are displayed over the keyboard.

**[Actual results]:**
- The awesomebar suggestions are not displayed over the keyboard.

**[Notes]:**
- The issue is not reproducible if the awesomebar suggestions are hovered.
- Attached a screen recording of the issue [here](https://drive.google.com/open?id=1p48Il0mAQuDeCw-_JDK3lJH_bVT8-fbY).

keianhzo:
Fixes #2442 Perform widgets parenting sorting only in case of equal depth

### DIFF
--- a/app/src/main/cpp/BrowserWorld.cpp
+++ b/app/src/main/cpp/BrowserWorld.cpp
@@ -717,8 +717,8 @@ BrowserWorld::State::SortWidgets() {
     Widget* wa = da->second.first;
     Widget* wb = db->second.first;
 
-    // Parenting sort
-    if (wa && wb && wa->IsVisible() && wb->IsVisible()) {
+    // Parenting sort only in case both elements have the same normalized z value
+    if (wa && wb && wa->IsVisible() && wb->IsVisible() && da->second.second == db->second.second) {
       if (IsParent(*wa, *wb)) {
         return true;
       } else if (IsParent(*wb, *wa)) {


app/src/main/cpp/BrowserWorld.cpp
```diff
@@ -717,8 +717,8 @@ BrowserWorld::State::SortWidgets() {
     Widget* wa = da->second.first;
     Widget* wb = db->second.first;
 
-    // Parenting sort
-    if (wa && wb && wa->IsVisible() && wb->IsVisible()) {
+    // Parenting sort only in case both elements have the same normalized z value
+    if (wa && wb && wa->IsVisible() && wb->IsVisible() && da->second.second == db->second.second) {
```
bluemarvin, line 721: Since the depth is a `float` we should probably be checking if the two depth values are within a certain epsilon of each other and not use `==`. For example:
```
std::fabs(da->second.second - db->second.second) < FLT_EPSILON
```

MortimerGoro:
I added another solution here: https://github.com/MozillaReality/FirefoxReality/pull/3533

keianhzo:
Closing this as there is another better approach.

PR #3546, opened 2020-06-24
Ensure the pref overrides are written out before GeckoRuntime is created

bluemarvin:


bluemarvin:
Moving the GeckoRuntime creation to earlier in the activity creation meant the pref overrides were not written out yet and thus WebXR was not enabled and WebRender was getting enabled on first run.

PR #3559, opened 2020-06-25
Convert master branch to use main

bluemarvin:


bluemarvin:
We don't want to land this until the `main` branch is created and set as the default and the index path has been updated.

keianhzo:
@Pike @Delphine Do you need to do any changes to the automation bot for the automatic PRs to work properly with the new `main ` branch?

Pike:
I guess this is best answered by @tomprince , but the answer is "yes".

One thing we need to change for sure is https://github.com/mozilla-l10n/android-l10n/blob/0caabe42a3638077cda59b3ba2f10d5cf0b5c2b1/config.toml#L19

PR #3561, opened 2020-06-26
Support for remote environments + seasons cubemaps

keianhzo:
Support for remote environments + seasons cubemaps


versions.gradle
```diff
@@ -32,7 +32,7 @@ versions.android_components = "44.0.0"
 // that we depend on directly for its rustlog package, and it's important
 // that it be kept insync with the version used by android-components above.
 versions.mozilla_appservices = "0.58.2"
-versions.mozilla_speech = "2.0.1"
+versions.mozilla_speech = "2.0.4"
```
bluemarvin, line 35: This should probably be a different PR?

keianhzo:
Moved all the envs but void and offworld to remote.

PR #3579, opened 2020-06-30
Add Thai to supported localized languages

daoshengmu:
Add Thai to supported language.


daoshengmu:
I feel we need to sort these language by alphabet when showing on the option menu. @keianhzo how do you think?

keianhzo:
@daoshengmu We don't have support for Thai yet as stated by @Delphine in https://github.com/mozilla-l10n/android-l10n/pull/243#discussion_r447316433, I think we should wait until we have to add it to the supported languages list as this setting will allow the user to change the FxR UI language and we don't have any yet.

daoshengmu:
We seems like not finish the Thai localization work yet. I don't see we have a folder of `values-thai`. Even mozilla-l10n/android-l10n#243 has been merged.

bluemarvin:
The automated l10n system is currently not working.

Delphine:
> We seems like not finish the Thai localization work yet. I don't see we have a folder of `values-thai`. Even [mozilla-l10n/android-l10n#243](https://github.com/mozilla-l10n/android-l10n/pull/243) has been merged.

In fact nothing has changed since my previous comment: we've never had Thai localization in Reality, which is why it doesn't show up: https://github.com/mozilla-l10n/android-l10n/blob/master/MozillaReality/FirefoxReality/l10n.toml

daoshengmu:
Close it because it is not ongoing.

PR #3583, opened 2020-06-30
Fixes #1537 Context menu Web Search

### LINKED ISSUE #1537: Context menu web search
There should be a way for users to select a piece of text on a webpage and easily perform a web search on that term with their default search engine. This might be similar to the context menu that appears when text is selected on a page in Android/mobile UI.

keianhzo:
Fixes #1537 Context menu Web Search


app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/dialogs/SelectionActionWidget.java
```diff
@@ -189,8 +215,15 @@ private View createSeparator() {
     }
 
     private void handleAction(View sender) {
+        String action = (String)sender.getTag();
+        if (action.equals(ACTION_WEB_SEARCH) && mSelectionText != null) {
+            mWidgetManager.getWindows().addTab(
+                    mWidgetManager.getWindows().getFocusedWindow(),
+                    UrlUtils.urlForText(getContext(), mSelectionText));
+        }
+
```
bluemarvin, line 224: Can we modify `addTab()` so that if the current window has a Session, it becomes the parent of the new Session so that the back button works. I want to be able to do a search with google, the hit the back button and go back to what I was looking at. I think we should do this with `what's new` and any other button that creates a new tab with web content but that can be a follow up.

keianhzo: Good call. I've added that.

PR #3589, opened 2020-07-01
Fix CocurrentlModificationException in debug mode

keianhzo:
This caused a crash mostly in the emulator when restoring sessions and only in DEBUG mode (release builds are not affected):

STRs:
- Open a few sessions/windows
- Exist the app (persist sessions)
- Kill the app
- Reopen

Result: Crash
Expected: Sessions should be restored

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStore.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStore.java
@@ -149,8 +149,12 @@ public class SessionStore implements
         if (BuildConfig.DEBUG) {
             mStoreSubscription = ComponentsAdapter.get().getStore().observeManually(browserState -> {
                 Log.d(LOGTAG, "Session status BEGIN");
-                browserState.getTabs().forEach(tabSessionState -> Log.d(LOGTAG, "BrowserStore Session: " + tabSessionState.getId()));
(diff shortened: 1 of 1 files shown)


app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStore.java
```diff
@@ -149,8 +149,12 @@ public void onTrackingProtectionLevelUpdated(int level) {
         if (BuildConfig.DEBUG) {
             mStoreSubscription = ComponentsAdapter.get().getStore().observeManually(browserState -> {
                 Log.d(LOGTAG, "Session status BEGIN");
-                browserState.getTabs().forEach(tabSessionState -> Log.d(LOGTAG, "BrowserStore Session: " + tabSessionState.getId()));
-                mSessions.forEach(session -> Log.d(LOGTAG, "SessionStore Session: " + session.getId()));
```
bluemarvin, line 153: How was this causing the exception?

keianhzo: `observeManually` get called in a pool thread every time the `BrowserStore` content changes so in some situations where we add/destroy sessions it might get called at the same time as a session is added/destroyed in the main thread. An easier way of reproducing it is just going to the tabs view and select and close all. 

We might also just remove that debugging code, I would like to move all that to a SessionStore unit test class, I'm doing some work on that but there is some refactoring involved and that's always tricky.

PR #3599, opened 2020-07-01
Add support for GV beta and release builds

### LINKED ISSUE #3369: WebXR/WebVR immersive mode fails when using beta and release versions of GeckoView
## Configuration

Firefox Reality version: All

Hardware: All

## Steps to Reproduce

1. Build FxR with a beta or release version of GeckoView

## Current Behavior

FxR fails to enter immersive mode.

## Expected Behavior

FxR should be able to enter immersive mode.

## Context
FxR currently uses GeckoView nightly for releases. This cause a great deal of churn and instability in the product. To help reduce issues FxR should move to beta and eventually the releases version of GeckoView. Currently when we use beta and release versions of GeckoView, immersive mode fails to work. I have verified that I was using the correct `moz_external_vr.h` file for the version of Gecko. Figuring out why immersive mode fails is the first step in moving to beta and release versions of GeckoView.

MortimerGoro:
Fixes #3369

I could add a flavor or user property to add more dynamic options to select the branch if we consider it's worth it

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -492,11 +492,10 @@ if (findProject(':wavesdk')) {
 }
 
 dependencies {
-    // To see what the latest geckoview-nightly version is go here:
-    // https://maven.mozilla.org/?prefix=maven2/org/mozilla/geckoview/geckoview-nightly-armeabi-v7a/
-    armImplementation deps.gecko_view.nightly_armv7a
-    arm64Implementation deps.gecko_view.nightly_arm64
-    x86_64Implementation deps.gecko_view.nightly_x86_64
+    // "nightly" or "beta" or "release"
+    def branch = "nightly"
+    arm64Implementation deps.gecko_view."${branch}_arm64"
+    x86_64Implementation deps.gecko_view."${branch}_x86_64"
 }
 
 if (gradle.hasProperty('geckoViewLocalArm') || gradle.hasProperty('geckoViewLocalX86')) {
--- a/versions.gradle
+++ b/versions.gradle
@@ -24,7 +24,9 @@ ext.deps = [:]
 def versions = [:]
 // GeckoView versions can be found here:
 // https://maven.mozilla.org/?prefix=maven2/org/mozilla/geckoview/
-versions.gecko_view = "79.0.20200625094452"
+versions.gecko_view_nightly = "79.0.20200625094452"
+versions.gecko_view_beta = "79.0.20200630191632"
+versions.gecko_view_release = "78.0.20200630195452"
 versions.android_components = "44.0.0"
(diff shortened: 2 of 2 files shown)


versions.gradle
```diff
@@ -56,9 +58,15 @@ ext.versions = versions
...
-gecko_view.nightly_x86_64 = "org.mozilla.geckoview:geckoview-nightly-x86_64:$versions.gecko_view"
+gecko_view.nightly_armv7a = "org.mozilla.geckoview:geckoview-nightly-armeabi-v7a:$versions.gecko_view_nightly"
+gecko_view.nightly_arm64 = "org.mozilla.geckoview:geckoview-nightly-arm64-v8a:$versions.gecko_view_nightly"
+gecko_view.nightly_x86_64 = "org.mozilla.geckoview:geckoview-nightly-x86_64:$versions.gecko_view_nightly"
+gecko_view.beta_armv7a = "org.mozilla.geckoview:geckoview-beta-armeabi-v7a:$versions.gecko_view_beta"
+gecko_view.beta_arm64 = "org.mozilla.geckoview:geckoview-beta-arm64-v8a:$versions.gecko_view_beta"
+gecko_view.beta_x86_64 = "org.mozilla.geckoview:geckoview-beta-x86_64:$versions.gecko_view_beta"
+gecko_view.release_armv7a = "org.mozilla.geckoview:geckoview-armeabi-v7a:$versions.gecko_view_release"
+gecko_view.release_arm64 = "org.mozilla.geckoview:geckoview-arm64-v8a:$versions.gecko_view_release"
+gecko_view.release_x86_64 = "org.mozilla.geckoview:geckoview-x86_64:$versions.gecko_view_release"
```
bluemarvin, line 69: I don't think we need the armv7a anymore?

MortimerGoro: correct, I have removed the armv7 dependencies

app/build.gradle
```diff
@@ -492,11 +492,10 @@ if (findProject(':wavesdk')) {
 }
 
 dependencies {
-    // To see what the latest geckoview-nightly version is go here:
-    // https://maven.mozilla.org/?prefix=maven2/org/mozilla/geckoview/geckoview-nightly-armeabi-v7a/
-    armImplementation deps.gecko_view.nightly_armv7a
-    arm64Implementation deps.gecko_view.nightly_arm64
-    x86_64Implementation deps.gecko_view.nightly_x86_64
+    // "nightly" or "beta" or "release"
+    def branch = "nightly"
```
bluemarvin, line 496: I wonder if branch should be set in `versions.gradle`? Probably fine since we need to think about how we are going to handle this for future releases.

PR #3633, opened 2020-07-06
Update supported languages strings comment to reflect all uses

keianhzo:
Relates to https://github.com/mozilla-l10n/android-l10n/pull/243#discussion_r447292984

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -220,11 +220,13 @@
          Below this string appears a list of locales for the user to choose as their preferred languages for web content. -->
     <string name="settings_language_choose_language_content_description">Web pages are sometimes offered in more than one language. Choose the languages you want to display in order of preference to the left side.</string>
 
-    <!-- This string is used to label a radio button in the 'Settings' Voice Search and the Display Language dialogs that, when pressed,
-    changes the app and the language of the speech-recognition-based search to 'English (US)'.
-    The string is also used in the keyboard space bar to show the current keyboard language and in
+    <!-- The string is used in the keyboard space bar to show the current keyboard language and in
     the keyboard languages panel list. -->
-    <string name="settings_language_english">English (US)</string>
+    <string name="settings_language_english">English</string>
+
+    <!-- This string is used to label a radio button in the 'Settings' Voice Search and the Display Language dialogs that, when pressed,
+    changes the app and the language of the speech-recognition-based search to 'English (US)'. -->
+    <string name="settings_language_english_us">English (US)</string>
 
     <!-- This string is used to label a radio button in the 'Settings' Voice Search and the Display Language dialogs that, when pressed,
     changes the app and the language of the speech-recognition-based search to 'English (UK)'. -->
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -220,88 +220,118 @@
          Below this string appears a list of locales for the user to choose as their preferred languages for web content. -->
     <string name="settings_language_choose_language_content_description">Web pages are sometimes offered in more than one language. Choose the languages you want to display in order of preference to the left side.</string>
 
-    <!-- This string is used to label a radio button in the 'Settings' Language dialog that, when pressed,
-         changes the app and the language of the speech-recognition-based search to 'English (US)'. -->
+    <!-- This string is used to label a radio button in the 'Settings' Voice Search and the Display Language dialogs that, when pressed,
+    changes the app and the language of the speech-recognition-based search to 'English (US)'.
+    The string is also used in the keyboard space bar to show the current keyboard language and in
+    the keyboard languages panel list. -->
     <string name="settings_language_english">English (US)</string>
 
-    <!-- This string is used to label a radio button in the 'Settings' Language dialog that, when pressed,
-         changes the app and the language of the speech-recognition-based search to 'English (UK)'. -->
+    <!-- This string is used to label a radio button in the 'Settings' Voice Search and the Display Language dialogs that, when pressed,
+    changes the app and the language of the speech-recognition-based search to 'English (UK)'. -->
     <string name="settings_language_english_uk">English (UK)</string>
 
-    <!-- This string is used to label a radio button in the settings language dialog that, when pressed,
-         changes the app and the language of the speech-recognition-based search to 'Chinese (Traditional/Zhuyin)'. -->
+    <!-- This string is used to label a radio button in the 'Settings' Voice Search and the Display Language dialogs that, when pressed,
+    changes the app and the language of the speech-recognition-based search to 'Chinese (Traditional/Zhuyin)'.
+    The string is also used in the keyboard space bar to show the current keyboard language and in
+    the keyboard languages panel list. -->
     <string name="settings_language_traditional_chinese">Chinese (Traditional/Zhuyin)</string>
 
-    <!-- This string is used to label a radio button in the settings language dialog that, when pressed,
-         changes the app and the language of the speech-recognition-based search to 'Chinese (Simplified/Pinyin)'. -->
+    <!-- This string is used to label a radio button in the 'Settings' Voice Search and the Display Language dialogs that, when pressed,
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -221,87 +221,129 @@
...
   
     <!-- This string is used to label a radio button in the settings language dialog that, when pressed,
-     changes the app and the language of the speech-recognition-based search to 'Dutch'. -->
+    changes the app and the language of the speech-recognition-based search to 'Dutch'.
+    The string is also used in the keyboard space bar to show the current keyboard language and in
+    the keyboard languages panel list. -->
     <string name="settings_language_dutch">Dutch</string>
 
     <!-- This string is used to label a radio button in the settings language dialog that, when pressed,
-         changes the app and the language of the speech-recognition-based search to 'Thai'. -->
```
Delphine, line 304: As mentioned in the l10n PR, we don't have the app localized in Thai, so the comment will need to be updated to reflect that (I understand it should be that this string will "only" be used to change the keyboard layout?)

app/src/main/res/values/strings.xml
```diff
@@ -221,87 +221,129 @@
...
+    The string is also used in the keyboard space bar to show the current keyboard language and in
+    the keyboard languages panel list. -->
     <string name="settings_language_danish">Danish</string>
 
     <!-- This string is used to label a radio button in the settings language dialog that, when pressed,
-     changes the app and the language of the speech-recognition-based search to 'Norwegian'. -->
+    changes the app and the language of the speech-recognition-based search to 'Norwegian'.
+    The string is also used in the keyboard space bar to show the current keyboard language and in
+    the keyboard languages panel list. -->
     <string name="settings_language_norwegian">Norwegian</string>
```
Delphine, line 311: So I was wondering if we need this string for "Norwegian"? We already have both Norwegian Bokmal and Norwegian Nynorsk below, which are the usual Norwegian locales (and should suffice, I believe). 
Unless it's necessary on your side for some reason to add "Norwegian" as well? Honestly no idea :)

app/src/main/res/values/strings.xml
keianhzo: You are right, I've updated this PR to reflect that.

app/src/main/res/values/strings.xml
keianhzo: We are using the "Norwegian" string for the keyboard spacebar and the keyboard language list as there are not specific keyboard layouts for Bokmal and Nynorsk. Actually this happens also with Spanish/Spanish(Spain) and English(US)/English(UK) so I've updated those too.

I've also updated all the comments to reflect that we are using the language strings both in the Display and the Voice search languages settings panels.

app/src/main/res/values/strings.xml
```diff
@@ -220,88 +220,118 @@
          Below this string appears a list of locales for the user to choose as their preferred languages for web content. -->
     <string name="settings_language_choose_language_content_description">Web pages are sometimes offered in more than one language. Choose the languages you want to display in order of preference to the left side.</string>
 
-    <!-- This string is used to label a radio button in the 'Settings' Language dialog that, when pressed,
-         changes the app and the language of the speech-recognition-based search to 'English (US)'. -->
+    <!-- This string is used to label a radio button in the 'Settings' Voice Search and the Display Language dialogs that, when pressed,
+    changes the app and the language of the speech-recognition-based search to 'English (US)'.
+    The string is also used in the keyboard space bar to show the current keyboard language and in
+    the keyboard languages panel list. -->
```
Delphine, line 226: I'm confused. You mention that for the keyboard, only an "English" string would appear (as for Spanish and Norwegian). However, this string is "English (US)". So the comment doesn't seem quite right (shouldn't it be like the one for "English (UK)" below, and not call our keyboard?)
Am I missing something?

app/src/main/res/values/strings.xml
```diff
@@ -220,88 +220,118 @@
...
+    <!-- This string is used to label a radio button in the 'Settings' Voice Search and the Display Language dialogs that, when pressed,
+    changes the app and the language of the speech-recognition-based search to 'Danish'.
+    The string is also used in the keyboard space bar to show the current keyboard language and in
+    the keyboard languages panel list. -->
     <string name="settings_language_danish">Danish</string>
 
-    <!-- This string is used to label a radio button in the settings language dialog that, when pressed,
-     changes the app and the language of the speech-recognition-based search to 'Norwegian'. -->
+    <!-- This string is used to label a radio button in the 'Settings' Voice Search and the Display Language dialogs that, when pressed,
+    changes the app and the language of the speech-recognition-based search to 'Norwegian'. -->
```
Delphine, line 304: Shouldn't this call out the keyboard part as well? (vs the other two "Norwegians" below, since not used for keyboard)

app/src/main/res/values/strings.xml
```diff
@@ -220,88 +220,118 @@
...
-    <!-- This string is used to label a radio button in the settings language dialog that, when pressed,
- changes the app and the language of the speech-recognition-based search to 'Norwegian (Bokmaal)'. -->
+    <!-- This string is used to label a radio button in the 'Settings' Voice Search and the Display Language dialogs that, when pressed,
+    changes the app and the language of the speech-recognition-based search to 'Norwegian (Bokmaal)'. -->
     <string name="settings_language_norwegian_bokmaal">Norwegian (Bokmaal)</string>
 
-    <!-- This string is used to label a radio button in the settings language dialog that, when pressed,
- changes the app and the language of the speech-recognition-based search to 'Norwegian (Nynorsk)'. -->
+    <!-- The string is used in the keyboard space bar to show the current keyboard language and in
+    the keyboard languages panel list. -->
```
Delphine, line 312: I don't think this should call out the keyboard in comments. On the other hand: "...that, when pressed,
    changes the app and the language of the speech-recognition-based search to ..." seems missing)

app/src/main/res/values/strings.xml
```diff
@@ -220,88 +220,118 @@
...
+    <!-- This string is used to label a radio button in the 'Settings' Voice Search and the Display Language dialogs that, when pressed,
+    changes the app and the language of the speech-recognition-based search to 'Dutch'.
+    The string is also used in the keyboard space bar to show the current keyboard language and in
+    the keyboard languages panel list. -->
     <string name="settings_language_dutch">Dutch</string>
 
-    <!-- This string is used to label a radio button in the settings language dialog that, when pressed,
-         changes the app and the language of the speech-recognition-based search to 'Thai'. -->
+    <!-- The string is used in the keyboard space bar to show the current keyboard language and in
+    the keyboard languages panel list. -->
```
Delphine, line 334: This comment LGTM now

app/src/main/res/values/strings.xml
keianhzo: For some reason the default keyboard string for English is "English US", so we don't really have a string for just English. Should we maybe remove the "US" from there?

Delphine: Maybe we could rather add a string just for the keyboard, that says "English"? Not sure that's possible, but would be great if we can.
I'm not a fan of removing the "US" since the localization can differ from en-UK

app/src/main/res/values/strings.xml
keianhzo: I fixed this, now we have "Norwegian" only for the keyboard and the Norwegian dialects for the settings.

app/src/main/res/values/strings.xml
keianhzo: I've added this as a new commit. "English" will be used for the keyboard related tags and "English (US)"/"English (UK) for the language related settings similarly to what we do with Norwegian. Let me know if that sounds reasonable to you.

Delphine: That sounds fine! thank you

Delphine:
Oops I'm so sorry I merged this thinking it was in our repo! 😬

keianhzo:
It's ok, fortunately we had already branched the current release and we were going to do this anyway, so it's all good :)

PR #3639, opened 2020-07-07
Fixes #3637 Null check the alphabetic Caps Keyboard when switching shift

### LINKED ISSUE #3637: [Keyboard] The app crashes when clicking the Shift key
**[Affected Versions]:**
- Firefox Reality version 11 RC4 (c28ddb2).

**[Affected Platforms]:** 
- Pico Neo2 - Android 8.1.0

**[Prerequisites]:**

**[Steps to reproduce]:**
1. Focus the Awesomebar to display the keyboard.
2. Click the Shift key from the keyboard
3. Observe the behavior.

**[Expected results]:**
- The keyboard switches to uppercase.

**[Actual results]:**
- The app crashes.

**[Notes]:**
- This issue is reproducible with all keyboards (where Shift key is available) except the Thai keyboard.
- I was not able to verify this issue on the HTC devices because RC4 build can't be opened after installing/updating.
- Attached is a screen recording, logcat and the crash report of the issue:
	- Screen Recording: [link](https://drive.google.com/file/d/1VOj0Ucx1bPDwMVfcD8-bVU2J4UrF3nTF/view?usp=sharing).
	- Crash Report: [link](https://crash-stats.mozilla.org/report/index/f660c27d-0002-480a-a596-a37570200707).
	- Logcat: [link](https://drive.google.com/file/d/1VSFB_YjRhl4r2icGIRtx4mIPv9c2FEpl/view?usp=sharing).

keianhzo:
Fixes #3637 Null check the alphabetic caps Keyboard when entering a character. 

Also allow the alphabetic caps keyboard to be dismissed when a key is entered, I've checked other keyboards and it seems to work the same way as the rest.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/KeyboardWidget.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/KeyboardWidget.java
@@ -746,7 +746,10 @@ public class KeyboardWidget extends UIWidget implements CustomKeyboardView.OnKey
 
         if (mKeyboardView.getKeyboard() != getSymbolsKeyboard()) {
             if (isShifted || mIsCapsLock) {
-                mKeyboardView.setKeyboard(mCurrentKeyboard.getAlphabeticCapKeyboard());
+                if (mCurrentKeyboard.getAlphabeticCapKeyboard() != null) {
+                    mKeyboardView.setKeyboard(mCurrentKeyboard.getAlphabeticCapKeyboard());
+                }
+
             } else {
                 mKeyboardView.setKeyboard(mCurrentKeyboard.getAlphabeticKeyboard());
             }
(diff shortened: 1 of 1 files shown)


app/src/common/shared/org/mozilla/vrbrowser/ui/widgets/KeyboardWidget.java
```diff
@@ -1049,7 +1049,7 @@ private void handleText(String aText, boolean skipCase) {
             postInputCommand(() -> connection.commitText(text, 1));
         }
 
-        if (!mIsCapsLock && mCurrentKeyboard.getAlphabeticCapKeyboard() == null) {
```
daoshengmu, line 1049: We can't remove `&& mCurrentKeyboard.getAlphabeticCapKeyboard() == null`. It will cause when switching to the Cap mode, typing any character on the keyboard, it will force to switch back to the un-cap mode.

daoshengmu: It seems to be acceptable in iOS keyboard. Let me check it with an Android phone.

PR #3644, opened 2020-07-07
Fixes #3643 Telemetry crash fix

### LINKED ISSUE #3643: [Oculus Quest] The app keeps crashing in the background after the latest Oculus update
**[Affected Platforms]:** 
- Firefox Reality version 11 RC4 (c28ddb2).

**[Prerequisites]:**
- The Firefox Reality app is installed.

**[Steps to reproduce]:**
1. Open the FxR browser.
2. Click the Oculus button and Quit the app.
3. Wait for ~ 5 seconds and observe the behavior.

**[Expected results]:**
- The new device menu is displayed and nothing happens if no actions are made.

**[Actual results]:**
- The "Firefox Reality keeps stopping" prompt is displayed.

**[Notes]:**
- The prompt is displayed multiple times while navigating the device menu.
- This issue could also affect the production version but I'll follow up with a comment given the urgency of this issue.
	- Screen recording: [link](https://drive.google.com/file/d/1r_trjPgYLu2U1A_UEkivUb-jZszBgVhU/view?usp=sharing).
	- Logcat: [link](https://drive.google.com/file/d/1Etk1lHdMlJoX2Vq3EsvrYMeZ6xsZNka8/view?usp=sharing).

keianhzo:
Fixes #3643 The Telemetry service needs to get initialized in the Application onCreate as the service can be called when the app is not running and that only calls the Application onCreate, otherwise this crash happens.

I had to add a custom TestApplication for testing as the Unit testing initialization doesn't seems to provide a different process for the crash service as defined in the manifest and the GeckoRuntime crashes with:
`java.lang.IllegalArgumentException: Crash handler service must run in a separate process`

Also for some reason the Application onCreate is called from a different process the first time the app is launched and the GeckoRuntime is not correctly initialized and we don't get any content paints. I've added the same check as Fenix does to make sure we are running in the main process or just return otherwise.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserApplication.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/VRBrowserApplication.java
@@ -27,6 +27,7 @@ import org.mozilla.vrbrowser.ui.widgets.AppServicesProvider;
 import org.mozilla.vrbrowser.utils.BitmapCache;
 import org.mozilla.vrbrowser.utils.EnvironmentsManager;
 import org.mozilla.vrbrowser.utils.LocaleUtils;
+import org.mozilla.vrbrowser.utils.SystemUtils;
 
 public class VRBrowserApplication extends Application implements AppServicesProvider {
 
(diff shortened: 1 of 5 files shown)


app/src/common/shared/org/mozilla/vrbrowser/VRBrowserApplication.java
```diff
@@ -42,6 +43,17 @@
...
+        if (!SystemUtils.isMainProcess(this)) {
+            // If this is not the main process then do not continue with the initialization here. Everything that
+            // follows only needs to be done in our app's main process and should not be done in other processes like
+            // a GeckoView child process or the crash handling process. Most importantly we never want to end up in a
+            // situation where we create a GeckoRuntime from the Gecko child process.
+            return;
+        }
+
+        TelemetryWrapper.init(this, EngineProvider.INSTANCE.getDefaultClient(this));
+        GleanMetricsService.init(this, EngineProvider.INSTANCE.getDefaultClient(this));
```
bluemarvin, line 56: It's unfortunate glean requires this.

PR #3654, opened 2020-07-08
Assert that vrPrefsWorkAround method is called before creating the Gecko Runtime

MortimerGoro:
We have had different regressions related to not calling vrPrefsWorkAround before runtime. I'm adding a error in that case to prevent that from happenning again


MortimerGoro:
I also removed ADB extra preferences, having to wait for activity intent makes initialization from Application harder and with more risks.

We are going to switch to https://firefox-source-docs.mozilla.org/mobile/android/geckoview/consumer/automation.html#reading-configuration-from-a-file for testing/benchmarking preferences (cc @fernandojsg)

PR #3669, opened 2020-07-13
Fixes #3671 BrowserState-SessionStore sync

### LINKED ISSUE #3671: [FxA] The Log In page is displayed in a loading state when trying to login
**[Affected Versions]:**
- Firefox Reality version 11 RC7 (45f818a).

**[Affected Platforms]:** 
- Pico Neo2 - Android 8.1.0

**[Prerequisites]:**
- No FxA account is logged in.

**[Steps to reproduce]:**
1. Open the Settings menu.
2. Click the "Sign Into Account" card.
3. Observe the behavior.

**[Expected results]:**
- The FxA Log In page is correctly displayed.

**[Actual results]:**
- The FxA Log In page is displayed in a loading state.

**[Notes]:**
- The issue is also reproducible if trying to log in from the "Send Tab" buttons.
- The account remains logged in after updating the app. The issue is only reproducible when trying to log in.
	- Screen Recording: [link](https://drive.google.com/file/d/1u4Pt8C_trXvXb0cmZV96hhvKpgiflvb3/view?usp=sharing).

keianhzo:
Fixes #3671

Make sure to test the following scenarios:
- New Window creation: from either Tray/Context menu
- New Tab creation: from either the Tabs panel/Context menu
- New Background tab creation (Long press link and open tab from the Context menu)
- Session restore: quit and reopen app with active&suspended sessions (Suspended sessions won't be added to the Browser Store until first activated)
- Session suspension/restore: open 3+ windows in regular model, switch to private, open another 3+ windows and go back and forth between regular/private
- Session recreation: Open several active/suspeded sessions, got to Settings->Privacy and clear cache/data

**Logcat** filtering by `SessionStore` will show the `BrowserStore`-`SessionStore` sync state including the alive vs suspended totals. 

**Note**: Suspended session will be alive at the `SessionStore` level but not at the `BrowserStore` level, we remove them from the `BrowserStore` when suspended, we only want alive ones there.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/Session.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/Session.java
@@ -91,7 +91,6 @@ public class Session implements ContentBlocking.Delegate, GeckoSession.Navigatio
     private transient byte[] mPrivatePage;
     private transient boolean mFirstContentfulPaint;
     private transient long mKeepAlive;
-    private transient boolean mSessionAdded;
 
     public interface BitmapChangedListener {
         void onBitmapChanged(Session aSession, Bitmap aBitmap);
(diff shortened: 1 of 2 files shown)


app/src/common/shared/org/mozilla/vrbrowser/browser/engine/Session.java
```diff
@@ -472,9 +472,9 @@ private void restore() {
 
         // We call restore when a session is first activated and when it's recreated.
         // We only need to notify of the session creation if it's not a recreation.
-        if (mSessionRemoved) {
+        if (addSession) {
             mSessionChangeListeners.forEach(listener -> listener.onSessionAdded(this));
-            mSessionRemoved = false;
+            mSessionAdded = true;
         }
```
bluemarvin, line 478: I don't understand why `mSessionAdded` can't be used. It would imply it is out of sync.

app/src/common/shared/org/mozilla/vrbrowser/browser/engine/Session.java
```diff
@@ -541,7 +541,7 @@ void recreateSession() {
 
         mState = mState.recreate();
 
-        restore();
+        restore(false);
```
bluemarvin, line 544: I did not want a boolean instead of the flags. I was trying to not have to add any parameters to restore. Booleans as a function parameter are almost always confusing.

keianhzo: The problem is that `restore` is called from `recreateSession`and `setActive`. `recreateSession` never needs to add the session but when called from `setActive` it depends on if the session has already been added or not so we can't rely on a single boolean to handle all the three cases.

bluemarvin: If the boolean can't represent the current state we need to figure out a different method. I'm still not sure I understand why it would not be correct. Seems like a bug to me.

keianhzo: I have updated and:
- Moved the `onSessionAdd` code to an `createSession` method so we have all the session add/remove/link/unlink code in the Session unless until we do the refactor.
- Removed the session add tracking flag by removing the session from the `BrowserStore` on suspend. I think this works as we don't really want suspended sessions there so I think is a legit point of session removal from the BrowserStore.

I've tested all the scenarios and seems to stay in sync.

**Logcat** filtering by `SessionStore` will show the `BrowserStore`-`SessionStore` sync state including the alive vs suspended totals.

**Note**: Suspended session will be alive in the `SessionStore` but not in the `BrowserStore` now, we remove them from the `BrowserStore` when suspended, we only want alive ones there.

keianhzo:
@MortimerGoro Shouldn't happen as we are calling `createSession` right after creating the `Session` instance with sets the state but I've added an static method to handle the whole session creation and avoid missing the state creation.

PR #3672, opened 2020-07-13
Remove superfluous WorkManager initialization

keianhzo:
Originally the `Service` instance was created in `Application` `onCreate` which happens before any other component has been created so the `WorkManager` was not initialized and that’s why we needed to initialize it during the `Service` instantiation, just to make sure that it was always ready.
We have recently done a code refactor and now the `Service` instance is create in `Activity` `onCreate` so this is not an issue anymore, as the component has already had a chance and should be initialized.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -447,9 +447,6 @@ dependencies {
     implementation deps.android_components.glean
     implementation deps.app_services.rustlog
 
-    // TODO this should not be necessary at all, see Services.kt
-    implementation deps.work.runtime
-
     // Kotlin dependency
     implementation deps.kotlin.stdlib
     implementation deps.kotlin.coroutines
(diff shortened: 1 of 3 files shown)


sraturi:
would it be better to implement https://developer.android.com/topic/libraries/architecture/workmanager/advanced/custom-configuration#on-demand  so that the application does the lazy loading for the work manager? Here is a issue related to this https://github.com/mozilla-mobile/fenix/issues/11698

keianhzo:
@sraturi The `work-runtime` module is included in FxR as a dependency of several AC modules: 
- org.mozilla.components:support-sync-telemetry
- org.mozilla.components:service-firefox-accounts
- org.mozilla.components:feature-accounts

We just added that initialization to workaround the limitation stated above at the time of adding support for AC services. So what does in this case mean:

> On-demand initialization lets you create WorkManager only when that component is needed, instead of every time the app starts up

Shouldn't be this in any case a responsibility of the AC modules using it instead if the app using those modules?

bluemarvin:
@keianhzo Can we remove this too? https://github.com/MozillaReality/FirefoxReality/blob/db4bae0648a36435db129750dba33ce60d2dbbb2/versions.gradle#L113

keianhzo:
@bluemarvin Good catch, updated.

bluemarvin:
@keianhzo The unit tests are now failing.

keianhzo:
It seems that `GleanTestRule` requires `work-testing`. Fixed.

PR #3707, opened 2020-07-20
BrowserStore sync checks

keianhzo:
- Under some conditions `mSessions` or `browserState` can be null during session loggin. Added null checks to avoid a crash.
- Added a check upon session addition into the `BrowserStore` to avoid adding a duplicated session.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStore.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionStore.java
@@ -146,6 +146,9 @@ public class SessionStore implements
 
         if (BuildConfig.DEBUG) {
             mStoreSubscription = ComponentsAdapter.get().getStore().observeManually(browserState -> {
+                if (mSessions == null || browserState == null) {
+                    return null;
+                }
                 Log.d(LOGTAG, "Session status BEGIN");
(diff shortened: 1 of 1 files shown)


app/src/common/shared/org/mozilla/vrbrowser/browser/adapter/ComponentsAdapter.kt
```diff
@@ -27,9 +31,17 @@ class ComponentsAdapter private constructor(
...
+        if (store.state.tabs.stream().noneMatch {
+                    it.id == tab.id
+        }) {
+            store.dispatch(TabListAction.AddTabAction(
+                    tab = tab
+            ))
+
+        } else {
+            Log.e(LOGTAG, "Duplicated session add attempt: " + tab.id)
+        }
```
bluemarvin, line 44: If you would rather crash in this case, I was concerned that even after the previous patches we were still hitting a crash not realizing it was being caused by debug code.

bluemarvin: Since we don't seem to have any of these crashes in v11 I think it is safe to not check before storing.

keianhzo: I've removed the duplicated session code.

PR #3711, opened 2020-07-20
Use the lastest version of VRB

bluemarvin:


bluemarvin:
I found a crash, will wait to update until the crash fix lands in VRB.

PR #3742, opened 2020-07-29
Delete prefs cache when application VERSION_CODE changes on taskcluster builds

### LINKED ISSUE #3722: Figure out how to clear preferences that were set in the FxR pref override function when no longer needed.
When we remove preference overrides, the last value persists. We need to find a way to clear the pref cache so FxR uses the default value.

See: https://github.com/MozillaReality/FirefoxReality/blob/93eaa9beb61a9ebabb2d7ac72f16fa74c850bfc7/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionUtils.java#L28

bluemarvin:
Fixes #3722

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionUtils.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionUtils.java
@@ -5,6 +5,7 @@ import android.net.Uri;
 import android.os.Bundle;
 import android.util.Log;
 
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
 import org.mozilla.gecko.GeckoProfile;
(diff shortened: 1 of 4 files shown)


app/src/common/shared/org/mozilla/vrbrowser/browser/engine/SessionUtils.java
```diff
@@ -25,12 +26,38 @@ public static boolean isLocalizedContent(@Nullable String url) {
         return url != null && (url.startsWith("about:") || url.startsWith("data:"));
     }
 
+    private static void checkPrefsFile(Context aContext, @NonNull String path) {
+        int lastVersionCode = SettingsStore.getInstance(aContext).getPrefsLastResetVersionCode();
+        if (BuildConfig.VERSION_CODE > lastVersionCode) {
```
MortimerGoro, line 31: We should use != instead of for downgrades

bluemarvin: You have to uninstall first. You can’t install an older version number over a newer one.

MortimerGoro: ok, it can only be done for debuggable builds which is not the case of Taskcluster.

bluemarvin: Verified debug taskcluster builds have the same behavior as release builds.

PR #3754, opened 2020-08-03
Implement WebXR nativeFramebufferScaleFactor

### LINKED ISSUE #3724: Add support for WebXR nativeFramebufferScaleFactor
See https://bugzilla.mozilla.org/show_bug.cgi?id=1655006

@feiss found that the resolution of the FxR A-frame prototype is not very good to correctly evaluate the UI designs. I suggested that he used `framebufferScaleFactor` in the XRWebGLLayers constructor to force a higher resolution (there are not other good alternatives until WebXR layers are supported).

The workaround improved the quality on Oculus Browser but not in FxR. We need to correctly implement nativeFramebufferScaleFactor math to support this use case

MortimerGoro:
Fixes #3724

### DIFF
--- a/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
+++ b/app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
@@ -753,6 +753,12 @@ DeviceDelegateOculusVR::RegisterImmersiveDisplay(ImmersiveDisplayPtr aDisplay) {
   uint32_t width, height;
   m.GetImmersiveRenderSize(width, height);
   m.immersiveDisplay->SetEyeResolution(width, height);
+  // The maxScaleFactor that can be applied to the recommended render size.
+  // Oculus Browser uses a 1.68 maxScale factor.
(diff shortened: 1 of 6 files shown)


app/src/oculusvr/cpp/DeviceDelegateOculusVR.cpp
```diff
@@ -753,6 +753,8 @@ DeviceDelegateOculusVR::RegisterImmersiveDisplay(ImmersiveDisplayPtr aDisplay) {
   uint32_t width, height;
   m.GetImmersiveRenderSize(width, height);
   m.immersiveDisplay->SetEyeResolution(width, height);
+  const float scaleFactor = 2048.0f / width;
```
bluemarvin, line 756: Where does the `2048.0` magic number come from?

bluemarvin: Please add a comment where the `2048.0f` came from.

PR #1854, opened 2019-09-17
Replace telemetry service with Glean. (Stage 1)

daoshengmu:
Stage1. Per #1420

PR #3758, opened 2020-08-04
Set startup metrics after Glean is initialized

daoshengmu:
We were setting  `distribution_channel_name` before `Glean.initialize` and using `org_mozilla_vrbrowser.baseline` schema to query this metrics. But, it seems like the recent changes in Glean side makes us can't see `distribution_channel_name` in baseline pings.

In addition, even in `org_mozilla_vrbrowser.metrics` schema, we still have 20% chance to see the value of `distribution_channel_name` is null, and it starts from A-C around version 44 (https://sql.telemetry.mozilla.org/queries/73600/source#184108). So, I think we need to consider to move it back to `Glean.initialize` and make sure all our metric pings could have its values instead a null.

### DIFF
--- a/app/src/common/shared/org/mozilla/vrbrowser/telemetry/GleanMetricsService.java
+++ b/app/src/common/shared/org/mozilla/vrbrowser/telemetry/GleanMetricsService.java
@@ -63,16 +63,13 @@ public class GleanMetricsService {
         initialized = true;
 
         final boolean telemetryEnabled = SettingsStore.getInstance(aContext).isTelemetryEnabled();
-        if (telemetryEnabled) {
-            GleanMetricsService.start();
-        } else {
-            GleanMetricsService.stop();
-        }
         Configuration config = new Configuration(
                 ConceptFetchHttpUploader.fromClient(client),
                 Configuration.DEFAULT_TELEMETRY_ENDPOINT,
                 BuildConfig.BUILD_TYPE);
-        Glean.INSTANCE.initialize(aContext, true, config);
+
+        Glean.INSTANCE.initialize(aContext, telemetryEnabled, config);
+        setStartupMetrics();
     }
 
     // It would be called when users turn on/off the setting of telemetry.
--- a/app/src/test/java/org/mozilla/vrbrowser/GleanMetricsServiceTest.kt
+++ b/app/src/test/java/org/mozilla/vrbrowser/GleanMetricsServiceTest.kt
@@ -63,8 +63,15 @@ class GleanMetricsServiceTest {
         GleanMetricsService.testSetStartupMetrics()
         assertTrue(Distribution.channelName.testHasValue())
         assertEquals(Distribution.channelName.testGetValue(),
-                if (org.mozilla.vrbrowser.utils.DeviceType.isOculusBuild()) "oculusvr"
-                else BuildConfig.FLAVOR_platform)
+                org.mozilla.vrbrowser.utils.DeviceType.getDeviceTypeId());
+
+        // Make sure the distribution channel name is set after
+        // the telemetry system is switch off/on.
+        GleanMetricsService.stop();
+        GleanMetricsService.start();
+        assertTrue(Distribution.channelName.testHasValue())
+        assertEquals(Distribution.channelName.testGetValue(),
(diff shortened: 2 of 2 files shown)


daoshengmu:
@Dexterp37 Please help confirm I am doing right.

daoshengmu:
I suppose ` Glean.INSTANCE.setUploadEnabled(false)` will not clean up the value of `distribution_channel_name`, and we still keep that value for the metrics after making `setUploadEnabled(true)` later.

daoshengmu:
I wonder if this issue was caused by [Dropped "application lifetime" metrics in the Glean metrics ping](https://github.com/mozilla/glean/pull/1069). If even we bump into AC 51, are we good enough? Or instead, we should just do what I did in this PR, moving back this metric setup after `Glean.Initialize()`

app/src/common/shared/org/mozilla/vrbrowser/telemetry/GleanMetricsService.java
```diff
@@ -73,13 +73,13 @@ public static void init(@NonNull Context aContext, @NonNull Client client) {
...
                 BuildConfig.BUILD_TYPE);
         Glean.INSTANCE.initialize(aContext, true, config);
+        setStartupMetrics();
     }
 
     // It would be called when users turn on/off the setting of telemetry.
     // e.g., SettingsStore.getInstance(context).setTelemetryEnabled();
     public static void start() {
         Glean.INSTANCE.setUploadEnabled(true);
-        setStartupMetrics();
```
Dexterp37, line 82: The only startup metric you're setting is the `channel_name`. I remember discussing this in #1854, but I don't remember the details and I can't find it from a quick skim of the old PR. Why do we need this metric at all? Can't we use the [channel field from `Configuration`](https://mozilla.github.io/glean/javadoc/glean/mozilla.telemetry.glean.config/-configuration/channel.html)?

Dexterp37:
> 
> 
> We were setting `distribution_channel_name` before `Glean.initialize` and using `org_mozilla_vrbrowser.baseline` schema to query this metrics. But, it seems like the recent changes in Glean side makes us can't see `distribution_channel_name` in baseline pings.
> 
> In addition, even in `org_mozilla_vrbrowser.metrics` schema, we still have 20% chance to see the value of `distribution_channel_name` is null, and it starts from A-C around version 44 (https://sql.telemetry.mozilla.org/queries/73600/source#184108). So, I think we need to consider to move it back to `Glean.initialize` and make sure all our metric pings could have its values instead a null.

I believe that the problems above are a symptom of what I've commented in this PR.

app/src/common/shared/org/mozilla/vrbrowser/telemetry/GleanMetricsService.java
daoshengmu: That channel field is used for us to record the build is for `debug` or `release`.

app/src/common/shared/org/mozilla/vrbrowser/telemetry/GleanMetricsService.java
```diff
@@ -63,23 +63,19 @@ public static void init(@NonNull Context aContext, @NonNull Client client) {
...
+
+        Glean.INSTANCE.initialize(aContext, telemetryEnabled, config);
+        setStartupMetrics();
     }
 
     // It would be called when users turn on/off the setting of telemetry.
     // e.g., SettingsStore.getInstance(context).setTelemetryEnabled();
     public static void start() {
         Glean.INSTANCE.setUploadEnabled(true);
-        setStartupMetrics();
```
Dexterp37, line 82: You should retain this as well: if user toggle telemetry of and then on again, the metrics will be cleared and you'll need to set them again.

daoshengmu: Yes, you are right. I use Glean testing API can verify my channel name is cleared once users switch off Glean.

Dexterp37: Cool, let me know once you restore this line so that I will happily r+ :)

daoshengmu: Restored. :)

PR #3763, opened 2020-08-05
Battery levels api

bluemarvin:


bluemarvin:
This will be included in the Status Bar update PR.

PR #3764, opened 2020-08-06
Login Autocompletion and Sync support

keianhzo:
Closes ##719 Login Autocompletion and Sync support.


app/src/common/shared/org/mozilla/vrbrowser/browser/Accounts.kt
```diff
@@ -329,6 +329,10 @@ class Accounts constructor(val context: Context) {
             SyncEngine.History -> {
                 GleanMetricsService.FxA.historySyncStatus(value)
             }
+            //TODO
+//            SyncEngine.Passwords -> {
+//                GleanMetricsService.FxA.loginsSyncStatus(value)
+//            }
```
bluemarvin, line 335: Does this mean saved passwords are not synced back to FxA from FxR?

keianhzo: It was just a comment to remember to add Telemetry, I've removed it. I'll open a task to add Telemetry for both Addons and Autofill as the review process takes a while and we are already out of time.

app/src/common/shared/org/mozilla/vrbrowser/browser/LoginStorage.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    private val passwordsEncryptionKey by lazy {
+        getSecureAbove22Preferences().getString(PASSWORDS_KEY)
+                ?: generateEncryptionKey(KEY_STRENGTH).also {
+                    if (SettingsStore.getInstance(context).isPasswordsEncryptionKeyGenerated) {
+                        // We already had previously generated an encryption key, but we have lost it
+                        Log.d(LOG_TAG,"Passwords encryption key for passwords storage was lost and we generated a new one")
+                    }
+                    SettingsStore.getInstance(context).recordPasswordsEncryptionKeyGenerated()
+                    getSecureAbove22Preferences().putString(PASSWORDS_KEY, it)
```
MortimerGoro, line 42: Should we delete the value set in `getSecureAbove22Preferences` when logging out of FxA? I'm thinking on the use case of sharing a device with multiple people and see if there are secutity implications

keianhzo: `getSecureAbove22Preferences` just generates a key to encrypt the logins storage so I don't think there are security implications in using the same key after an FxA logout?

Regarding the store lifecycle after an FxA logout, there is no way of selectively deleting only the synced logins after an FxA logout so we have to just wipe out all the login storage. There is a `wipeLocal` method that should do that but unfortunately doesn't work as expected and it just removes everything (https://github.com/mozilla-mobile/android-components/issues/8079). I guess removing everything is better than leaving logins after an FxA logout.

PR #3957, opened 2021-08-24
Pico Neo3 controllers

keianhzo:
Pico Neo3 controllers update


keianhzo:
@MortimerGoro Can you take a look at this? Thanks!

app/src/main/cpp/moz_external_vr.h
```diff
@@ -150,6 +150,7 @@ enum class VRControllerType : uint8_t {
   PicoGaze,
   PicoG2,
   PicoNeo2,
+  PicoNeo3,
```
MortimerGoro, line 153: We shouldn't modify the moz_external_vr.h file without a corresponding patch in Gecko, or it might break WebXR.

I'm working on a OpenXR & WebXR Input refactor, and I'm going to get rid of all the input enum types in moz_external_vr.h and use the standard input profile strings instead. So, for now you can just reuse a controller with similar button mapping there (such as neo2) and we'll land the correct input implementation in Gecko once this is ready

MortimerGoro (changes requested):

PR #3958, opened 2021-08-24
Reset controllers state on pause

keianhzo:
This PR fixes an issue that caused the controllers not to be rendered after the activity is resumed

### DIFF
--- a/app/src/picovr/cpp/DeviceDelegatePicoVR.cpp
+++ b/app/src/picovr/cpp/DeviceDelegatePicoVR.cpp
@@ -71,6 +71,15 @@ struct DeviceDelegatePicoVR::State {
     bool IsRightHand() const {
       return hand == ElbowModel::HandEnum::Right;
     }
+
+    void Reset() {
+        created = false;
+        enabled = false;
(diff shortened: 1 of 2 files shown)


keianhzo:
@MortimerGoro

app/src/picovr/cpp/DeviceDelegatePicoVR.cpp
```diff
@@ -71,6 +71,15 @@ struct DeviceDelegatePicoVR::State {
...
       return hand == ElbowModel::HandEnum::Right;
     }
+
+    void Reset() {
+        created = false;
+        enabled = false;
+        touched = false;
+        is6DoF = false;
+        transform = vrb::Matrix::Identity();
+        hand = ElbowModel::HandEnum::Right;
```
MortimerGoro, line 81: The hand is currently set up in Initialize(). After Pause() & Resume() flow is called it looks like the ElbowModel will always be HandEnum::Right and never left.

MortimerGoro (changes requested):

PR #3965, opened 2021-10-01
OpenXR Input Refactor

MortimerGoro:
This PR makes the OpenXR input code more easy to understand, mantainable and scalable. The previous input code was very tied to the first Oculus prototype and not designed to handle different devices at the same time.

Now all the mappings are defined in the OpenXRInputMappings.h file


keianhzo (approved):
LGTM

PR #3994, opened 2022-01-21
GitHub actions gradle

svillar:
This is a temptative implementation of CI for push and PRs using gradle and github actions


svillar:
this is not for MR

